PR #15, opened 2017-06-07
Fixed issue #11: SPI constructors and SS pin usage modified

### LINKED ISSUE #11: SPI issue
Using another SPI instance created by constructor with arguments failed

fprwi6labs:
Fix #11:

- SPI constructor modified. SS pin should be set only if hardware NSS pin is required.
- Hardware NSS pin can be enabled. If NSS enabled, software CS is disabled.

See microcontroller documentation for hardware NSS pin available.

### DIFF
--- a/cores/arduino/stm32/clock.c
+++ b/cores/arduino/stm32/clock.c
@@ -98,13 +98,22 @@
   */
 
 /**
-  * @brief  Function called wto read the current micro second
+  * @brief  Function called to read the current micro second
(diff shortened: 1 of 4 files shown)


fpistm:
NSSPMode is only available for F0/F3/F7/L4.

fprwi6labs:
I removed NSSPMode field from the configuration. It is not an useful option in our case.

PR #19, opened 2017-06-07
Fix #18 PWM frequency

### LINKED ISSUE #18: PWM freq not correct on all pins.
PWM frequencies is not the same on all pins.
Depending of the timer used,  clock source is not the same (APB1/APB2) so the prescaler also.
`timHandle.Init.Prescaler         = (uint32_t)(SystemCoreClock / clock_freq) - 1`

fpistm:
Get the correct timer freq depending on timer clock source.

When TIMPRE bit of the RCC_DCKCFGR register is reset,
  * if APBx prescaler is 1, then TIMxCLK = PCLKx,
  * otherwise TIMxCLK = 2x PCLKx.

When TIMPRE bit in the RCC_DCKCFGR register is set,
 *   if APBx prescaler is 1,2 or 4, then TIMxCLK =HCLK,
 *   otherwise TIMxCLK = 4x PCLKx


fpistm:
Tests Fading/tone/Servo done on F0/F3/F4/F7/L4 -->ok

fprwi6labs:
PWM pins tested on L0/L4/F0/F3/F4 -> OK

I found another issue on F0 & F3: some PWM pins are not configured in `PeripheralPins.c`.
- F0: D3=PB3 & D6=PB10 not set
- F3: D6=PB10 not set

@fpistm can you confirm me that is a mistake? If yes I will push the patch.

fpistm:
Hi,
this is not really a mistake.
This only means by default PWM is not enabled on those pins but they could be added.
In fact, I have took as example mbed PeripheralPins.c for those boards that's why it is commented.

PR #29, opened 2017-06-12
Variant for Nucleo F030R8

fpistm:
Variant to add Nucleo F030R8 as example.
Start from a copy of Nucleo F091RC.
Then I try to commit major step to see what's changed.

Simple test executed:
Serial ok
Blink ok
Button ok
Analog ok.
Fading ok (pwm/IO)


RickKimball:
I probably should have removed all the comment lines in the board.txt file that reference the USB. The STM32F030R8 doesn't have native USB

https://github.com/stm32duino/Arduino_Core_STM32/pull/29/commits/a552dd7a953a0be3c36b633be351aae7960709c3#diff-fce09ad94aad2ba67472cd9142f00146R90

fpistm:
Yes, right

RickKimball:
I'd probably get rid if the "{build.usb_flags}" also

fpistm:
Right forgot it. updated.

RickKimball:
I noticed you are using a different scheme for the SYSCLK than what I did. Is there a way to be able to adjust the RCC_OscInitStruct.HSICalibrationValue with your configuration?  Using that variable I'm able to dial in my clock to be almost perfect.

fpistm:
In fact I used one from the CubeFW.
I have generate the systemClockConfig with cubeMx and it is the same like yours.
Use of RCC_OSCILLATORTYPE_NONE  tell that Oscillator configuration is unchanged
So it still using HSI.
I do not check which value is set by defaut for the calibration wether it set.
So added this seems better to be sure it is set.
RCC_OscInitStruct.OscillatorType = RCC_OSCILLATORTYPE_HSI;
RCC_OscInitStruct.HSIState = RCC_HSI_ON;
RCC_OscInitStruct.HSICalibrationValue = 15;
I will update it.

Maybe, we can also considering to declare the systemClockConfig as a weak function if user want tune it.

RickKimball:
Sure I like the idea of a flexible clock config. With the cortex-m0 some might want to run at a slower speed to get low power.

+1

fpistm:
Hi @RickKimball 
If you are agree, I will merge.
the last main diff with your porting is the linker script. I think.
I'm not very comfortable with that. But, no problem to take any update about that.

RickKimball:
Yeah the linker script change isn't a big deal. I was using mainly to see how much real memory was being used in the console window output.

Have you tested the SPI / I2C functions. I haven't looked at those yet.

fpistm:
Yes, i've made some tests for I2C/SPI:

For SPI, I've tested with the Arduino TFT shield (including SD) and with the mbed CI Test Shield (SD throug SPI).
For I2C, I've used the mbed CI Test Shield which include an eeprom and a temp/Hum sensor (LM75).

RickKimball:
OK I agree. Merge it!

RickKimball:
The README.md might be updated to indicate the F030R8 board is now supported.

RickKimball:
$ git diff README.md
diff --git a/README.md b/README.md
index 223d4c8..75ce92b 100644
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 Boards available:
 
   * STM32F0
+    * [Nucleo F030R8](http://www.st.com/en/evaluation-tools/nucleo-f030r8.html)
     * [Nucleo F091RC](http://www.st.com/en/evaluation-tools/nucleo-f091rc.html)
 
   * STM32F1 (soon)
$

fpistm:
done

PR #32, opened 2017-06-15
Fix SPI transfer16

troth:
When the SPI device is configured for MSB first, the 16 bit data being
sent and received on the SPI bus has the bytes swapped if the
SPI.transfer16() method is used. Swapping the bytes before and after the
call to spi_transfer() fixes the problem.

Tested using an AS5147P magnetic encoder spi device. Before the fix, saw
frame errors due to the bytes being swapped, when the same unmodified
code worked perfectly on an AVR based SparkFun RedBoard.

Signed-off-by: Theodore A. Roth <troth@openavr.org>

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -201,9 +201,15 @@ byte SPIClass::transfer(uint8_t _pin, uint8_t data, SPITransferMode _mode)
 uint16_t SPIClass::transfer16(uint8_t _pin, uint16_t data, SPITransferMode _mode)
 {
   uint16_t rx_buffer = 0;
+  uint16_t tmp;
 
   if (_pin > SPI_CHANNELS_NUM)
     return rx_buffer;
+
+  if (spiSettings[_pin].msb) {
+    tmp = ((data & 0xff00) >> 8) | ((data & 0xff) << 8);
+    data = tmp;
+  }
 
   if(_pin != g_active_id) {
     spi_init(&_spi, spiSettings[_pin].clk, spiSettings[_pin].dMode, spiSettings[_pin].msb);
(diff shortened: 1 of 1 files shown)


troth:
Found a byte swapping bug in the SPI.transfer16() method.

Tested on a Nucleo-F429ZI board connected to an AS5147P magnetic encoder.

fprwi6labs:
Hi @troth 

I tried with and without your modification and got the same result with my sketch test.
I don't have your magnetic device to try with it.

I need more information to be sure my test is not wrong:
Which board do you use?
What is your SPI configuration (speed, data mode, bit order)? 
And how do you configure the SPI? With SPISettings()?

troth:
The sketch code I used for testing is in this gist: ttps://gist.github.com/troth/a8f3084f0a21dfba72b43e90bb751275

I ran the same sketch on both a Nucleo-F429ZI board and on an AVR based Sparkfun RedBoard.

fprwi6labs:
I tested with LIS2DH12. 
I agree rx data are not in MSB order. Data should be swapped after transferred.
But if I swap tx data the device do not answer. I'm not sure about the fact to swap tx data before transfer.
I need to make more test about this point.

troth:
After 5 minutes looking at the data sheet for the LIS2DH12, I'm not sure that it is a candidate for using the transfer16() method.

I'll can post the screen captures from my oscope later if that would be useful.

fprwi6labs:
You are right. My device do not have 16bits registers. I don't have any device with 16bits register. 
I think you are more in position to bring this patch than me.
Do not hesitate to patch anything else if you find another mistake.
Thanks @troth 

@fpistm you can merge.

Issue #20, opened 2017-06-08
Merge Arduino_Core_STM32F1


fpistm:
https://github.com/stm32duino/Arduino_Core_STM32F1 have to be merged in this repo.

CMSIS ST and HALfrom STM32Cube_FW_F1 already included in this repo.

RickKimball:
I noticed you had a schedule for one of the other Issues * the low power one * and you mention a plan for week 43 on that one.  Is the F1 Core merge on the schedule with an ETA?

fprwi6labs:
@RickKimball 

The F1 core is planed for the next week. 
Be ready for a big test session. We need tester like you ;)

RickKimball:
Which boards? NUCLEO-F103RB + Bluepill? * I don't have a NUCLEO

fprwi6labs:
NUCLEO-F103RB + BluePill + MapleMini + STM32vldiscovery. All variant currently available.
I perform tests on nucleo. I have a Bluepill too but if you can do some on maple and discovery it will be perfect.

rogerclarkmelbourne:
I have the Maple Mini, so I will test during the weekend

fpistm:
Thanks @fprwi6labs

Adminius:
do you have special sketches/hw=extension board for testing?

fprwi6labs:
No, I haven't. I only use the example sketches.

fpistm:
List of boards to merge:
- [ ]     NUCLEO-F103RB
- [ ]     DISCO-F100RB (STM32VLDISCOVERY)
- [ ]     BLUEPILL-F103C8
- [ ]     MAPLEMINI-F103CB

PR #45, opened 2017-06-29
STM32F1 merged (replaced by #61)

fprwi6labs:
STM32F1 boards added. #20 

The following variants are available:
- NUCLEO-F103RB
- DISCO-F100RB (STM32VLDISCOVERY)
- BLUEPILL-F103C8
- MAPLEMINI-F103CB

I tested the following features only on the Nucleo board:
ADC, PWM, I2C (master & slave), SPI, UART, Servo, Tone, EEPROM

I require some help to test the other variants and especially all upload method for BluePill and MapleMini. In advance, thank you for your help.


rogerclarkmelbourne:
I will pull to my local repo and test later

rogerclarkmelbourne:
I will pull to my local repo and test later

rogerclarkmelbourne:
There are merge conflicts 

 cores/arduino/stm32/analog.c
cores/arduino/stm32/interrupt.c

fpistm:
Right Roger,
kept the NP not NC
kept the UNUSED.

rogerclarkmelbourne:
OK

I'll manually update, but I think you need to change interrupt.c as then it will auto merge when frederic merges it

analog.c I think frederic will need to manually merge

variants/BLUEPILL_F103C8/PeripheralPins.c
```diff
@@ -0,0 +1,195 @@
...
+  {PB0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC1_IN8
+  // {PB0,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC2_IN8
+  {PB1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC1_IN9
+  // {PB1,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC2_IN9
+  {NC,   NC,    0}
+};
+
+//*** DAC ***
+
+const PinMap PinMap_DAC[] = {
```
fpistm: could be deleted.
See https://github.com/stm32duino/Arduino_Core_STM32/commit/d5d4318de318e560363a5884730d4b4dacb0d164

variants/DISCO_F100RB/PeripheralPins.c
```diff
@@ -0,0 +1,206 @@
...
+  {PC3,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 13, 0)}, // ADC1_IN13
+  {PC4,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 14, 0)}, // ADC1_IN14
+  {PC5,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC1_IN15
+  {NC,   NC,    0}
+};
+
+//*** DAC ***
+
+const PinMap PinMap_DAC[] = {
+  {PA4,   DAC1, STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // DAC_OUT1
```
fpistm: Rename DAC1 by DAC

variants/DISCO_F100RB/PeripheralPins.c
```diff
@@ -0,0 +1,206 @@
...
+  {PA5,   DAC1, STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // DAC_OUT2
```
fpistm: ditto

variants/DISCO_F100RB/PeripheralPins.c
```diff
@@ -0,0 +1,206 @@
...
+const PinMap PinMap_SPI_SSEL[] = {
+    {PA4,   SPI1, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, 0)},
+    {PA15,  SPI1, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, 1)}, // GPIO_Remap_SPI1
+    {PB12,  SPI2, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, 0)},
+    {NC,    NC,    0}
+};
+
+//*** CAN ***
+
+const PinMap PinMap_CAN_RD[] = {
```
fpistm: Delete empty array
See https://github.com/stm32duino/Arduino_Core_STM32/commit/d5d4318de318e560363a5884730d4b4dacb0d164

variants/DISCO_F100RB/variant.h
```diff
@@ -0,0 +1,146 @@
...
+ #define ADC_RESOLUTION                12
+ #define DACC_RESOLUTION               12
+
+ //PWR resolution
+ #define PWM_RESOLUTION                8
+ #define PWM_FREQUENCY                 1000
+ #define PWM_MAX_DUTY_CYCLE            255
+
+//On-board LED pin number
+#define LED_BUILTIN				21
```
fpistm: replace tab by space

variants/MAPLEMINI_F103CB/PeripheralPins.c
```diff
@@ -0,0 +1,195 @@
...
+  {PB0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC1_IN8
+  // {PB0,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC2_IN8
+  {PB1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC1_IN9
+  // {PB1,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC2_IN9
+  {NC,   NC,    0}
+};
+
+//*** DAC ***
+
+const PinMap PinMap_DAC[] = {
```
fpistm: delete empty array

variants/NUCLEO_F103RB/PeripheralPins.c
```diff
@@ -0,0 +1,213 @@
...
+    // {PC4,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 14, 0)}, // ADC2_IN14
+    {PC5,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC1_IN15
+    // {PC5,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC2_IN15
+    {NC,   NC,    0}
+};
+
+//*** DAC ***
+
+const PinMap PinMap_DAC[] = {
+    {NC,   NC,    0}
```
fpistm: delete empty arry

variants/BLUEPILL_F103C8/PeripheralPins.c
```diff
@@ -0,0 +1,195 @@
...
+  // {PA5,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 5, 0)}, // ADC2_IN5
+  {PA6,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC1_IN6
+  // {PA6,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC2_IN6
+  {PA7,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 7, 0)}, // ADC1_IN7
+  // {PA7,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 7, 0)}, // ADC2_IN7
+  {PB0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC1_IN8
+  // {PB0,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC2_IN8
+  {PB1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC1_IN9
+  // {PB1,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC2_IN9
+  {NC,   NC,    0}
```
fpistm: All
 {NC,   NC,    0}
have to be replace by
 {NC,   NP,    0}

variants/DISCO_F100RB/PeripheralPins.c
```diff
@@ -0,0 +1,206 @@
...
+  {PA7,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 7, 0)}, // ADC1_IN7
+  {PB0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC1_IN8
+  {PB1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC1_IN9
+  {PC0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 10, 0)}, // ADC1_IN10
+  {PC1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 11, 0)}, // ADC1_IN11
+  {PC2,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 12, 0)}, // ADC1_IN12
+  {PC3,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 13, 0)}, // ADC1_IN13
+  {PC4,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 14, 0)}, // ADC1_IN14
+  {PC5,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC1_IN15
+  {NC,   NC,    0}
```
fpistm: All 
{NC, NC, 0}
have to be replaced by
{NC, NP, 0}

variants/MAPLEMINI_F103CB/PeripheralPins.c
```diff
@@ -0,0 +1,195 @@
...
+  // {PA5,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 5, 0)}, // ADC2_IN5
+  {PA6,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC1_IN6
+  // {PA6,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC2_IN6
+  {PA7,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 7, 0)}, // ADC1_IN7
+  // {PA7,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 7, 0)}, // ADC2_IN7
+  {PB0,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC1_IN8
+  // {PB0,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 8, 0)}, // ADC2_IN8
+  {PB1,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC1_IN9
+  // {PB1,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 9, 0)}, // ADC2_IN9
+  {NC,   NC,    0}
```
fpistm: All
{NC, NC, 0}
have to be replaced by
{NC, NP, 0}

variants/NUCLEO_F103RB/PeripheralPins.c
```diff
@@ -0,0 +1,213 @@
...
+    // {PC1,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 11, 0)}, // ADC2_IN11
+    {PC2,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 12, 0)}, // ADC1_IN12
+    // {PC2,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 12, 0)}, // ADC2_IN12
+    {PC3,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 13, 0)}, // ADC1_IN13
+    // {PC3,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 13, 0)}, // ADC2_IN13
+    {PC4,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 14, 0)}, // ADC1_IN14
+    // {PC4,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 14, 0)}, // ADC2_IN14
+    {PC5,   ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC1_IN15
+    // {PC5,   ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 15, 0)}, // ADC2_IN15
+    {NC,   NC,    0}
```
fpistm: All
{NC, NC, 0}
have to be replaced by
{NC, NP, 0}

fpistm:
I think USB CDC could be more generic and available for other variant which support it.
I have already integrate a CDC on this pull:
https://github.com/arduino-org/arduino-core-stm32f4/pull/44

I will do this stuff later. currently, I think we could kept this as it is.

fpistm:
@rogerclarkmelbourne , 
@fprwi6labs will rebase when he will update the pull with review comment.

rogerclarkmelbourne:
One thing I noticed, because you have the BluePill and the Maple mini via a sub menu, and then have all the same upload options or both, 

You end up with the original maple mini bootloader upload as the default option.

But that version of the bootloader is not available for the Blue Pill, so by default it won't work.

BTW. I have tied contacting one of the manufacturers in china who make a maple mini clone (Baite Online Store), to say they should use the new bootloader and also all their links to Leaflabs are now out of date.

But they have not been very communicative, so I don't expect they will change this :-(

rogerclarkmelbourne:
D:\Documents\Arduino\hardware\stm32duino\Arduino_Core_STM32\variants\MAPLEMINI_F103CB/variant.h:19:0: error: unterminated #ifndef

 #ifndef _VARIANT_ARDUINO_STM32_

variants/BLUEPILL_F103C8/usb/usbd_conf.c
```diff
@@ -45,7 +45,6 @@
 /* Private macro -------------------------------------------------------------*/
 /* Private variables ---------------------------------------------------------*/
 PCD_HandleTypeDef hpcd_USB_FS;
-void Error_Handler(void);
```
fpistm: To a future enhancement, maybe it should be fine to add one.

RickKimball:
Native USB doesn't seem to work on a bluepill. It fails to enumerate. I get this message:
```
Jun 30 13:48:26 kimballr kernel: [25020.593703] usb 1-3.4: USB disconnect, device number 44
Jun 30 13:48:28 kimballr kernel: [25022.844067] usb 1-3.4: new full-speed USB device number 45 using ehci-pci
Jun 30 13:48:28 kimballr kernel: [25022.916068] usb 1-3.4: device descriptor read/64, error -32
Jun 30 13:48:28 kimballr kernel: [25023.092045] usb 1-3.4: device descriptor read/64, error -32
Jun 30 13:48:28 kimballr kernel: [25023.268066] usb 1-3.4: new full-speed USB device number 46 using ehci-pci
Jun 30 13:48:28 kimballr kernel: [25023.340070] usb 1-3.4: device descriptor read/64, error -32
Jun 30 13:48:29 kimballr kernel: [25023.516088] usb 1-3.4: device descriptor read/64, error -32
Jun 30 13:48:29 kimballr kernel: [25023.692084] usb 1-3.4: new full-speed USB device number 47 using ehci-pci
Jun 30 13:48:29 kimballr kernel: [25024.100042] usb 1-3.4: device not accepting address 47, error -32
Jun 30 13:48:29 kimballr kernel: [25024.172041] usb 1-3.4: new full-speed USB device number 48 using ehci-pci
Jun 30 13:48:30 kimballr kernel: [25024.580044] usb 1-3.4: device not accepting address 48, error -32
Jun 30 13:48:30 kimballr kernel: [25024.580190] usb 1-3-port4: unable to enumerate USB device
```

Using the same hardware it works fine with the libmaple core:
```
Jun 30 13:45:16 kimballr kernel: [24831.108039] usb 1-3.4: new full-speed USB device number 44 using ehci-pci
Jun 30 13:45:16 kimballr kernel: [24831.203456] usb 1-3.4: New USB device found, idVendor=1eaf, idProduct=0004
Jun 30 13:45:16 kimballr kernel: [24831.203462] usb 1-3.4: New USB device strings: Mfr=1, Product=2, SerialNumber=0
Jun 30 13:45:16 kimballr kernel: [24831.203466] usb 1-3.4: Product: Maple
Jun 30 13:45:16 kimballr kernel: [24831.203469] usb 1-3.4: Manufacturer: LeafLabs
Jun 30 13:45:16 kimballr kernel: [24831.203913] cdc_acm 1-3.4:1.0: ttyACM1: USB ACM device
Jun 30 13:45:16 kimballr mtp-probe: checking bus 1, device 44: "/sys/devices/pci0000:00/0000:00:04.1/usb1/1-3/1-3.4"
Jun 30 13:45:16 kimballr mtp-probe: bus: 1, device: 44 was not an MTP device
```
I'm running ubuntu 16.04 lts, the device is actually a red pill (same as blue pill except it has the appropriate 1k5 pull up on PA12 from the manufacturer).  The device is plugged into a USB 2.0 powered hub.  Native USB has never been a problem on this device. It works properly with STM32CUBEMX configured as a CDC USB device. As it does with libopencm3 acting as a BMP programming device. And finally as I already mentioned libmaple CDC works fine.

RickKimball:
For the BLUEPILL-F103C8 how did you decide to map the pins? They don't match the way the pins are laid out nor do they match the original libmaple simple scheme.

fprwi6labs:
@RickKimball I reordered the pin mapping of the BluePill following the old variant.h from the repository STM32F1. I don't know why I didn't copy it before.
Could you confirm me that is the good one now?

I found my BluePill board so I will check the Native USB.

RickKimball:
Regarding the USB not working.

The BLUEPILL boards don't actually have a USB_DISC_PIN. To simulate a USB reset sequence, we have setting the USBD+ pinmode to an output, setting the value low, delaying for the usb reset period then setting the mode back to an input so that the pull up resistor pulls it high.  Then the nomal USB init sequence is run.  The code as written seems to expects a real USB_DISC_PIN that would have an NPN transistor.  The BLUE PILL does not have any hardware like that.

rogerclarkmelbourne:
I recently had to make a change about this in libmaple

USB_DISC port and pin were defined in board.txt and passed as defined to the compiler.

So I changed the code so that if either of these were null, the code to init and toggle this pin were not executed.

There is a separate define in LibMaple which is also used to toggle USBD+, but I should probably change it, so that the define for USB_DISC port or pin being null is used instead

This does also assume that USB_SERIAL is defined (so that the code knows that USB serial should be compiled in)

Testato:
excuse the intrusion, but i'm here only for remember that Bluepill is the most important variant for the success of this core.
Every maker on every group start he's adventure on ST by a Bluepill.
So it is very important that the upload, the bootloader, etc work simple, at least like on the Roger Arduino_STM32

danieleff:
1. You need to let the build.txt parameter VECT_TAB_OFFSET to override the default
```diff
--- a/system/STM32F1xx/system_stm32f1xx.c
+++ b/system/STM32F1xx/system_stm32f1xx.c
@@ -110,8 +110,10 @@
 /*!< Uncomment the following line if you need to relocate your vector Table in
      Internal SRAM. */ 
 /* #define VECT_TAB_SRAM */
+#ifndef VECT_TAB_OFFSET
 #define VECT_TAB_OFFSET  0x0 /*!< Vector Table base offset field. 
                                   This value must be a multiple of 0x200. */
+#endif
```

2. USB_LP_CAN1_RX0_IRQHandler is needed, either in variants/usb/....c, or USBSerial.cpp

```c++
void USB_LP_CAN1_RX0_IRQHandler(void) {
    HAL_PCD_IRQHandler(&hpcd_USB_FS); 
}
```

rogerclarkmelbourne:
I though we had already added bootloader uploads to the old F1 repo.

One of the files ( headers ) in the HAL had to be mofied, as it did not allow overriding of the vector table offset, but this may have been fixed in newer versions of the Cube, because I know that the Cube HAL files for some other MCUs ,e g the F4 did not need to be modified as they already had the guard , ifdefs around the vector table define.

The linker also needs to be changed to support using the bootloader

RickKimball:
> USB_LP_CAN1_RX0_IRQHandler is needed, either in variants/usb/....c, or USBSerial.cpp
> void USB_LP_CAN1_RX0_IRQHandler(void) {
>     HAL_PCD_IRQHandler(&hpcd_USB_FS); 
> }

This gets it to enumerate, thanks! Now to figure out why it isn't spewing data.

rogerclarkmelbourne:
Rick did you change system_stm32f1xx.c

to match whats in the old F1 repo
e.g. add the guard ifdef 

https://github.com/stm32duino/Arduino_Core_STM32F1/blob/25671e086825821eeeb50d903a6c293fce6515e3/cores/arduino/libstm32f1/system_stm32f1xx.c#L113-L116

RickKimball:
I didn't.I'm using it with an stlink.

fpistm:
> @Testato 
> excuse the intrusion, but i'm here only for remember that Bluepill is the most important variant for the success of this core.
> Every maker on every group start he's adventure on ST by a Bluepill.
> So it is very important that the upload, the bootloader, etc work simple, at least like on the Roger Arduino_STM32

If a core include a variant this is to be functional...
Any help are welcome

RickKimball:
@fprwi6labs 

> @RickKimball I reordered the pin mapping of the BluePill following the old variant.h from the repository STM32F1. I don't know why I didn't copy it before.
> Could you confirm me that is the good one now?

Yes the digital pins seem to work fine:
```
unsigned nextpin = 0;
unsigned long previousMillis = 0;        // will store last time LED was updated
unsigned ledState;

void displayPinInfo(unsigned index) {
  unsigned p = digitalPin[index];
  unsigned port = p >> 4 & 0x0f;
  unsigned pin = p & 0xf;

  Serial.print("PIN "); Serial.print(index); Serial.print(" = P"); Serial.print("ABCDEF"[port]); Serial.println(pin);

}

void setup() {
  Serial.begin(115200);
  while (!Serial);
  Serial.println("Digital Pin Test\r\nPress enter to go to next pin");
  pinMode(nextpin, OUTPUT);
  displayPinInfo(nextpin);
}


void loop() {
  if ( Serial.available() ) {
    int c = Serial.read();

    pinMode(nextpin, INPUT);
    nextpin++;
    switch (nextpin) {
      case 11: // skip USB pins
      case 12:
        nextpin = 13;
        break;
      case 35:
        nextpin = 0;
        break;
    }

    pinMode(nextpin, OUTPUT);
    displayPinInfo(nextpin);
  }

  unsigned long currentMillis = millis();

  if (currentMillis - previousMillis >= 100) {
    // save the last time you blinked the LED
    previousMillis = currentMillis;

    digitalWrite(nextpin, ledState);
    ledState ^= 1;
  }

}
```

RickKimball:
I was looking at the USB CDC code and it seems overly layered and complex. Even worse it seems to be based on code that people have identified as having unresolved locking problems. Do you know the status of the problem described in this post?

https://community.st.com/thread/10392

The problem was identified  in Aug of 2015 and in Feb of 2016 ST said it would be fixed soon.  Since then no fix has appeared.

The USB CDC code here is based on STM32Cube USB CDC yes?  Does it seem wise to rely on a chunk of code that seems suspect?

fpistm:
I have to review USB CDC part, it was validated with Arduino for the arduino_core_stm32f4. I know there is some update in it.

RickKimball:
This code gets stuck:
```
void setup() {
  Serial.begin(9600);
  while (!Serial) {
    ; // wait for serial port to connect. Needed for native USB port only
  }
  Serial.println("ASCII Table ~ Character Map");
}
```
The problem is USBSerial::operator bool() never returns true.  Even worse, it isn't even looking at the line state to see if the host has asserted the DTR signal. This function should return true if a DTR is asserted and false if it isn't
```
USBSerial::operator bool()
{
	// this is here to avoid spurious opening after upload
	if (millis() < 500)
		return false;

	bool result = false;

	return result;
}
```
This brings up another problem with the way the USB CDC code is written.  In 
https://github.com/fprwi6labs/Arduino_Core_STM32/blob/STM32F1/variants/BLUEPILL_F103C8/usb/usbd_cdc_if.c the function   CDC_Control_FS  () should be toggling the dtr_pin variable, not constantly incrementing it.
```
...
  case CDC_SET_CONTROL_LINE_STATE:
    dtr_pin ^= 1; //DTR pin is enabled
    break;
...
```

RickKimball:
Why is there a delay(1) in main.cpp?

https://github.com/fprwi6labs/Arduino_Core_STM32/blob/master/cores/arduino/main.cpp#L38

fprwi6labs:
@RickKimball 

> Why is there a delay(1) in main.cpp?
> https://github.com/fprwi6labs/Arduino_Core_STM32/blob/master/cores/arduino/main.cpp#L38

It is an old artefact from an old Arduino version. Arduino deleted it. We can do the same.

Testato:
On Bluepill we need two board.txt entry, one for the 64KB and one for 128KB of flash memory size

RickKimball:
analogWrite ( the pwm_start version ) seems to introduce a spurious low pulses of about ~5 us in the middle of the PWM signal.

Using the code below, I watched PA7 of a BLUEPILL on my scope and triggered on a pulse of less than 10us.  It should be producing a 500us on 500us off PWM pulse.  However, the brief 5us pulse is there.  If you zoom out sufficiently on your scope it might co unnoticed. 
```
void setup() { }

void loop() {
  analogWrite(7,127);
  delay(100);
}
```
Maybe the analogWrite should check to see if the pin is already in PWM mode and just change the duty cycle instead of completely re-initializing the timer settings every time analogWrite is called. This seems to be the cause of the low pulse as it disables/enables the timer.

Maybe the HAL macro __HAL_TIM_SET_COMPARE() could be used to just change the duty cycle when the frequency isn't changing (the case of a simple repeated analogWrite() call)

RickKimball:
Do you want us to create pull requests for the problems above or would you prefer to look at the problem and write a solution yourself?

fpistm:
Hi @RickKimball,

There is a lot of feedback.
I'm wondering if the pull should be split in several one.
The nucleo F103 is ok.
There are issue on other F103 based board (BP, MM), mainly around usb.
Probably that an issue on the PWM is a good idea, PWM needs some review, I think, as it seems there is  other issue (resolution).

Adminius:
if you change analogWrite function, would it be possible to add possibility to change **TIM_COUNTERMODE_UP** to i.e. **TIM_CounterMode_CenterAligned1**  ?
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/stm32/analog.c#L729
It would be enough if user can call init function manually for each output or for all ouputs like **analogResolution**.

fpistm:
@Testato 
I  will open an new issue about that to track it:
> On Bluepill we need two board.txt entry, one for the 64KB and one for 128KB of flash memory size
Goal of this PR is to have the same status than older Arduino_Core_STM32F1

fpistm:
First part merged:
https://github.com/stm32duino/Arduino_Core_STM32/pull/60
Second part will be added soon

fpistm:
Please made further comment on this PR: #61

PR #38, opened 2017-06-23
Stm32 ethernet lib

fprwi6labs:
NativeEthernet library added to allow STM32 boards with Ethernet feature to connect to the internet.
More information [here](https://github.com/stm32duino/wiki/wiki/NativeEthernet).

NUCLEO-F429ZI is ready to use the NativeEthernet library.

PR #52, opened 2017-07-04
Ethernet f4

fprwi6labs:
This PR must replace #38 because we isolated the STM32Ethernet & LwIP libraries in independent repositories.

This PR adds Ethernet feature for all cores and Ethernet configuration for NUCLEO-F429ZI variant.


cores/arduino/main.cpp
```diff
@@ -45,6 +45,9 @@ int main( void )
 
 	for (;;)
 	{
+		// Define by Ethernet library. It is defined as __weak.
+		stm32_eth_scheduler();
```
fpistm: Think about add registration callback mechanism during instance init. This way it was more generic and usable for further lib

Issue #56, opened 2017-07-06
NUCLEO F401RE support


Adminius:
first try:
https://github.com/Adminius/Arduino_Core_STM32/commit/0646d9167f8ab9755a30b91fd9275436c3d7b296

i used CubeF4 to generate some files. but i don't kwon if it was right settings

i can compile and upload, but the board do nothing. (blink doesn't blink)
any tips?

Adminius:
http://stm32duino.com/viewtopic.php?f=48&t=2304&p=31082

Adminius:
tested with auto genereted file, same issue
https://github.com/Adminius/Arduino_Core_STM32/commit/9235195e70490512397734388ce5ea24af037c25

fpistm:
Hi @Adminius,

it seems this is your clock config which is not correct.

Try this one, with it LED is blinking ;):
```
/**
  * @brief  System Clock Configuration
  *         The system Clock is configured as follow : 
  *            System Clock source            = PLL (HSI)
  *            SYSCLK(Hz)                     = 84000000
  *            HCLK(Hz)                       = 84000000
  *            AHB Prescaler                  = 1
  *            APB1 Prescaler                 = 2
  *            APB2 Prescaler                 = 1
  *            HSI Frequency(Hz)              = 16000000
  *            PLL_M                          = 16
  *            PLL_N                          = 336
  *            PLL_P                          = 4
  *            PLL_Q                          = 7
  *            VDD(V)                         = 3.3
  *            Main regulator output voltage  = Scale2 mode
  *            Flash Latency(WS)              = 2
  * @param  None
  * @retval None
  */
WEAK void SystemClock_Config(void)
{
  RCC_ClkInitTypeDef RCC_ClkInitStruct;
  RCC_OscInitTypeDef RCC_OscInitStruct;

  /* Enable Power Control clock */
  __HAL_RCC_PWR_CLK_ENABLE();

  /* The voltage scaling allows optimizing the power consumption when the device is
     clocked below the maximum system frequency, to update the voltage scaling value
	 regarding system frequency refer to product datasheet.  */
  __HAL_PWR_VOLTAGESCALING_CONFIG(PWR_REGULATOR_VOLTAGE_SCALE2);

  /* Enable HSI Oscillator and activate PLL with HSI as source */
  RCC_OscInitStruct.OscillatorType = RCC_OSCILLATORTYPE_HSI;
[...]

PR #64, opened 2017-07-13
Replaced hard coded EEPROM base address with generic code

fprwi6labs:
Patch following a problem encountered in #56. 
Hard coded addresses has been replaced with more generic code definition.

@fpistm the F7 and F2 need to be tested! Other are OK.


fprwi6labs:
Modifcation done!

PR #79, opened 2017-08-14
I2C update following new STM32F1 HAL release

fprwi6labs:
The new STM32F1xx HAL release allow to remove specific set up for STM32F1 boards inside I2C core and library.
The specific examples for F1 has been reworked too.


fpistm:
@fprwi6labs Ok to merge.

PR #84, opened 2017-08-19
Full Alternate function management for F1 series

fpistm:
This add all AFIO REMAP for STM32 F1 series


RickKimball:
I gave this a try on the DISCO_F100RB, no difference, Serial.print() works find but Serial.available() never gets any data. I went into debug. Again it seems to enter USART2_IRQHandler for output but never for input. 

This works now thanks to your pin map changes, thanks!
pinMode(PC8,OUTPUT); 

One thing I did notice is that the DISCO_F100RB is using the HSI. This board has both an 8MHz HSE and a 32K xtal.

I put a scope on the pin and we are back to the situation where the TX pin works and the RX pin doesn't In addition, the pin signal level.

I think the problem is that the RX pin isn't being set to input.  I used the debugger to stop in uart_init
When I looked at the GPIO_InitStruct of the UART_RX pin just before it called HAL_GPIO_Init()
the mode was wrong. I used to debugger to fix it .. setting Mode to 0 instead of 2

```uart_init (obj=obj@entry=0x20000218 <Serial+148>) at /home/kimballr/Arduino/hard
ware/st/stm32_alt/cores/arduino/stm32/uart.c:224
(gdb) p GPIO_InitStruct 
$12 = {Pin = 8, Mode = 2, Pull = 0, Speed = 3}
(gdb) set GPIO_InitStruct.   
Display all 200 possibilities? (y or n)
(gdb) set GPIO_InitStruct.Mode = 0
(gdb) p GPIO_InitStruct 
$13 = {Pin = 8, Mode = 0, Pull = 0, Speed = 3}
```
I let the debugger continue and it magically started working.216

I think Line 216 in uart_init.c should be setting Mode to 0 
```
 GPIO_InitStruct.Mode        = STM_PIN_MODE(pinmap_function(obj->pin_rx,PinMap_UART_RX));
``` 
However, It is getting set to 2

fpistm:
Hum interesting.
After checking it seems that for F1 in [PeripheralPins.c](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/variants/DISCO_F100RB/PeripheralPins.c#L146), we need to set:
`STM_MODE_INPUT`
instead of 
`STM_MODE_AF_PP`

See the Mbed one for Nucleo-F103RB:
https://developer.mbed.org/users/mbed_official/code/mbed-dev/file/default/targets/TARGET_STM/TARGET_STM32F1/TARGET_NUCLEO_F103RB/PeripheralPins.c
I need to understand why but probably linked to specific AF...
I will do the test with Nucleo-F103RB.
Thanks @RickKimball  for your time and sorry to not saw this before....

fpistm:
For F1, AF implies to configure as INPUT. For Other series, AF is a specific GPIO config which handle Input or Output depending of the AF selected.
I will update PeripheralPins.c for UART RX and Can RD.

PR #91, opened 2017-08-23
USB reworked

fprwi6labs:
USB configuration is more generic and free of the core.
It is easier to import new USB class from a library.

HIDComposite has been moved inside an independent  library used by Keyboard and Mouse libraries.
USBSerial has been moved inside is own library too.

USBSerial must be tested with BluePill and Maple Mini boards.

"USB interface" menu is now only used to enable USB.

Suggestions are welcome.


RickKimball:
Are you going to add BluePill and MapleMini to this or do I have to manually merge those changes?

fpistm:
@RickKimball goal is to have all merged. This PR is a first proposal version from Fabien to manage USB features: 

- HID
- CDC 
- USBMicrophone (which is ongoing and will be available here [USBMicrophone](https://github.com/stm32duino/USBMicrophone)

Still to include DFU.

I have to benchmark and review this PR to see if it could be applicable. Mainly around USB performance as in @rogerclarkmelbourne core several works have been done to enhance it.
FYI Fabien is out of office during 1 week so don't worry if he will not answer during this period.

fprwi6labs:
Done! I add src directory in libraries.

fprwi6labs:
We have the solution to propose this library by default (not an external library) and include USBSerial.h in the Arduino.h. And add the flag USBCON to avoid compilation when USB is not used. What do you think?

chwon64:
I am using 2017.9.22 release. This version doesn't support USB for STML0 CPU.
Do you have plan when it will be supported?

fpistm:
Hope soon, currently working on it as a background task.

chwon64:
That's great. Thanks for quick response.

hei5enbrg:
is usb-cdc working for stm32f411 in the current release?

fpistm:
Currently not. It will be one of the next feature to add

hei5enbrg:
OK. Even mbed lacks, last time I checked.
I'll have to use stmcube APIs then. Thanks

fpistm:
In fact, it was functionnal, but required some clean. First will be to rebsse it properly

libraries/USBSerial/src/USBSerial.h
```diff
@@ -0,0 +1,80 @@
...
+    virtual int peek(void);
+    virtual void flush(void);
+    virtual int read(void);
+
+    int availableForWrite(void);
+    virtual size_t write(uint8_t);
+    inline size_t write(unsigned long n) { return write((uint8_t)n); }
+    inline size_t write(long n) { return write((uint8_t)n); }
+    inline size_t write(unsigned int n) { return write((uint8_t)n); }
+    inline size_t write(int n) { return write((uint8_t)n); }
```
matthijskooijman: Why all these `write` overloads? The official Arduino AVR core's CDC serial (and, AFAIK also other Stream subclasses) do not have these overloads: https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/USBAPI.h#L103-L104

Also, since integer types are implicitly convertable to uint8_t there is no need for them (as long as there is only single-argument version of write, which is how the Stream class was intended).

fpistm: Hi @matthijskooijman 
this was a legacy from https://github.com/arduino-org/arduino-core-stm32f4
https://github.com/arduino-org/arduino-core-stm32f4/blob/master/cores/arduino/usb_serial.h#L59

fpistm: Right,
in the new PR for USB, I'm inline with CDC form Arduino

PR #83, opened 2017-08-17
Pin name PXn could be used instead of Dx

fpistm:
With this PR the following call are equivalent (for Nucleo F429ZI):

**Digital:**
```
digitalWrite(33, HIGH); 
digitalWrite(D33, HIGH);
digitalWrite(PB0, HIGH);
digitalWrite(pinNametoDigitalPin(PB_0), HIGH);
```
**AnalogRead:**
```
analogRead(A0);
analogRead(0);
analogRead(PA3);
analogRead(pinNametoDigitalPin(PA_3));
analogRead(D78);
analogRead(78);
```

**AnalogWrite:**
```
analogWrite(A0);
analogWrite(PA3);
analogWrite(pinNametoDigitalPin(PA_3));
analogWrite(D78);
analogWrite(78);
```
Warning `analogWrite(0);` write on D0 (PG9)

PR #100, opened 2017-09-05
Fix SPI CS pin trouble

fprwi6labs:
I fixed a bug inside SPI library caused by the use of a CS pin other than Arduino digital pins.
Now, all pins can be used like a CS pin for the SPI.


variants/DISCO_F100RB/variant.h
```diff
@@ -133,7 +133,7 @@ enum {
 
 //SPI definitions
 //define 16 channels. As many channel as digital IOs
-#define SPI_CHANNELS_NUM        16
+#define SPI_CHANNELS_NUM        PEND
```
fpistm: Pleaser use NUM_DIGITAL_PINS instead of PEND for all
Amend the commit to avoid twice commit for this

fprwi6labs: Done!

fpistm: Thanks! ;)

fpistm:
Thanks @fprwi6labs 
But this will increase memory consumption:
https://github.com/stm32duino/Arduino_Core_STM32/blob/06921778e08936d1fac10ebf210609414c3b13d0/libraries/SPI/SPI.h#L163
I think, we need to study more before merge.

LMESTM:
@fprwi6labs can you please describe the original root cause of the bug and how this is fixed here ?   

For now please find my 2 cents as I just discussed it with @fpistm 
As I read it, SPI_CHANNELS_NUM[] could be used as a table of CS pins that can be used on the SPI bus. You can have several devices on the same bus, but each of them has its own CS pin. So I would think that there is 1 device  per channel and the table is used to store the CS pin. 
But you should not need to increase the maximum number of channels higher than 16, you just need to be able to store the pin number in the table, and the pin number needs to be handled as a generic pin (can be D2 or PH3 or PA12 ...)

fprwi6labs:
@LMESTM 

Firstly, the bug is caused by this code:
`if(_pin > SPI_CHANNELS_NUM)
     return;`

So when you use a CS pin higher than SPI_CHANNELS_NUM you are out.
The solution was to increase SPI_CHANNELS_NUM to the maximum digital pin number.
And @fpistm points a side effect of my fix.

Secondly, when you look at the avr code, the CS pin is saved only one time when you call the function begin(). So I think it is not to the SPI library to save a list of CS pin but to the user if he has several devices on the same SPI bus.

Finally, spiSettings saved the SPI settings but it is 'idiot' to save multiple settings for the same instance. You should call beginTransaction() to set the SPI before to send data.

I exposed my project to @fpistm to rework the SPI library. I will propose to you soon the code.

LMESTM:
One more note: I just add a look to SPI.cpp and SPI.h files and the _pin is used as an index to the spiSettings[] table, so the _pin is directly used as an channel index. To be able to use any pin as CS, the table index must be separated from the _pin ...  so a deeper change is needed

LMESTM:
> Firstly, the bug is caused by this code:
if(_pin > SPI_CHANNELS_NUM) return;

I found out indeed

> The solution was to increase SPI_CHANNELS_NUM to the maximum digital pin number.
And @fpistm points a side effect of my fix.

Yes. I agree this is due to SPI implementation which uses _pin as the index of settings.

> Secondly, when you look at the avr code, the CS pin is saved only one time when you call the function begin(). So I think it is not to the SPI library to save a list of CS pin but to the user if he has several devices on the same SPI bus.

I agree the user will store and use the CS pin, which is used as a reference to a device.

> Finally, spiSettings saved the SPI settings but it is 'idiot' to save multiple settings for the same instance. You should call beginTransaction() to set the SPI before to send data.

I don't think this is 'idiot'.  You can have several devices on the SPI bus, on different CS pins, that have different settings. And you don't want to re-define the settings before each transfer, so the SPI library do the job of applying the right settting. This is efficient: If it's the same device as before, you keep the setting, and only when device changes you re-apply the corresponding setting.  I'm not saying that you must store the _pin, but you'd need to find a way to map the table index to a given _pin ...

fprwi6labs:
> I don't think this is 'idiot'. You can have several devices on the SPI bus, on different CS pins, that have different settings. And you don't want to re-define the settings before each transfer, so the SPI library do the job of applying the right settting. This is efficient: If it's the same device as before, you keep the setting, and only when device changes you re-apply the corresponding setting. I'm not saying that you must store the _pin, but you'd need to find a way to map the table index to a given _pin ...

I agree but when I read the documentation of the Arduino SPI I understand that the user should follow the following sequence:
- call beginTransaction() to set the bus
- call transfer() to send data
- call endTransaction() to release the bus

The native Arduino SPI library is simpler than our. But I can try to save several settings.

libraries/SPI/src/SPI.h
```diff
@@ -144,26 +158,90 @@ class SPIClass {
...
-    //uint32_t id;
-    SPISettings spiSettings[SPI_CHANNELS_NUM+1];
-    int8_t g_active_id;
-    uint8_t _pin;
-    spi_t _spi;
+    spi_settings_t  spiSettings[NB_SPI_SETTINGS];
+    int16_t         _CSpin;
+    spi_t           _spi;
+
+    uint8_t getIdx(uint8_t _pin)
```
fpistm: Merge getIdx and addIdx and use a second arg to define if a new idx could be added with a default value (no add by default)
This will avoid to loop twice if there is no idx for pin then call addIdx.

libraries/SPI/src/SPI.cpp
```diff
@@ -78,73 +94,100 @@ void SPIClass::usingInterrupt(uint8_t interruptNumber)
...
+
+  if(_pin > NUM_DIGITAL_PINS)
     return;
 
-  spiSettings[_pin].clk = settings.clk;
-  spiSettings[_pin].dMode = settings.dMode;
-  spiSettings[_pin].bOrder = settings.bOrder;
-  if(spiSettings[_pin].bOrder == MSBFIRST) {
-    spiSettings[_pin].msb = MSBFIRST;
+  idx = getIdx(_pin);
```
fpistm: See my comment in SPI.h for getIdx and AddIdx to avoid loop  twice

libraries/SPI/src/SPI.h
```diff
@@ -97,45 +97,59 @@ class SPISettings {
     friend class SPIClass;
 };
 
-
+typedef struct {
+  int16_t pinCS;
```
fpistm: Why not add the pinCS in the SPISettings class?

libraries/SPI/src/SPI.h
```diff
@@ -55,6 +44,17 @@ enum SPITransferMode {
...
 };
 
+// Indicates the user controls himself the CS pin
+#define CS_PIN_CONTROLLED_BY_USER  NUM_DIGITAL_PINS
+
+// Defines a default timeout delay in seconds for the SPI transfer
+#define SPI_TRANSFER_TIMEOUT    5000
+
+// Defines the number of settings saved per SPI instance. Must be < 254.
+#ifndef NB_SPI_SETTINGS
```
fpistm: Add in the comment, it could be redefined in the variant.h

libraries/SPI/src/SPI.cpp
```diff
@@ -78,73 +94,100 @@ void SPIClass::usingInterrupt(uint8_t interruptNumber)
...
+  if(idx == NB_SPI_SETTINGS) {
+    idx = addPin(_pin);
+    if(idx == NB_SPI_SETTINGS)
+      return;
```
LMESTM: this case here is not clear - is this an error ?

libraries/SPI/src/SPI.cpp
```diff
@@ -78,73 +94,100 @@ void SPIClass::usingInterrupt(uint8_t interruptNumber)
...
+  _CSpin = _pin;
 }
 
-void SPIClass::end(uint8_t _pin)
+void SPIClass::endTransaction(uint8_t _pin)
 {
-  UNUSED(_pin);
-  end();
+  if(_pin > NUM_DIGITAL_PINS)
+    return;
```
LMESTM: same as above

libraries/SPI/src/SPI.cpp
```diff
@@ -167,39 +215,49 @@ void SPIClass::setClockDivider(uint8_t _pin, uint8_t _divider)
...
 
 
-//Transfer a message on the selected SPI. The _pin is the CS of the SPI that
-//identifies the SPI instance.
-//If the _mode is set to SPI_CONTINUE, keep the spi instance alive.
+/* Transfer a message on the selected SPI. The _pin is the CS.
+ * The transfer function can reconfigure the SPI instance if the CS pin is
+ * different from the previous one.
+ * If the _mode is set to SPI_CONTINUE, keep the spi instance alive. That means
+ * the CS pin is not reset. Becareful in case you use several CS pin.
```
LMESTM: Good comment thx. 1 typo:  Becareful => Be careful

libraries/SPI/src/SPI.h
```diff
@@ -55,6 +44,17 @@ enum SPITransferMode {
   SPI_LAST
 };
 
+// Indicates the user controls himself the CS pin
+#define CS_PIN_CONTROLLED_BY_USER  NUM_DIGITAL_PINS
+
+// Defines a default timeout delay in seconds for the SPI transfer
+#define SPI_TRANSFER_TIMEOUT    5000
+
```
LMESTM: This seems not related to CS related change and should be in a different commit.
The timeout was hardcoded to 10000 before and now is defined as 5000. 
Is it really 5000 seconds ??   Do we consider that timeout can be fixed, whatever the number of bytes and frequency ?

variants/DISCO_F100RB/variant.h
```diff
@@ -132,16 +132,7 @@ enum {
...
-#define SPI_CHANNELS_NUM        16
-
-//default chip salect pin
-#define BOARD_SPI_DEFAULT_SS    43
-
-//In case SPI CS channel is not used we define a default one
-#define BOARD_SPI_OWN_SS        SPI_CHANNELS_NUM
-
-#define SS                      BOARD_SPI_DEFAULT_SS
+#define SS                      43
```
LMESTM: Not related to this commit, but was is 43 ? can't we use a pin name like e.g. D10 or PB13 ?

fpistm: 43 == D43

libraries/SPI/src/SPI.cpp
fprwi6labs: Yes this is an error because there is no more place available for a new CS pin.

libraries/SPI/src/SPI.cpp
fprwi6labs: This is also an error because _pin must be a defined pin (presents in digitalPin[]).

libraries/SPI/src/SPI.h
fprwi6labs: It is not 5000 seconds but 5000 milliseconds. I took advantage of the fix to standardize the timeout value. 
This timeout doesn't change whatever the number of bytes or frequency.

variants/DISCO_F100RB/variant.h
fprwi6labs: See PR #83 for more information.

libraries/SPI/src/SPI.h
LMESTM: The comment above the definition is
// Defines a default timeout delay in seconds for the SPI transfer
so you need to change that ... 
How many bytes does it allow to send @ slowest supported frequency ?

variants/DISCO_F100RB/variant.h
LMESTM: I understand that D43 is same as 43 and same as PX_y ...but D43 or 43 cannot be read anywhere on my board schematics or on the board itself, so I would prefer to use a commonly used names (either  D0/D16 A0/A5 or PX_y ....

fpistm: Agree that on VLdisco there is no Arduino connector and so using Dx is not well appropriate while all PXn names are on the board silkscreen print. But it's not the subject of this PR. Currently, variant.h using the same naming (Dx). Probably a new PR could be done to update all variants which do not have Arduino connector or those one using a pin number > 15.

LMESTM: I agree, that's why I started my comment with "Not related to this commit" ;-)

libraries/SPI/src/SPI.h
fprwi6labs: Unless I am mistaken, we can send 78kbytes at slowest frequency = 125000.

LMESTM: ok that sounds like a reasonable number. Even 1 sec might be enough then (~15kBytes) ...

Issue #111, opened 2017-09-23
TIMER_TONE with value of TIM14 or TIM15 do not seem to work


ekawahyu:
I am porting the core to work with STM32F072RB Discovery Kit. Everything seems to be working just fine, except that tone() does not generate any sound with TIM14 or TIM15. I haven't done any test with all of timers, but a quick test with TIM3 seemed to generate some tone.

Related thread:
http://www.stm32duino.com/viewtopic.php?f=49&t=2614&p=34733#p34733

fpistm:
Reproduced on Nucleo-F091RC.
@fprwi6labs  any clues on this issue?

fprwi6labs:
Reproduced too on Nucleo-F091RC. TIM1 failed too. 
IRQ handler for timer 15,16,17, ... are missing. But it do not explain why TIM14 or TIM1 don't work.

fprwi6labs:
@fpistm the IRQHandler names are wrong for F0. I will fix them.

fpistm:
Thanks @fprwi6labs 
I've forgot to add those IRQ handler and update name which is different per series...

fpistm:
#112 merged

PR #112, opened 2017-09-25
Add missing timer IRQHandler and fix wrong timer IRQHandler

fprwi6labs:
Fix issue #111

### DIFF
--- a/cores/arduino/stm32/timer.c
+++ b/cores/arduino/stm32/timer.c
@@ -1029,69 +1029,39 @@ void attachIntHandle(stimer_t *obj, void (*irqHandle)(stimer_t *))
 /*                            TIMx IRQ HANDLER                                */
 /******************************************************************************/
 
+#if defined(TIM1_BASE)
 /**
-  * @brief  TIM1 & TIM10 irq handler
+  * @brief  TIM1 IRQHandler common with TIM10 and TIM16 on some STM32F1xx
   * @param  None
   * @retval None
   */
-void TIM1_UP_TIM10_IRQHandler(void)
+void TIM1_IRQHandler(void)
 {
   if(timer_handles[0] != NULL) {
     HAL_TIM_IRQHandler(timer_handles[0]);
   }
 
+#if defined(STM32F1xx) || defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
+#if defined (TIM10_BASE)
   if(timer_handles[9] != NULL) {
     HAL_TIM_IRQHandler(timer_handles[9]);
   }
+#endif
+#endif
+
+#if defined(STM32F1xx) || defined(STM32F3xx) || defined(STM32L4xx)
+#if defined (TIM16_BASE)
+  if(timer_handles[15] != NULL) {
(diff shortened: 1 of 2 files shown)


cores/arduino/stm32/timer.c
```diff
@@ -1029,199 +1029,343 @@ void attachIntHandle(stimer_t *obj, void (*irqHandle)(stimer_t *))
...
 /******************************************************************************/
 
+#if defined(TIM1_BASE)
 /**
-  * @brief  TIM1 & TIM10 irq handler
+  * @brief  TIM1 IRQHandler
   * @param  None
   * @retval None
   */
+#if defined(STM32F0xx)
```
fpistm: Why not use TIM1_IRQHandler and define the proper one like the other (#define TIM1_IRQHandler TIM1_BRK_UP_TRG_COM_IRQHandler ...) in timer.h
in order to be homogeneous?

cores/arduino/stm32/timer.c
```diff
@@ -1029,199 +1029,343 @@ void attachIntHandle(stimer_t *obj, void (*irqHandle)(stimer_t *))
...
+void TIM1_UP_TIM16_IRQHandler(void)
+#elif defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
 void TIM1_UP_TIM10_IRQHandler(void)
+#endif
 {
   if(timer_handles[0] != NULL) {
     HAL_TIM_IRQHandler(timer_handles[0]);
   }
 
+#if defined(STM32F1xx) || defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
```
fpistm: So this one could be removed as handle by TIM10_IRQHandler

cores/arduino/stm32/timer.c
```diff
@@ -1029,199 +1029,343 @@ void attachIntHandle(stimer_t *obj, void (*irqHandle)(stimer_t *))
...
   if(timer_handles[9] != NULL) {
     HAL_TIM_IRQHandler(timer_handles[9]);
   }
+#endif
+
+#if defined(STM32F1xx) || defined(STM32F3xx) || defined(STM32L4xx)
```
fpistm: So this one could be removed as handle by TIM16_IRQHandler

cores/arduino/stm32/timer.c
fprwi6labs: I tried but when timer 10 or 16 are present, there is an error about redefinition:
'error: redefinition of 'TIM1_UP_TIM16_IRQHandler'

fpistm: On which MCU ?

however, for TIM10 handler, it seems strange for STM32F1xx
```
+#elif defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
 void TIM1_UP_TIM10_IRQHandler(void)
```
but 
```
+#if defined(STM32F1xx) || defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
   if(timer_handles[9] != NULL) {
     HAL_TIM_IRQHandler(timer_handles[9]);
```

PR #114, opened 2017-09-26
initialize state before calling HAL_XXX_Init for SPI and I2C

VVESTM:
If we use a dynamic instanciation, the state is not well
initialized. This can ends with a blocking or strange
behaviour.


cores/arduino/stm32/spi_com.c
```diff
@@ -371,6 +371,8 @@ void spi_init(spi_t *obj, uint32_t speed, spi_mode_e mode, uint8_t msb)
   }
 #endif
 
+  handle->State = HAL_SPI_STATE_RESET;
```
fpistm: Could you move this line in the  `  /* Fill default value */` (start at line number 222)

VVESTM:
New version following @fpistm remarks

PR #115, opened 2017-09-27
Use interrupt mode for I2C transfers

VVESTM:
I faced an issue while coding a demo for L475 IoT board that use several sensors on i2c2 and BTLE on SPI3. With the actual polling mode for I2C, the transfers can be disturbed by SPI. 

Commit message : 
  Switch from polling mode to IT mode for I2C transfers.
  For example, this is needed if we want to use SPI and
  I2C in the same time.

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -255,14 +255,12 @@ void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, u
 
   handle->State = HAL_I2C_STATE_RESET;
 
-  if(master == 0) {
-    HAL_NVIC_SetPriority(obj->irq, 0, 1);
-    HAL_NVIC_EnableIRQ(obj->irq);
+  HAL_NVIC_SetPriority(obj->irq, 0, 1);
+  HAL_NVIC_EnableIRQ(obj->irq);
 #ifdef STM32F1xx
-    HAL_NVIC_SetPriority(obj->irqER, 0, 1);
-    HAL_NVIC_EnableIRQ(obj->irqER);
+  HAL_NVIC_SetPriority(obj->irqER, 0, 1);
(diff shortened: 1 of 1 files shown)


cores/arduino/stm32/twi.c
```diff
@@ -338,7 +336,10 @@ i2c_status_e i2c_master_write(i2c_t *obj, uint8_t dev_address,
   HAL_StatusTypeDef status = HAL_OK;
 
   // Check the communication status
-  status = HAL_I2C_Master_Transmit(&(obj->handle), dev_address, data, size, I2C_TIMEOUT_TICK);
+  status = HAL_I2C_Master_Transmit_IT(&(obj->handle), dev_address, data, size);
+
+  // wait for transfer completion
+  while(HAL_I2C_GetState(&(obj->handle)) != HAL_I2C_STATE_READY){};
```
fprwi6labs: Maybe add a timeout in case of error.

cores/arduino/stm32/twi.c
```diff
@@ -380,9 +381,11 @@ i2c_status_e i2c_master_read(i2c_t *obj, uint8_t dev_address, uint8_t *data, uin
 {
   i2c_status_e ret = I2C_ERROR;
 
-  if(HAL_I2C_Master_Receive(&(obj->handle), dev_address, data, size, I2C_TIMEOUT_TICK) == HAL_OK) {
+  if(HAL_I2C_Master_Receive_IT(&(obj->handle), dev_address, data, size) == HAL_OK) {
     ret = I2C_OK;
   }
+  // wait for transfer completion
+  while(HAL_I2C_GetState(&(obj->handle)) != HAL_I2C_STATE_READY){};
```
fprwi6labs: ditto

cores/arduino/stm32/twi.c
LMESTM: Either check for a timeout indeed or maybe at least cehck for possible errors

PR #120, opened 2017-10-02
Remove I2C buffer length limitation required by M24SR64-Y library

fprwi6labs:
The [M24SR64-Y library](https://github.com/stm32duino/M24SR64-Y/pull/1) needs an I2C buffer of 255 bytes. 
So we must to give the possibility to increase the I2C buffer size.


fprwi6labs:
#131 should remove this PR. I have to rework the NFC library to use the new methods of the Wire library.

PR #121, opened 2017-10-04
Fixed included headers

fpistm:
This PR aims at cleaning header file include.
Before some issue/warning was raised depending of which header was included or the order.
Ex: define in the variant.h was defined or not depending of include use.

With this PR, it is now ok.


fpistm:
I've tested some of them without any issue.
I've checked include of the library while they include Arduino.h it is ok

fprwi6labs:
I will remove the variant.h if included in the current libraries on development and only use Arduino.h.
This PR is OK for me.

Issue #124, opened 2017-10-04
Add new method to Wire library


fprwi6labs:
It seems necessary to add a new method to the Wire library to avoid buffer limitation.
We could create a method which pass directly a buffer of data to send directly to the i2c core without to use the `write()` method several time:
`uint8_t write(uint8_t *buffer, uint16_t size);`  shouldn't call `write(data[i])` but save and pass the pointer  to `endTransmission()`.

LMESTM:
This would definitely improve performances - possibly significantly

fprwi6labs:
Same issue for `requestFrom()`. We could add a method to pass a buffer from the user layer and not use the internal `rxBuffer[]`.

bperrybap:
I doub't it will add much, if any, i/o performance given that Wire runs as 100Khz by default.
At the higher speeds like at 1M, there might be some gain, but even then it would likely only benefit larger transfers like when using an EEPROM/NVRAM.
I think before anything like this were added to the API, that it should be prototyped and actually measured with a logic analyzer to see just what benefits, if any, you actually get.

Now from a functional perspective, if the new functions could handle arbitrary transfer sizes by doing multiple smaller sized blocks to fully send/receive larger blocks of data on behalf of the application, that would be a very beneficial feature.
i.e. you could read/write large blocks of data to an EEPROM vs having to do all the fragmentation into 32 byte requests up in the application.

fprwi6labs:
In fact, increase performance is not the goal of this issue.

The goal of this issue is to avoid a problem of memory usage. A size of 32 bytes in the internal buffer (inside the library) is enough for most of applications but sometime you need to exchange more than 32 bytes in one time. So to not increase the memory usage (when not needed), a new method is added to use a buffer with a size known by the application.

The STM32 HAL is able to manage a large number of bytes but not the Wire library without the patch #131

fpistm:
Currently, I've an issue with an I2C eeprom with this patch on lower address. I'm searching why.

PR #131, opened 2017-10-11
Add new methods to Wire library

fprwi6labs:
This solves issue #124 
The library can now use an external buffer to write and read data in master mode. 
The internal buffer size limitation (BUFFER_LENGTH) is no more a problem (#120) because we don't use in this case the internal buffers. 

Note: when you use `requestFrom()` with an external buffer, do not use `available()` and `read() ` functions to read the data. `requestFrom()` will return a not null value if data are available and the data are stored in the external buffer.

I took advantage of the fix to upgrade the library with the Arduino libraries rules.


cores/arduino/stm32/twi.c
```diff
@@ -367,6 +367,11 @@ void i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint8_t size)
 {
   uint8_t i = 0;
 
+  // Protection to not override the TxBuffer
+  if(size > I2C_TXRX_BUFFER_SIZE) {
+    size = I2C_TXRX_BUFFER_SIZE;
+  }
+
```
LMESTM: Shouldn't we rather return en error?
Otherwise application would get an incomplete frame and will not get an error ...

libraries/Wire/src/Wire.cpp
```diff
@@ -259,20 +293,23 @@ size_t TwoWire::write(uint8_t data)
...
+  */
+size_t TwoWire::write(const uint8_t *pdata, size_t quantity)
 {
   if(transmitting){
   // in master transmitter mode
-    for(size_t i = 0; i < quantity; ++i){
-      write(data[i]);
-    }
+    txBufferLength = quantity;
+    pWriteData = (uint8_t*)pdata;
```
LMESTM: before the patch, write() was called and transmission happens.
Now only 2 variables are updated - how does transmission actually start ?

cores/arduino/stm32/twi.c
fprwi6labs: I quickly patched a possible memory leak so I kept the function "void". 
I can return 0 in case of error if you rather.

libraries/Wire/src/Wire.cpp
fprwi6labs: I think you confused master and slave mode. I only change the master mode.
The transmission only happens when endtransmission() is called (before and now).

LMESTM: well what I see here is that you remove the lines:
-    for(size_t i = 0; i < quantity; ++i){
 -      write(data[i]);
 -    }
and replace them by txBufferLength and pWriteData assignment ...  it looks like changing the dynamic

cores/arduino/stm32/twi.c
LMESTM: What I mean is that you need to properly manage errors.

libraries/Wire/src/Wire.cpp
LMESTM: well what I see here is that you remove the lines:
for(size_t i = 0; i < quantity; ++i){
 write(data[i]);
}
and replace them by txBufferLength and pWriteData assignment ... it looks like changing the dynamic

fprwi6labs: Of course I changed the dynamic because I pass directly the data buffer pWriteData to the endTransmission() function (and so to the HAL) and I don't use anymore the internal buffer (low size) which is filled in the write() function.

cores/arduino/stm32/twi.c
fprwi6labs: Done!

libraries/Wire/src/Wire.cpp
LMESTM: Ok - I got confused by write() function as I thought it would actually transmit data on I2C lines. Thanks for the clarification.

fpistm: In fact, @LMESTM is right. the dynamic is not the same and so the txbuffer construction is broken.
I've seen the issue with test on i2C eeprom. See code below:
```
        Wire.beginTransmission(ctrlByte);
        if (_nAddrBytes == 2) Wire.write( (byte) (addr >> 8) );   //high addr byte
        Wire.write( (byte) addr );                                //low addr byte
        Wire.write(values, nWrite);
        txStatus = Wire.endTransmission();
```

The 2 first calls of `Wire.write()` put the bytes in the internal `txBuffer` but when `Wire.write(values, nWrite);` is called, it initialize `pWriteData`.
Finally when `Wire.endTransmission();` is called:
```
    // Select the buffer where are stored the data to send
    if(pWriteData != NULL) {
      data = pWriteData;
    } else {
      data = txBuffer;
    }
```
the `txBuffer` datas are ignored....

fprwi6labs: There are 2 methods completely different now. You can't mix them as you see.

But maybe I found a solution to mix them and keep all older sketches compatibles.
I look at this.

libraries/Wire/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Two new method has been added to allow the use of large number of bytes to write
+or read.
+
+* **writeBuffer(uint8_t \* pdata, size_t quantity)**: pass pdata directly to the
+I2C layer withtout fill the internal Wire library buffer. This method must be
+called only once after _beginTransmission()_. _endTransmission()_ must be called
+immediately after. No other _write()_ method have to be called between
+_beginTransmission()_ and _endTransmission()_ when _writeBuffer()_ is used.
+
```
LMESTM: I understand the idea and this would work ok - but I am not sure that this requirement (call only once, call endTRansmission immediately after ...) is acceptable - Isn't there an alternative that is more compliant with arduino defaut API ? Coudln't we simply increase internal buffer size ?  Or maybe manage malloc & free of extra buffers in case size becomes > 255  then in endTransmission you would send the default buffers + extra ones ? ... other feedbacks are welcome of course

libraries/Wire/README.md
```diff
@@ -0,0 +1,19 @@
...
+* **requestFrom(uint8_t address, uint16_t quantity, uint8_t \*pdata)**: this
+function allows to read data and store them directly in pdata buffer.
+Do not call _available()_ and _read()_ after this function because data are
+directly stored in pdata buffer.
```
LMESTM: Same here, if new methods are not easy to understand, I am not sure whether we should offer them

libraries/Wire/README.md
fpistm: We have discussed with @fprwi6labs  about that. Changing the static buffer to a dynamic one will consume memory space.  If several I2C instances are used and extra buffer size is required this will reach quickly the maximum space.
Proposal was to kept the current write function and add a new one with the usage restriction.

libraries/Wire/src/Wire.cpp
```diff
@@ -254,27 +296,55 @@ size_t TwoWire::write(uint8_t data)
...
+}
+
+/**
+  * @brief  This function must be called after beginTransmission() and before
+  *         endTransmission() only once. This function is only use in master mode.
+  * @param  pdata: pointer to the buffer data
+  * @param  quantity: number of bytes to write
+  * @retval number of bytes read. If > 0 then data are present in pdata buffer.
+  */
+size_t TwoWire::writeBuffer(uint8_t *pdata, size_t quantity)
```
fpistm: Add a warning about the pdata to ensure user do not freed or changed this buffer before endTransmission()

fprwi6labs: Done!

LMESTM:
@fpistm @fprwi6labs as said above, ok with me as it will not break existing libraries and we may live with this proposal until we possibly find a better one  - but I would not encourage a wide use of this API as long as it is not robust for begin/end and other APIs.

fprwi6labs:
Updated to use dynamic memory allocation. No new method added. Compatibility kept.
In master, the internal buffer size is now limited to the heap size.

libraries/Wire/src/Wire.cpp
```diff
@@ -97,6 +99,10 @@ void TwoWire::begin(int address)
 
 void TwoWire::end(void)
 {
+  free(txBuffer);
```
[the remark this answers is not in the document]
fprwi6labs: `free()` accepts null pointer 
> If ptr is a null pointer, the function does nothing.

libraries/Wire/src/Wire.cpp
```diff
@@ -109,6 +115,13 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 {
   UNUSED(sendStop);
   if (master == true) {
+    rxBuffer = allocateBuffer(rxBuffer, quantity);
+    // error if not memory block available to allocate the buffer
```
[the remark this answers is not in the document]
fprwi6labs: done!

libraries/Wire/src/Wire.cpp
```diff
@@ -241,8 +255,9 @@ size_t TwoWire::write(uint8_t data)
 {
   if(transmitting){
   // in master transmitter mode
-    // don't bother if buffer is full
-    if(txBufferLength >= BUFFER_LENGTH){
+    txBuffer = allocateBuffer(txBuffer, txBufferLength + 1);
+    // error if not memory block available to allocate the buffer
```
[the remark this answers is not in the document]
fprwi6labs: done!

libraries/Wire/src/Wire.cpp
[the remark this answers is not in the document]
fpistm: ok, i've got a doubt ;)

fpistm:
Hi @fprwi6labs,
could you have a look at this commit to enhance buffer usage:
https://github.com/fpistm/Arduino_Core_STM32/commit/c4b9f246ed9bf50839117087c480bfe60f11f4f6

fprwi6labs:
Hi @fpistm 

> could you have a look at this commit to enhance buffer usage:
fpistm/Arduino_Core_STM32@c4b9f24

It is OK for me.

fpistm:
Ok thanks. Waiting @LMESTM feedback

libraries/Wire/src/Wire.cpp
```diff
@@ -377,6 +411,37 @@ void TwoWire::onRequest( void (*function)(void) )
...
+  if(length < BUFFER_LENGTH) {
+    length = BUFFER_LENGTH;
+  }
+
+  buffer = (uint8_t *)realloc(buffer, length * sizeof(uint8_t));
+  return buffer;
+}
+
+/**
+  * @brief  Reset the buffer. Reduce is size if greater than BUFFER_LENGTH.
```
LMESTM: typo: is size / its size
comment says that it reduce the size only if greater than BUFFER_LENGHT but code seems to do it unconditionnally

libraries/Wire/src/Wire.cpp
```diff
@@ -377,6 +411,37 @@ void TwoWire::onRequest( void (*function)(void) )
...
+  * @param  buffer: pointer to the allocated buffer
+  * @retval pointer to the new buffer location
+  */
+uint8_t *TwoWire::resetBuffer(uint8_t *buffer)
+{
+  buffer = (uint8_t *)realloc(buffer, BUFFER_LENGTH * sizeof(uint8_t));
+  if(buffer != nullptr) {
+    memset(buffer, 0, BUFFER_LENGTH);
+  }
```
LMESTM: in case of nullptr, we need to raise an error (no memory ...). 
I 'm not sure this is checked after every call to resetBuffer ...

libraries/Wire/src/Wire.cpp
```diff
@@ -66,9 +66,11 @@ void TwoWire::begin(uint8_t address)
 {
   rxBufferIndex = 0;
   rxBufferLength = 0;
+  rxBuffer = resetBuffer(rxBuffer);
```
LMESTM: if null raise en error ?

libraries/Wire/src/Wire.cpp
```diff
@@ -66,9 +66,11 @@ void TwoWire::begin(uint8_t address)
...
 
   txBufferIndex = 0;
   txBufferLength = 0;
+  txBuffer = resetBuffer(txBuffer);
```
LMESTM: ditto

libraries/Wire/src/Wire.cpp
```diff
@@ -216,9 +224,15 @@ uint8_t TwoWire::endTransmission(uint8_t sendStop)
     break;
     }
 
+    //Reduce buffer size to free memory in case of large memory use
+    if(txBufferLength > BUFFER_LENGTH) {
+      txBuffer = resetBuffer(txBuffer);
```
LMESTM: raise an errror if null ?

libraries/Wire/src/Wire.cpp
```diff
@@ -241,8 +255,9 @@ size_t TwoWire::write(uint8_t data)
 {
   if(transmitting){
   // in master transmitter mode
-    // don't bother if buffer is full
-    if(txBufferLength >= BUFFER_LENGTH){
+    txBuffer = allocateBuffer(txBuffer, txBufferLength + 1);
```
LMESTM: maybe not efficient to allocate +1 byte each time - maybe you could go by steps (allocateBuffer(txBuffer, txBufferLength + BUFFER_LENGTH) each time it is needed) because reallocation can be costly in performances? 
In that way BUFFER_LENGHT would be used as the "unit" of buffer allocation.

libraries/Wire/src/Wire.cpp
fpistm: Error is raised later if rx/txBuffer is NULL

libraries/Wire/src/Wire.cpp
fpistm: Error is raised later if rx/txBuffer is NULL

libraries/Wire/src/Wire.cpp
fpistm: Right not efficient that's why I've proposed an enhancement in this commit:
https://github.com/fpistm/Arduino_Core_STM32/commit/c4b9f246ed9bf50839117087c480bfe60f11f4f6
Anyway I will add the allocation / BUFFER_LENGTH

libraries/Wire/src/Wire.cpp
fpistm: Will be removed in my PR.

libraries/Wire/src/Wire.cpp
fpistm: Change in the enhancement

libraries/Wire/src/Wire.cpp
fpistm: Checked in the R/W function

fpistm:
This PR has been replaced by  #152

Issue #123, opened 2017-10-04
I2C needs to be reworked to use IT mode


fprwi6labs:
The I2C layer needs some patches (twi.c in core):

- IRQHandler name do not match for F0 & L0
- ER_IRQn needs to be attached for all boards (except F0 & L0)
- Errors need to be taken account in master_read & master_write
- Error Callback must be reviewed to work in slave & master mode
- Other (to see)?

fpistm:
Buffer size could be reviewed to avoid #120

fpistm:
FYI: I've revert the commit which introduce I2C IT: 
https://github.com/stm32duino/Arduino_Core_STM32/commit/625fe2df7e67611fef9bf3e5a89a3265be7920c4

Let's reintroduced this thanks this issue solving.

PR #135, opened 2017-10-17
I2C IT mode

fprwi6labs:
Issue #123 fixed.

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -101,7 +101,16 @@
   */
 
 /*  Family specific description for I2C */
-#define I2C_NUM (5)
+#if defined(STM32F7xx) || defined(STM32L4xx)
+#define I2C_NUM (4)
+#elif defined(STM32F2xx) || defined(STM32F3xx) || defined(STM32F4xx) || defined(STM32L0xx)
+#define I2C_NUM (3)
+#elif defined(STM32F0xx) || defined(STM32F1xx) || defined(STM32L1xx)
+#define I2C_NUM (2)
+#else
+#error "Unknown Family - unknown I2C_NUM"
+#endif
+
 static I2C_HandleTypeDef* i2c_handles[I2C_NUM];
 
 /**
@@ -142,7 +151,6 @@ void i2c_init(i2c_t *obj)
   */
 void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, uint32_t ownAddress, uint8_t master)
 {
-  UNUSED(master);
   if(obj == NULL)
     return;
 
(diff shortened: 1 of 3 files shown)


cores/arduino/stm32/twi.c
```diff
@@ -101,7 +101,7 @@
   */
 
 /*  Family specific description for I2C */
-#define I2C_NUM (5)
+#define I2C_NUM (4)
```
LMESTM: can this value be defined in the family / target specific files ?

cores/arduino/stm32/twi.c
```diff
@@ -512,9 +527,14 @@ void HAL_I2C_ListenCpltCallback(I2C_HandleTypeDef *hi2c)
   */
 void HAL_I2C_ErrorCallback(I2C_HandleTypeDef *hi2c)
 {
-  HAL_I2C_EnableListen_IT(hi2c);
+  i2c_t *obj = get_i2c_obj(hi2c);
+
+  if(obj->isMaster == 0) {
+    HAL_I2C_EnableListen_IT(hi2c);
```
LMESTM: can you comment or explain this ErrorCallback management ? 
If there is an error, shouldn't we report the error to application ?

cores/arduino/stm32/twi.c
fpistm: It could be something like this:
```
#if defined(STM32F7xx) || defined(STM32L4xx)
#define I2C_NUM (4)
#elif defined(STM32F0xx) || defined(STM32F1xx) || defined(STM32L1xx)
#define I2C_NUM (2)
#else // STM32F2xx || STM32F3xx || STM32F4xx || STM32L0xx
#define I2C_NUM (3)
#endif
```

cores/arduino/stm32/twi.c
fprwi6labs: In master mode, the callback is useless because the error is reported to the Arduino API with `i2c_master_write()` and `i2c_master_read()`.
In slave mode, there is no mechanism in Arduino API to report an error so the error callback forces the slave to listen again.

cores/arduino/stm32/twi.c
fprwi6labs: Done!

cores/arduino/stm32/twi.c
LMESTM: Ok Good. So please explain this either with comments in the code, or in the commit message, this is useful information that we cannot guess when reading the code

fprwi6labs: Done!

cores/arduino/stm32/twi.c
```diff
@@ -170,53 +175,56 @@ void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, u
...
-      obj->irqER = I2C1_ER_IRQn;
-#endif
-      i2c_handles[0] = handle;
+    __HAL_RCC_I2C1_CLK_ENABLE();
+    __HAL_RCC_I2C1_FORCE_RESET();
+    __HAL_RCC_I2C1_RELEASE_RESET();
+
+    obj->irq = I2C1_EV_IRQn;
+#if !defined(STM32F0xx) && !defined(STM32L0xx)
+    obj->irqER = I2C1_ER_IRQn;
```
LMESTM: one new proposal : in every place where I2Cx_ER_IRQn is being used  couldn't you use 
#if defined (I2C1_ER_IRQn)
obj->irqER = I2C1_ER_IRQn;
#endif
this would be more future proof I think

cores/arduino/stm32/twi.c
```diff
@@ -101,7 +101,13 @@
   */
 
 /*  Family specific description for I2C */
-#define I2C_NUM (5)
+#if defined(STM32F7xx) || defined(STM32L4xx)
+#define I2C_NUM (4)
+#elif defined(STM32F0xx) || defined(STM32F1xx) || defined(STM32L1xx)
+#define I2C_NUM (2)
+#else // STM32F2xx || STM32F3xx || STM32F4xx || STM32L0xx
+#define I2C_NUM (3)
```
LMESTM: What do you thin about 
#elif defined( STM32F2xx) || defined(STM32F3xx) || defined(STM32F4xx) || #defined(STM32L0xx)
#else
#error Unknown Family - unknown I2C_NUM
#endif
so that 3 is not the default for any new family and we will be remonded to complete this part

cores/arduino/stm32/twi.c
LMESTM: where is it done ?

cores/arduino/stm32/twi.c
fpistm: I2C1_ER_IRQn is a enum value this could not be done.

cores/arduino/stm32/twi.c
fprwi6labs: Above in the header of the function.

cores/arduino/stm32/twi.c
fprwi6labs: Done!

PR #138, opened 2017-10-24
Fix SPI memory leak

fprwi6labs:
SPI library patched to remove overrun of an array causing random behaviour.
There is no problem with UART layer.
The [WiFi library](https://github.com/stm32duino/WiFi-ISM43362-M3G-L44/pull/1) is OK now.

### DIFF
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -57,7 +57,7 @@ void SPIClass::begin(uint8_t _pin)
     return;
 
   idx = pinIdx(_pin, ADD_NEW_PIN);
-  if(idx == NB_SPI_SETTINGS)
+  if(idx >= NB_SPI_SETTINGS)
     return;
 
   if((_pin != CS_PIN_CONTROLLED_BY_USER) && (_spi.pin_ssel == NC)) {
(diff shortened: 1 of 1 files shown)


libraries/SPI/src/SPI.cpp
```diff
@@ -262,22 +262,23 @@ uint16_t SPIClass::transfer16(uint8_t _pin, uint16_t data, SPITransferMode _mode
   if (_pin > NUM_DIGITAL_PINS)
     return rx_buffer;
 
-  if (spiSettings[_pin].msb) {
-    tmp = ((data & 0xff00) >> 8) | ((data & 0xff) << 8);
-    data = tmp;
+  uint8_t idx = pinIdx(_pin, GET_IDX);
+  if(idx == NB_SPI_SETTINGS) {
```
LMESTM: Maybe this would even be safer to check 
if(idx >= NB_SPI_SETTINGS)

fpistm: So should be applied on all those same tests done in SPI.cpp

fprwi6labs: Done!

PR #145, opened 2017-11-06
NUCLEO-F302R8 variant support

knielsen:
This pull request implements support for NUCLEO-F302R8.

The NUCLEO-F302R8 is similar to NUCLEO-F303RE, however a few pins are swapped, and the MCU has less flash/ram and peripherals.

The support was created following https://github.com/stm32duino/wiki/wiki/Add-a-new-variant-(board)

The PeripheralPins.c was generated from stm32cube files. The other files were edited from F303RE. Things were checked against board user manual and MCU datasheet.

Apart from boards.txt and variants/NUCLEO_F302R8/, there is a small fix in spi_com.c to support STM32's without SPI1 (the STM32F302R8 has only SPI2 and SPI3).


fpistm:
Hi @knielsen,

Thanks for the PR.
I've made some update/correction here:
https://github.com/fpistm/Arduino_Core_STM32/tree/Nucleo_F302R8

To sum up:
 - Change the system clock config to use the HSE and run at 72MHz.
 - Update the f_cpu in board.txt
 - Update the ldscript with the one generated by stm32cubemx (RAM size was wrong in the one you used for example)
 - Remove the UART_EMUL as not really tested for all series. I will required some test and probably moved as SoftwareSerial library.
 - Fix Duplicated pins (A6-11) to be aligned with PinMap_ADC in PeripheralPins.c
 - Finally I've squashe in one commit the variant you add

I've tested SPI (TFT with SD), I2C (eeproom and teperature sensors), tone, servo and Firmata --> ok
If you have any questions/feedback, do not hesitate.
If you are agree I will merge the new branch.

knielsen:
> I've made some update/correction here:
> https://github.com/fpistm/Arduino_Core_STM32/tree/Nucleo_F302R8

Thanks for looking at the patch and finding the problems you pointed out!

> To sum up:
>  - Change the system clock config to use the HSE and run at 72MHz.

That seems wrong?

There is no external HSE crystal mounted on the NUCLEO-F302R8 board. From
the user manual http://www.st.com/resource/en/user_manual/dm00105823.pdf:

  "HSE oscillator on-board from X3 crystal (not provided)"

Reading closer, there are two versions of the board (MB1136 C-01 / MB1136
C-02), and one of them has an 8MHz MCO clock signal going into the HSE that
can be used. However, this would require RCC_OscInitStruct.HSEState =
RCC_HSE_BYPASS, not RCC_HSE_ON.

And this will not work on the version C-01 board without ST-LINK MCO input,
nor with a board that has been broken off from the ST-LINK part.

That's why I decided to use the HSI originally. Do you agree, or did I miss
something?

>  - Update the f_cpu in board.txt

Good catch! Unfortunately, it is still incorrect, it should be 72 MHz, not
7.2 MHz :-) One zero missing:

  -Nucleo_64.menu.pnum.NUCLEO_F302R8.build.f_cpu=8000000L
  +Nucleo_64.menu.pnum.NUCLEO_F302R8.build.f_cpu=7200000L

Incidentally, I noticed the same wrong 8 MHz value for NUCLEO_F303RE (should
be 72 MHz) and NUCLEO_F091RC (should be 48 MHz I think).

>  - Update the ldscript with the one generated by stm32cubemx (RAM size
> was wrong in the one you used for example)

Ouch, I surely meant to put 16 KByte RAM in there, no idea how it ended up
as 8 kByte :-/. Your new linker script looks correct, thanks!

>  - Remove the UART_EMUL as not really tested for all series. I will
> required some test and probably moved as SoftwareSerial library.

Right, I wasn't sure about the uart emulation. The STM32F302R8 has 3
hardware USARTs with multiple possible pinouts, so there shouldn't be too
much need for software serial anyway.

Should I also remove the #define TIMER_UART_EMULATED TIM17 ?

And I'll enable some TIM17 pins in PeripheralPins.c. I notived that
NUCLEO_F103RB for example also enables the pins for TIMER_UART_EMULATED in
PinMap_PWM.

>  - Fix Duplicated pins (A6-11) to be aligned with PinMap_ADC in PeripheralPins.c

Oh, I see, I totally missed that, thanks for fixing.

>  - Finally I've squashe in one commit the variant you add

Ok.

I have pushed an updated branch to the pull request, with the
above-suggested changes on top of your fpistm:Nucleo_F302R8 branch.

Let me know what you think,

 - Kristian.

fpistm:
Nice shot too ;)
Right for the HSE.
ok for f_cpu (in fact it is only set as an information, the systemcoreclock  value is set to the correct one avec clock config)
TIMER_UART_EMULATED TIM17  could be removed

fpistm:
I've updated my branch with your update.
Removed the TIMER_UART_EMULATED
Set the RCC PLL MUL to 16 instead of 9 in order to be at 64MHz.

knielsen:
> Set the RCC PLL MUL to 16 instead of 9 in order to be at 64MHz.

Agree. Interesting that the STM32F302R8 isn't able to run at the full 72 MHz from the HSI.
Maybe the f_cpu in boards.txt (and the comments in variant.cpp) should be updated to reflect the 64 MHz.

> Removed the TIMER_UART_EMULATED

Thanks, looks good now I think!

fpistm:
Corrected.
I will merge the #147

fpistm:
Merged thanks #147

Issue #41, opened 2017-06-27
Use recipe.hooks feature of the platform.txt file


fprwi6labs:
Orignally posted in #38 by @RickKimball

I think you could take advantage of the recipe.hooks feature of the platform.txt file to conditionally override defines and gcc command line arguments on a per sketch basis.  This seems to be a fairly new feature (since Arduino 1.6.something) something they added since I last looked to accomplish something like this.

You might approach it like this:

In the stm32 tools directory create a new tool called 'cpif' which copies a file called gcc_options.txt into the build directory if the file exists in the sketch directory or copies /dev/null if it doesn't
```
$ cat cpif 
#!/bin/bash
set -x
echo $0 $*

sketchdir=${1%.*}
dest=${2}/gcc_options.txt

options="$HOME/Arduino/$sketchdir/gcc_options.txt"

if [ -f $options ]
then
   cp $options $dest
else
   cp /dev/null $dest
fi
```
In platform.txt you add a recipe hook that is called before starting the build process.
```
 ## Compile c files
+recipe.hooks.prebuild.1.pattern=cpif {build.project_name} {build.path} b4all
 recipe.c.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.c.flags} -mcpu={build.mcu} -DF_CPU={build ...
```
modify the board.txt to use the @file feature of gcc or just add it to the platform.txt for all boards
```
Nucleo_64.menu.Nucleo_64_board.NUCLEO_F030R8.build.extra_flags=-DSTM32F030x8 @{build.path}/gcc_options.txt
```
When the user wants to override defines and gcc command line arguments they create a file called gcc_options.txt in the sketch directory. Here I imagine I want to override the F_CPU value
```
-U F_CPU
-D F_CPU=24000000UL
```
When I compile it uses the sketch specific defines and overrides.
```
[...]

PR #148, opened 2017-11-09
Disabled Core callback feature

fpistm:
STM32Ethernet do not use anymore this feature.
Disabled it to avoid unuseful call in the main while{}.
Define CORE_CALLBACK allow to enable it.


fprwi6labs:
Where will CORE_CALLBACK be defined?

fpistm:
Currently anywhere.
It could be documented for the #41.
Goal is only to disable it to avoid call in the loop.

PR #152, opened 2017-11-15
Change Wire Rx/Tx buffers managements

fpistm:
Rx/Tx Buffers are now dynamically allocated (min size: BUFFER_LENGTH)
`TwoWire::write(const uint8_t *data, size_t quantity)` avoid copy byte/byte in the buffer

Replace #131 to solve issue #124


libraries/Wire/src/Wire.cpp
```diff
@@ -323,11 +331,10 @@ int TwoWire::read(void)
...
-    /*  Reduce buffer size to free memory in case of large memory use when no more
-        data available */
-    if((rxBufferIndex == rxBufferLength) && (rxBufferLength > BUFFER_LENGTH)) {
-      rxBuffer = resetBuffer(rxBuffer);
-    }
+    /* Reset rx buffer when no more data available */
+    /*if(rxBufferIndex == rxBufferLength) {
+      resetRxBuffer();
+    }*/
   }
```
LMESTM: Why keeping commented source code ?

libraries/Wire/src/Wire.cpp
```diff
@@ -411,37 +416,6 @@ void TwoWire::onRequest( void (*function)(void) )
...
-  */
-uint8_t *TwoWire::resetBuffer(uint8_t *buffer)
-{
-  buffer = (uint8_t *)realloc(buffer, BUFFER_LENGTH * sizeof(uint8_t));
-  if(buffer != nullptr) {
-    memset(buffer, 0, BUFFER_LENGTH);
-  }
-  return buffer;
-}
-
```
LMESTM: why moving to .h file ? maybe you can simply move it up in the C file ?

libraries/Wire/src/Wire.h
```diff
@@ -56,8 +58,24 @@ class TwoWire : public Stream
     static void onRequestService(void);
     static void onReceiveService(uint8_t*, int);
 
-    uint8_t *allocateBuffer(uint8_t *buffer, size_t length);
-    uint8_t *resetBuffer(uint8_t *buffer);
+    inline void allocateRxBuffer(size_t length){ if(rxBufferAllocated < length) {
+        // By default we allocate BUFFER_LENGTH bytes. It is the min size of the buffer.
```
LMESTM: missing return lines

libraries/Wire/src/Wire.h
```diff
@@ -56,8 +58,24 @@ class TwoWire : public Stream
...
+        if(length < BUFFER_LENGTH) { length = BUFFER_LENGTH; }
+        rxBuffer = (uint8_t *)realloc(rxBuffer, length * sizeof(uint8_t));
```
LMESTM: should't you check for a MAX SIZE somewhere to report an error rather than allocating all the available heap memory ?

libraries/Wire/src/Wire.cpp
fpistm: Just to show it is possible to reset the Rx buffer here. But think it is not required.

libraries/Wire/src/Wire.h
fpistm: It was a choice but I can add one.

libraries/Wire/src/Wire.h
fpistm: I think about it but I would not add more check.
If all heap is used then buffer will be at nullptr then an error will be raised.

libraries/Wire/src/Wire.cpp
fpistm: Move it as `inline` function in the class to avoid function call

LMESTM: inline is fine - but why not keep it in wire.cpp file ?

Issue #153, opened 2017-11-20
I2C broken on STM32F4 since IT mode used


fpistm:
Since the I2C IT mode is used #135, I2C tests with STM32 F4 based boards are broken.
(reproduced on several boards: 429ZI, 401RE, 411RE).
Seems the write operation do not end properly: ACK failed.
Changing the timing change the error occurrences.

Blocke for next release.

cparata:
Hi all,
[this](https://github.com/stm32duino/Arduino_Core_STM32/pull/155) PR should fix this issue.
Best Regards,
Carlo

PR #155, opened 2017-11-21
Fix I2C bug for F4 platforms in IT mode

cparata:
Hi all,
this patch should fix [this](https://github.com/stm32duino/Arduino_Core_STM32/issues/153) bug. I successfully tested it with Nucleo-F401RE and X-NUCLEO-NFC01A1 and X-NUCLEO-NFC04A1. In order to fix it, I took inspiration by Cube HAL examples for I2C in IT mode. Please, let me know if it works also on your side. 
Best Regards,
Carlo

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -353,19 +353,25 @@ i2c_status_e i2c_master_write(i2c_t *obj, uint8_t dev_address,
 {
   i2c_status_e ret = I2C_ERROR;
   uint32_t tickstart = HAL_GetTick();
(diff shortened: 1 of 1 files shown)


cparata:
Unfortunately, it seems that the "i2c_IsDeviceReady" is only a low level API that is not available in the Wire library. So we cannot use that API for Arduino libraries. Anyway, the proposed patch should fix the I2C issue without any change in the Wire library.

LMESTM:
@cparata ok looks good to me  !

fpistm:
This Fix #153

Issue #165, opened 2017-12-04
New Board: STM32F030F4 Demo Board


Adminius:
Hi,
i'm trying to add support for this china board.
![foto](https://user-images.githubusercontent.com/17052262/33550096-feb2b7ec-d8ec-11e7-83e4-816a2d5b5dc3.png)

1. Question: I think nobody will have "custom" boards in official Core. Right? => how to share this board with others?
2. STM32CubeMX doen't export stm32f030x4.h... so i use stm32f030x6.h instead, is it ok?
3. I tried to compile Blink.ino, but: `Sketch uses 10124 bytes (61%) of program storage space. Maximum is 16384 bytes.
Global variables use 3704 bytes (90%) of dynamic memory, leaving 392 bytes for local variables. Maximum is 4096 bytes.
Low memory available, stability problems may occur.`
Any chance to get more RAM?

fpistm:
Hi,
1/ it is possible to add custom boards. This core mainly manage MCU. Boards is only a dedicated "pinmapping".  (ex #61 I have to end).
2/ I think you should use this one: STM32F030F4Px.xml
(if this is this one http://www.hotmcu.com/stm32f030f4p6-minimum-systerm-boardcortexm0-p-208.html)
3. You could try to disable some instance (ex: No Serial)
But doing more stuff than a blink will not use a lot of more RAM as instance will be already here.

fpistm:
Hi @Adminius,
is it ok to close this issue?

Adminius:
@fpistm pls not yet. i was on vacation. 
it's ready, but i have to test it.

fpistm:
No worries ;)

Adminius:
btw. this board/issue is not blocker for next release ;)

Sod-Almighty:
I have this board. How can I use it with this library? I don't see any options to select this model of chip.

Adminius:
it is not possible yet... i have to find a time to test my implementation. after that i can commit it to this repository.

P.S. what are you thinking about this PIN-Order?
![grafik](https://user-images.githubusercontent.com/17052262/36977258-d7784346-2080-11e8-9c6b-330ed1c48404.png)

This board have not enough pins for standart Arduino Layout (e.g. SPI on D10-D13)
I also "converted" UART-Header to i2c header, because i2c is only available on this pins, but UART can be switched to PA2/PA3

fpistm:
[...]

PR #167, opened 2017-12-05
Clean variants files

fpistm:
- Common definitions are moved to pins_arduino_var.h.
  They could be redefined in variant.h if needed.
- Removed *UART_EMUL* definitions as it is needed to be reworked in SoftwareSerial library.
- Pin number over D15 replaced by pin name (more relevant)
- Fixed some typos

### DIFF
--- a/cores/arduino/pins_arduino_var.h
+++ b/cores/arduino/pins_arduino_var.h
@@ -100,6 +100,54 @@ uint32_t pinNametoDigitalPin(PinName p);
                                      (digitalPinFirstOccurence(p) == PIN_SERIAL_TX))
 #endif
 
+// Default Definitions, could be redefined in variant.h
+#ifndef ADC_RESOLUTION
+#define ADC_RESOLUTION              12
+#endif
+#ifndef DACC_RESOLUTION
+#define DACC_RESOLUTION             12
+#endif
+#ifndef PWM_RESOLUTION
+#define PWM_RESOLUTION              8
+#endif
+#ifndef PWM_FREQUENCY
+#define PWM_FREQUENCY               1000
+#endif
+#ifndef PWM_MAX_DUTY_CYCLE
+#define PWM_MAX_DUTY_CYCLE          255
+#endif
+
+// Default for Arduino connector compatibility
+// SPI Definitions
+#ifndef SS
+#define SS                          10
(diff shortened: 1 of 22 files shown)


Adminius:
https://github.com/stm32duino/Arduino_Core_STM32/issues/165
F030F4 has not enough pins for "SDA 14 / SCL 15"
should i use pin doubles and/or "NC" for that cases?
Example:

> enum {
  PA13, //D0  - SWCLK
  PC14, //D1  - SWDIO
  PA10, //D2  - SDA (RX UART Header)
  PA9,  //D3  - SCL (TX UART Header)
  PB1,  //D4
  PA7,  //D5  - MOSI
  PA6,  //D6  - MISO
  PA5,  //D7  - SCK
  PA4,  //D8  - SS - LED
  PA3,  //D9  - RX
  PA2,  //D10 - TX 
  PA1,  //D11
  PA0,  //D12
 // Duplicated pins in order to be aligned with PinMap_ADC 
  PA0_1,  //D13/A0 = D12
  PA1_1,  //D14/A1 = D11
  PA2_1,  //D15/A2 = D10
  PA3_1,  //D16/A3 = D9
  PA4_1,  //D17/A4 = D8
  PA5_1,  //D18/A5 = D7
  PA6_1,  //D19/A6 = D6
  PA7_1,  //D20/A7 = D5
  PB1_1,  //D21/A8 = D4
  PEND
};

fpistm:
by default, for board with Arduino header,  SCL and SDA are on D14/D15 , if the board you add do not have it, only define it properly in the variant.h:
Based on your example (I used PXn name to match silkscreen of your board):
```
//I2C Definitions
#define SDA                    PA10
#define SCL                     PA9
```

fpistm:
> Could we move also LED_BUILTIN and USER_BTN?

I'm wondered the same question, using those default values:
```
#define LED_BUILTIN             13
#define USER_BTN                PC13
```
but some boards do not have LED or button, so I preferred to not define one by default.
Are you agreed with that?

fprwi6labs:
Yes, I am.
In fact I missed the boards without LED or button.

fpistm:
With this PR, variant is now cleaned and include only specific pin mapping

PR #175, opened 2017-12-12
Introduce new ADC API for DMA-based high sampling rate

damienWi6labs:
Implement DMA API for ADC for F0, F1, F3, F4, L0 and L4 families. The DMA is used for maximize the ADC sampling rate. Measure the sampling rate for each ADC and verify that they match the datasheet. For example on L4 board measure a sampling rate of 5,33 Ms/s on the three ADCs. Also check that we can make two ADC conversions at the same by using the DMA with two differents ADCs.

### DIFF
--- a/cores/arduino/stm32/analog.c
+++ b/cores/arduino/stm32/analog.c
@@ -50,6 +50,7 @@
 #include "analog.h"
 #include "timer.h"
 #include "PinAF_STM32F1.h"
+#include "dma.h"
 
 #ifdef __cplusplus
  extern "C" {
@@ -69,27 +70,28 @@
 /** @addtogroup STM32F4xx_System_Private_Defines
   * @{
   */
-#if defined(ADC_SAMPLETIME_15CYCLES)
-#define SAMPLINGTIME        ADC_SAMPLETIME_15CYCLES;
-#elif defined(ADC_SAMPLETIME_13CYCLES_5)
-#define SAMPLINGTIME        ADC_SAMPLETIME_13CYCLES_5;
-#elif defined(ADC_SAMPLETIME_19CYCLES_5)
-#define SAMPLINGTIME        ADC_SAMPLETIME_19CYCLES_5;
-#elif defined(ADC_SAMPLETIME_16CYCLES)
-#define SAMPLINGTIME        ADC_SAMPLETIME_16CYCLES;
-#elif defined(ADC_SAMPLETIME_12CYCLES_5)
-#define SAMPLINGTIME        ADC_SAMPLETIME_12CYCLES_5;
+
+// Use smalled sample time decreases ADC conversion time
+#if defined(ADC_SAMPLETIME_1CYCLE_5)
+#define SAMPLINGTIME        ADC_SAMPLETIME_1CYCLE_5;
+#elif defined(ADC_SAMPLETIME_2CYCLES_5)
+#define SAMPLINGTIME        ADC_SAMPLETIME_2CYCLES_5;
+#elif defined(ADC_SAMPLETIME_3CYCLES)
+#define SAMPLINGTIME        ADC_SAMPLETIME_3CYCLES;
+#elif defined(ADC_SAMPLETIME_4CYCLES)
+#define SAMPLINGTIME        ADC_SAMPLETIME_4CYCLES;
 #else
 #error "ADC SAMPLINGTIME could not be defined"
 #endif
 
+// Use smaller clock prescaler to keep ADC conversion time as short as possible
 #ifndef STM32F1xx
-#ifdef ADC_CLOCK_SYNC_PCLK_DIV2
-#define ADC_CLOCK_DIV       ADC_CLOCK_SYNC_PCLK_DIV2
+#ifdef ADC_CLOCK_SYNC_PCLK_DIV1
+#define ADC_CLOCK_DIV       ADC_CLOCK_SYNC_PCLK_DIV1
 #elif defined(ADC_CLOCK_ASYNC_DIV1)
 #define ADC_CLOCK_DIV       ADC_CLOCK_ASYNC_DIV1
-#elif defined(ADC_CLOCKPRESCALER_PCLK_DIV2)
-#define ADC_CLOCK_DIV       ADC_CLOCKPRESCALER_PCLK_DIV2
+#elif defined(ADC_CLOCK_SYNC_PCLK_DIV2)
+#define ADC_CLOCK_DIV       ADC_CLOCK_SYNC_PCLK_DIV2
 #else
 #error "ADC_CLOCK_DIV could not be defined"
 #endif
@@ -110,10 +112,48 @@
   * @}
   */
 
-/** @addtogroup STM32F4xx_System_Private_Variables
-  * @{
-  */
+/* Adc context                                                                */
+AdcContext context[ADC_NUM];
+
+/* GPIO pin use for DAC                                                       */
 static PinName g_current_pin = NC;
+
+
+/**
+  * @brief get indice for fulfill DMA table
+  * @param instance : pointer to the instance
+  * @retval         : indice for the corresponding peripheral
+  */
+uint8_t get_indice(void *instance)
+{
+  uint8_t indice = 255;
+  
+  #ifdef ADC1_BASE
+  if (instance == ADC1)
+  {
+    indice = 0;
+  }
+  #endif
+  #ifdef ADC2_BASE
+  if (instance == ADC2)
+  {
+     indice = 1;
+  }
+  #endif
+  #ifdef ADC3_BASE
+  if (instance == ADC3)
+  {
+    indice = 2;
+  }
+  #endif
+  #ifdef ADC4_BASE
+  if (instance == ADC4)
+  {
+    indice = 3;
+  }
+  #endif
+  return indice;
+}
 
 /**
   * @}
@@ -388,13 +428,19 @@ void dac_stop(PinName pin)
   *        This function configures the hardware resources used in this example:
   *           - Peripheral's clock enable
   *           - Peripheral's GPIO Configuration
+  *           - Peripheral's DMA
   * @param hadc: ADC handle pointer
   * @retval None
   */
 void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
 {
   GPIO_InitTypeDef  GPIO_InitStruct;
-  GPIO_TypeDef *port;
+  GPIO_TypeDef *port = NULL;
+  uint8_t indice = 255;
+  void *channel = NULL;
+  IRQn_Type irqHandler;
+  
+  
   /*##-1- Enable peripherals and GPIO Clocks #################################*/
   /* ADC Periph clock enable */
   if(hadc->Instance == ADC1) {
@@ -434,18 +480,47 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
 #ifdef __HAL_RCC_ADC_CLK_ENABLE
   __HAL_RCC_ADC_CLK_ENABLE();
 #endif
+
 /* For STM32F1xx, ADC prescaler is confgured in SystemClock_Config (variant.cpp) */
 #if defined(__HAL_RCC_ADC_CONFIG) && !defined(STM32F1xx)
   /* ADC Periph interface clock configuration */
   __HAL_RCC_ADC_CONFIG(RCC_ADCCLKSOURCE_SYSCLK);
 #endif
 
-  /* Enable GPIO clock ****************************************/
-  port = set_GPIO_Port_Clock(STM_PORT(g_current_pin));
-
   /*##-2- Configure peripheral GPIO ##########################################*/
+#ifdef ADC1_BASE
+  if (hadc->Instance == ADC1)
+  {
+    port = set_GPIO_Port_Clock(STM_PORT(context[0].current_pin));
+    GPIO_InitStruct.Pin = STM_GPIO_PIN(context[0].current_pin);
+  }
+#endif
+  
+#ifdef ADC2_BASE
+  if (hadc->Instance == ADC2)
+  {
+    port = set_GPIO_Port_Clock(STM_PORT(context[1].current_pin));
+    GPIO_InitStruct.Pin = STM_GPIO_PIN(context[1].current_pin);
+  }
+ #endif
+  
+#ifdef ADC3_BASE
+  if (hadc->Instance == ADC3)
+  {
+    port = set_GPIO_Port_Clock(STM_PORT(context[2].current_pin));
+    GPIO_InitStruct.Pin = STM_GPIO_PIN(context[2].current_pin);
+  }
+#endif
+  
+#ifdef ADC4_BASE
+  if (hadc->Instance == ADC4)
+  {
+    port = set_GPIO_Port_Clock(STM_PORT(context[3].current_pin));
+    GPIO_InitStruct.Pin = STM_GPIO_PIN(context[3].current_pin);
+  }
+#endif
+
   /* ADC Channel GPIO pin configuration */
-  GPIO_InitStruct.Pin = STM_GPIO_PIN(g_current_pin);
 #ifdef GPIO_MODE_ANALOG_ADC_CONTROL
   GPIO_InitStruct.Mode = GPIO_MODE_ANALOG_ADC_CONTROL;
 #else
@@ -453,6 +528,69 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
 #endif
   GPIO_InitStruct.Pull = GPIO_NOPULL;
   HAL_GPIO_Init(port, &GPIO_InitStruct);
+  
+  indice = get_indice(hadc->Instance);
+  
+  if (context[indice].use_dma)
+  {
+    /*##-3- Ask DMA informations ###############################################*/
+    if (dma_allocate(hadc->Instance, &channel, &irqHandler))
+    {
+      /*##-4- Configure the DMA ##################################################*/
+          
+      #if defined(STM32L0xx) || defined(STM32L4xx)
+      context[indice].DmaHandle.Init.Request             = DMA_REQUEST_0;
+      #endif
+
+      #if defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
+      #ifdef ADC1_BASE
+      if (hadc->Instance == ADC1)
+      {
+        context[indice].DmaHandle.Init.Channel           = DMA_CHANNEL_0;
+      }
+      #endif
+      #ifdef ADC2_BASE
+      if (hadc->Instance == ADC2)
+      {
+        context[indice].DmaHandle.Init.Channel           = DMA_CHANNEL_1;
+      }
+      #endif
+      #ifdef ADC3_BASE
+      if (hadc->Instance == ADC3)
+      {
+        context[indice].DmaHandle.Init.Channel           = DMA_CHANNEL_2;
+      }
+      #endif
+      #endif //STM32F2xx STM32F4xx STM32F7xx
+
+      context[indice].DmaHandle.Instance                 = channel;
+      context[indice].DmaHandle.Init.Direction           = DMA_PERIPH_TO_MEMORY;
+      context[indice].DmaHandle.Init.PeriphInc           = DMA_PINC_DISABLE;
+      context[indice].DmaHandle.Init.MemInc              = DMA_MINC_ENABLE;
+      context[indice].DmaHandle.Init.PeriphDataAlignment = DMA_PDATAALIGN_HALFWORD;
+      context[indice].DmaHandle.Init.MemDataAlignment    = DMA_MDATAALIGN_WORD;
+      context[indice].DmaHandle.Init.Mode                = DMA_NORMAL;
+      context[indice].DmaHandle.Init.Priority            = DMA_PRIORITY_MEDIUM;
+
+      #if defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
+      context[indice].DmaHandle.Init.FIFOMode            = DMA_FIFOMODE_DISABLE;
+      context[indice].DmaHandle.Init.FIFOThreshold       = DMA_FIFO_THRESHOLD_HALFFULL;
+      context[indice].DmaHandle.Init.MemBurst            = DMA_MBURST_SINGLE;
+      context[indice].DmaHandle.Init.PeriphBurst         = DMA_PBURST_SINGLE;
+      #endif //STM32F2xx STM32F4xx STM32F7xx
+          
+      /* Deinitialize  & Initialize the DMA for new transfer */
+      HAL_DMA_DeInit(&context[indice].DmaHandle);
+      HAL_DMA_Init(&context[indice].DmaHandle);
+
+      /* Associate the DMA handle */
+      __HAL_LINKDMA(hadc, DMA_Handle, context[indice].DmaHandle);
+
+      /* NVIC */
+      HAL_NVIC_SetPriority(irqHandler, 1, 0);
+      HAL_NVIC_EnableIRQ(irqHandler);
+    }
+  }
 }
 
 /**
@@ -462,6 +600,8 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
   */
 void HAL_ADC_MspDeInit(ADC_HandleTypeDef* hadc)
 {
+  uint8_t indice = 255;
+
 #ifdef __HAL_RCC_ADC_FORCE_RESET
   __HAL_RCC_ADC_FORCE_RESET();
 #endif
@@ -543,6 +683,14 @@ void HAL_ADC_MspDeInit(ADC_HandleTypeDef* hadc)
 #ifdef __HAL_RCC_ADC_CLK_DISABLE
   __HAL_RCC_ADC_CLK_DISABLE();
 #endif
+
+  indice = get_indice(hadc->Instance);
+
+  if (context[indice].use_dma)
+  {
+    dma_free(hadc->Instance);
+    HAL_DMA_DeInit(hadc->DMA_Handle);
+  }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+      M]0M   =MM]1 .M]'  MM]     jM](  1       r    j  1  --,   "!         
+      ]0MT   ]M]M  jM@   ]M]     M]P  j       J     j j     4     1        
+      MM]'   M]0P  j]1  .M]M    j]M'  J      j'     ",?     j     1        
+     _]M]    M]0`  jM1 .MNMM,  .]M'   1     .'       11     1    j'        
+     jM]1   jM]@   j]L_]'?M]M__MP'    \     J        1G    J    .'         
+     j]0(   jM]1   "M]P'  "N]M/-      "L__J L________'?L__- *__,'          
+     "-'    "--                                                            
+                                                                           
+----------------------------------------------------------------------------
+
```
LMESTM: Is ASCII art really needed ?

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+#include "stm32_def.h"
+#include "dma.h"
+#include "analog.h"
+
+#ifdef __cplusplus
+ extern "C" {
+#endif
+
+extern ADC_HandleTypeDef AdcHandleTab[ADC_NUM];
+
```
LMESTM: dma.c file should not have any reference to ADC, so that it can later be re-used by oher devices

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+  
+#ifdef __HAL_RCC_DMA1_CLK_ENABLE
+  __HAL_RCC_DMA1_CLK_ENABLE();
+#endif
+
+#ifdef __HAL_RCC_DMA2_CLK_ENABLE
+  __HAL_RCC_DMA2_CLK_ENABLE();
+#endif
+
+  channel = pinmap_find_dmaChannel(hadc->Instance, DmaMap_DMA);
```
LMESTM: Instead of DmaMap_DMA a generic table can be used, where any kind of devices Serial, ADC could be listed with corresponding ADC instance and channel. ... 
Then the table could be parsed with a for loop instead of checks like (if (hadc->Instance == ADC3))

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+  #ifdef ADC1_BASE
+  if (hadc->Instance == ADC1)
```
LMESTM: rather use a for loop over the DMA descritpion table for finding a match.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+  #endif
+  DmaHandleTab[indice].Init.FIFOMode              = DMA_FIFOMODE_DISABLE;
+  DmaHandleTab[indice].Init.FIFOThreshold         = DMA_FIFO_THRESHOLD_HALFFULL;
+  DmaHandleTab[indice].Init.MemBurst              = DMA_MBURST_SINGLE;
+  DmaHandleTab[indice].Init.PeriphBurst           = DMA_PBURST_SINGLE;
+  #endif //STM32F4xx
+  
+  /* Deinitialize  & Initialize the DMA for new transfer */
+  HAL_DMA_DeInit(&DmaHandleTab[indice]);
+  HAL_DMA_Init(&DmaHandleTab[indice]);
```
LMESTM: Once DMA instance and channel are found, the DMA init itself could be left in the ADC driver

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+  
+  /* Associate the DMA handle */
+  __HAL_LINKDMA(hadc, DMA_Handle, DmaHandleTab[indice]);
+
+  /* Set clock as active */
+  channelNumber = pinmap_find_dmaChannelNumber(hadc->Instance, DmaMap_DMA);
+  dmaNumber     = pinmap_find_dma(hadc->Instance, DmaMap_DMA);
+  dmaTab[dmaNumber-1][channelNumber] = true;
+
```
LMESTM: rather use a generic address pointer to instance for pinmap_find_dmaChannelNumber and pinmap_find_dma

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+static bool dmaTab[DMA_NUMBER][CHANNEL_NUMBER];
+
+/**
+  * @brief  This function will init the dma and channel corresponding 
+  *         to the ADC peripheral
+  * @param  pin : the ADC peripheral to use
+  * @retval none
+  */
+void dma_init_adc(ADC_HandleTypeDef *hadc)
+{
```
LMESTM: As mentioned DMA PAI should not mention ADC at all. 
instead of dma_init_adc I would rather consider a generic function like dma_allocate to look for the proper DMA instance and channel with a generic (void*) pointer to device. then the function would look for an instance match in the table with pinmap_find_dmaChannel.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/**
+  * @brief  This function will deinit the dma and channel corresponding 
+  *         to the ADC peripheral
+  * @param  hadc : the ADC peripheral to use
+  * @retval none
+  */
+void dma_deinit_adc(ADC_HandleTypeDef *hadc)
+{
```
LMESTM: rathe have dma_free or dma_put to release a DMA handle.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+
+#ifdef ADC1_BASE
+void DMA_ADC1_IRQHandler1(void)
+{
+  #ifdef STM32F4xx
+  if (AdcHandleTab[0].DMA_Handle != NULL)
+  {
+    HAL_DMA_IRQHandler(AdcHandleTab[0].DMA_Handle); // ADC1
+  }
+  else
```
LMESTM: could be 
elsif (AdcHandleTab[0].DMA_Handle != NULL)

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+
+/**
+  * @brief  DMA IRQ Handle 
+  * @param  None
+  * @retval None
+  */
+
+#ifdef ADC1_BASE
```
LMESTM: Wouldn't it be better to keep generic name ?
DMA1_Channel1_IRQHandler

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,315 @@
...
+}
+#endif
+
+#if defined(STM32L011xx) || defined(STM32L021xx) || defined(STM32L031xx) || \
+    defined(STM32L041xx) || defined(STM32L051xx) || defined(STM32L061xx) || \
+    defined(STM32L071xx) || defined(STM32L081xx) || defined(STM32F091xC) || \
+    defined(STM32F098xx)
+void DMA_ADC1_IRQHandler3(void)
+{
+  HAL_DMA_IRQHandler(AdcHandleTab[0].DMA_Handle);
```
LMESTM: Not checking handle is != NULL ?

cores/arduino/stm32/analog.c
```diff
@@ -434,25 +438,57 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
 #if defined(__HAL_RCC_ADC_CONFIG) && !defined(STM32F1xx)
   /* ADC Periph interface clock configuration */
   __HAL_RCC_ADC_CONFIG(RCC_ADCCLKSOURCE_SYSCLK);
 #endif
 
-  /* Enable GPIO clock ****************************************/
-  port = set_GPIO_Port_Clock(STM_PORT(g_current_pin));
-
   /*##-2- Configure peripheral GPIO ##########################################*/
+#ifdef ADC1_BASE
```
LMESTM: How is this part of change related to "prepare ADC DMA for F3 family" ?

cores/arduino/stm32/dma.c
damienWi6labs: It's remove

cores/arduino/stm32/dma.c
damienWi6labs: dma_allocate create. A enum is create, we need it for identify the peripheral.

LMESTM: sorry I don't understand your answer. 

My proposal is 
replace : void dma_init_adc(ADC_HandleTypeDef *hadc)
by : void dma_init(void *any_handle_typedef)

so that it can later be used also by other devices than ADC

cores/arduino/stm32/dma.c
damienWi6labs: pinmap_find_dmaChannel is similar to other functions like pin_pinName or pinmap_find_function. Why should it be different? DmaMap_DMA contains all informations needed for DMA like PinMap_ADC contains all informations needed for ADC.

cores/arduino/stm32/dma.c
damienWi6labs: This is what we have done. But in dma_init() we also need an other parameter for know on what peripheral we work (ADC, UART). So the prototype is : 
void dma_allocate(void *periph_handle, Peripheral periph);

cores/arduino/stm32/dma.c
damienWi6labs: We need to know is the ADC exist before the "if" condition. Some device have only one ADC...
The same method is used in timer.c

cores/arduino/stm32/dma.c
damienWi6labs: Is it not better to do all the DMA things in DMA layer?

cores/arduino/stm32/dma.c
damienWi6labs: dma_free() is create. The prototype is void dma_free(void *periph_handle, Peripheral perip).
Again, we need to know on which peripheral we work.

cores/arduino/stm32/dma.c
damienWi6labs: Done. Null pointer is check.

cores/arduino/stm32/analog.c
damienWi6labs: It was a old commit, i rename it.

cores/arduino/stm32/dma.c
damienWi6labs: We can't keep generic name because the are different from one device to another. For example for F3 family this is DMA2_Channel1_IRQHandler and for F4 family this is DMA2_Stream1_IRQHandler.  (Channel/Stream)

cores/arduino/stm32/dma.c
damienWi6labs: We will see if it possible in next release.

cores/arduino/stm32/dma.c
fpistm: One important thing: PeripheralsPins.c contains arrays for pin mapping.
What a pin is able to do for a dedicated peripheral and how.
So adding this kind of array not linked to a pin is not correct.
+const DmaMap DmaMap_DMA[] = {
+    {ADC1, 1, DMA1_Channel1, DMA1_Channel1_IRQn        , 1},
+    {ADC1, 1, DMA1_Channel2, DMA1_Ch2_3_DMA2_Ch1_2_IRQn, 2},
+    {ADC1, 2, DMA2_Channel5, DMA1_Ch4_7_DMA2_Ch3_5_IRQn, 5},
+    {NP,   0, NP,            NP,                         0}
+};

cores/arduino/stm32/dma.c
damienWi6labs: I don't really understand what you mean. If and elseif can't have the same condition. I can put a condition on the elseif if that's what you want.

cores/arduino/stm32/dma.c
damienWi6labs: A DMA is link to a peripheral not a pin... i agree we that. Can we keep this table an put it in another file? Oterwise, what do you propose?

fpistm: Currently, I do not reviewed this PR so I have no proposal. Just a note on PeripheralPins.c file usage.

cores/arduino/stm32/dma.c
damienWi6labs: After study, it's not possible to remove this variable from DMA. The DMA link a peripheral, it need to have the information that is in the ADC structure. If we don't share this variable between ADC and DMA, DMA will not work.

cores/arduino/stm32/dma.c
LMESTM: Ok I understand that - I just mean is it really needed that the name contains 'ADC'
it could be always DMA2_Stream1_IRQHandler anyway and redefined only for few families.

cores/arduino/stm32/dma.c
LMESTM: The idea was to do the DMA allocation in dma.c and keep the dma actual configuration in each driver ADC and others later. Otherwise this file here will become huge and needs to be updated each time when you need to add DMA support for a new driver. If we make this file simpler, then up to each driver to add the calls to DMA based on the "allocated" DMA information.  My point is not that the code is wrong, but rather the split between the files can be changed

cores/arduino/stm32/dma.c
LMESTM: I agree with proposal to keep the table that is needed but move it to dma.c as this is a table about dma channels and streams, not pins  - is it ok ?

damienWi6labs: The table if different from one board to another. If we move it to dma.c, we will frame it with #ifdef. dma.c will become a huge file. Is it ok for you?

cores/arduino/stm32/dma.c
damienWi6labs: Ok done in next release.

cores/arduino/stm32/dma.c
damienWi6labs: If i understand well, dma_allocate() just give you the irqHandler and the channel to use, then all the initialization of the DMA will be done in analog.c file, in HAL_ADC_MspInit() function.

LMESTM: That would be the idea of the proposal  indeed - please think about it and let let me know if you think it makes sense.  Also I think that the DMA mapping table can be generic and could contain any kind of peripheral instances (ADC, serial, I2C, etc ... ) so the name and types should be generic
Depending on peripherals that are used, the DMA settings will differ (periph to mem, or mem to periph, etc.) so it is rather up to the peripheral driver to set these options.

damienWi6labs: Done, let me know if it's OK. I didn't rebase the git history yet if you want to go back.

cores/arduino/stm32/PeripheralPins.h
```diff
@@ -71,5 +71,8 @@ extern const PinMap PinMap_Ethernet[];
 //*** QUADSPI ***
 extern const PinMap PinMap_QUADSPI[];
 
+//*** DMA ***
+extern const DmaMap DmaMap_DMA[];
+
```
LMESTM: as suggested by @fpistm the DmAMap should be declared and managed in dma.c, not in PeripheralPins as there is no pin in this map.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+  //SPI_HandleTypeDef *spi  = NULL;  // to do
+  
+  switch(periph) {
+    case ADC_PERIPH :
+    {
+      hadc = (ADC_HandleTypeDef*)periph_handle;
+      *channel = pinmap_find_dmaChannel(hadc->Instance, DmaMap_DMA);
+      
+      /* Set clock as active */
+      channelNumber = pinmap_find_dmaChannelNumber(hadc->Instance, DmaMap_DMA);
```
LMESTM: this should not be pinmap_find_dmaChannelNumber but dma_find_dmaChannelNumber

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+      dmaNumber     = pinmap_find_dma(hadc->Instance, DmaMap_DMA);
```
LMESTM: same here

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+  * @brief  DMA IRQ Handle 
+  * @param  None
+  * @retval None
+  */
+
+#ifdef ADC1_BASE
+void DMA2_Stream0_IRQHandler(void)
+{
+  #if defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
+  if (AdcHandleTab[0].DMA_Handle != NULL)
```
LMESTM: from dma point of view this is a HandleTab, we don't need to to know it's a Adc  one.

cores/arduino/stm32/pinmap.c
```diff
@@ -109,3 +109,39 @@ PinName pin_pinName(const PinMap* map) {
     return (PinName)NC;
   }
 }
+
+uint8_t pinmap_find_dma(void* peripheral, const DmaMap* dma) {
```
LMESTM: DMA functions that are not related to pins should be moved to dma.c file

cores/arduino/stm32/pinmap.h
```diff
@@ -36,6 +36,15 @@ typedef struct {
     int function;
 } PinMap;
 
+typedef struct {
```
LMESTM: same for the structure definition

variants/DISCO_F100RB/PeripheralPins.c
```diff
@@ -250,3 +250,8 @@ const PinMap PinMap_SPI_SSEL[] = {
 //*** QUADSPI ***
 
 //*** No QUADSPI ***
+
+const DmaMap DmaMap_DMA[] = {
```
LMESTM: the maps need to be moved as well to a file that is not related to the pins

cores/arduino/stm32/analog.c
```diff
@@ -552,54 +653,86 @@ void HAL_ADC_MspDeInit(ADC_HandleTypeDef* hadc)
...
   ADC_ChannelConfTypeDef  AdcChannelConf = {};
   __IO uint16_t uhADCxConvertedValue = 0;
+  void *peripheral;
+  uint8_t indice = 0;
 
-  AdcHandle.Instance = pinmap_peripheral(pin, PinMap_ADC);
+  peripheral =  pinmap_peripheral(pin, PinMap_ADC);
 
-  if (AdcHandle.Instance == NP) return 0;
+  #ifdef ADC1_BASE
```
LMESTM: getting indice from the peripheral is done in many places, couldn't you create a function where to get indice based on peripheral instance ? 

indice = get_indice(peripheral );
g_current_pinTab[indice] = pin;

that could be applied in many other places

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+  * @brief  This function will init the dma and channel corresponding 
+  *         to the wanted peripheral
+  * @param  periph_handle : pointer to the peripheral HandleTyeDef
+  * @param  periph        : the peripheral type to use
+  * @retval none
+  */
+void dma_allocate(void *periph_handle, Peripheral periph, void **channel, IRQn_Type *irqHandler, uint8_t *indice)
+{
```
LMESTM: in case dma cannot be allocated there is a need to manage an error - how is it done for now ?

LMESTM: Why passing a periph handle and a peripheral type ? 
Can't we pass pass the peripheral Instance directly ? 
dma_allocate(uint32_t *Instance, void **channel, IRQn_Type *irqHandler, uint8_t *indice);
as we know that Instance is actually the peripheral memory address.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+      hadc = (ADC_HandleTypeDef*)periph_handle;
+      *channel = pinmap_find_dmaChannel(hadc->Instance, DmaMap_DMA);
+      
+      /* Set clock as active */
+      channelNumber = pinmap_find_dmaChannelNumber(hadc->Instance, DmaMap_DMA);
+      dmaNumber     = pinmap_find_dma(hadc->Instance, DmaMap_DMA);
+      
+      /* NVIC configuration for DMA Input data interrupt */
+      *irqHandler = pinmap_find_dmaIrqHandler(hadc->Instance, DmaMap_DMA);
+      #ifdef ADC1_BASE
```
LMESTM: Why is the indice managed here and not directly in ADC driver ?
This has nothing to do with DMA and does not seem to be used here.

cores/arduino/stm32/analog.c
```diff
@@ -434,25 +444,114 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
+        
+    #if defined(STM32L0xx) || defined(STM32L4xx)
+    DmaHandleTab[indice].Init.Request             = DMA_REQUEST_0;
+    #endif
+
+    #if defined(STM32F2xx) || defined(STM32F4xx) || defined(STM32F7xx)
+    #ifdef ADC1_BASE
+    if (hadc->Instance == ADC1)
+    {
+      DmaHandleTab[indice].Init.Channel             = DMA_CHANNEL_0;
```
LMESTM: dma channel should have been assigned by dma.c  no ?

cores/arduino/stm32/analog.c
```diff
@@ -434,25 +444,114 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
+    DmaHandleTab[indice].Init.MemBurst              = DMA_MBURST_SINGLE;
+    DmaHandleTab[indice].Init.PeriphBurst           = DMA_PBURST_SINGLE;
+    #endif //STM32F2xx STM32F4xx STM32F7xx
+        
+    /* Deinitialize  & Initialize the DMA for new transfer */
+    HAL_DMA_DeInit(&DmaHandleTab[indice]);
+    HAL_DMA_Init(&DmaHandleTab[indice]);
+
+    /* Associate the DMA handle */
+    __HAL_LINKDMA(hadc, DMA_Handle, DmaHandleTab[indice]);
```
LMESTM: where is DMA_Handle defined ?

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+  * @brief  This function will deinit the dma and channel corresponding 
+  *         to the wanted peripheral
+  * @param  periph_handle : pointer to the peripheral HandleTyeDef
+  * @param  periph        : the peripheral type to use
+  * @retval none
+  */
+void dma_free(void *periph_handle, Peripheral periph)
+{
```
LMESTM: same as allocate, simply pass the Instance instead of handle and periph type.
Or peripheral instance and DMA handle if handle is needed.
In this way you don't need ADC types

cores/arduino/stm32/dma.c
LMESTM: Where is the DMA_Handle table initiliazed with proper DMA_handle values ?

cores/arduino/stm32/analog.c
```diff
@@ -434,25 +444,114 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
+  
+  if (use_dma)
+  {
+    /*##-3- Ask DMA informations ###############################################*/
+    dma_allocate(hadc, ADC_PERIPH, &channel, &irqHandler, &indice);
+
+    /*##-4- Configure the DMA ##################################################*/
+        
+    #if defined(STM32L0xx) || defined(STM32L4xx)
+    DmaHandleTab[indice].Init.Request             = DMA_REQUEST_0;
```
LMESTM: I think that indice has maybe not properly be set in this function.

cores/arduino/stm32/analog.c
```diff
@@ -110,10 +112,12 @@
...
 
-/** @addtogroup STM32F4xx_System_Private_Variables
-  * @{
-  */
+ADC_HandleTypeDef AdcHandleTab[ADC_NUM];
+static DMA_HandleTypeDef DmaHandleTab[ADC_NUM] = {NULL};
+static PinName g_current_pinTab[ADC_NUM];
 static PinName g_current_pin = NC;
+void (*cb[ADC_NUM])(void *user_data);
+static bool use_dma = false;
```
LMESTM: why isn't a "use_dma" for each ADC like use_dma[ADC_NUM] ?

cores/arduino/stm32/analog.c
```diff
@@ -110,10 +112,12 @@
   * @}
   */
 
-/** @addtogroup STM32F4xx_System_Private_Variables
-  * @{
-  */
+ADC_HandleTypeDef AdcHandleTab[ADC_NUM];
+static DMA_HandleTypeDef DmaHandleTab[ADC_NUM] = {NULL};
+static PinName g_current_pinTab[ADC_NUM];
 static PinName g_current_pin = NC;
```
LMESTM: looks like g_current_pin  is not used anymore right ?

cores/arduino/stm32/analog.c
```diff
@@ -809,6 +1079,53 @@ void pwm_stop(PinName pin)
...
 
 
+/**
+  * @brief  Conversion DMA transfer callback in non blocking mode 
+  * @param  hadc: ADC handle
+  * @retval None
+  */
+void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef *AdcHandle)
+{
+  void *user_data = NULL;
```
LMESTM: what's the point of user_data if it is always NULL ?

cores/arduino/wiring_analog.c
```diff
@@ -61,6 +61,21 @@ uint32_t analogRead(uint32_t ulPin)
   return value;
 }
 
+//Start a DMA acquisition on ADC
+//the initialization of the analog PIN is done through this function
+void analogReadDma( uint32_t ulPin, uint32_t *pData, uint32_t lData, void (*callback)(void *user_data))
```
LMESTM: if we use a user_data parameter in the callback, then the user data paramrter must also be passed by application.

cores/arduino/stm32/analog.c
```diff
@@ -110,10 +112,12 @@
...
 
-/** @addtogroup STM32F4xx_System_Private_Variables
-  * @{
-  */
+ADC_HandleTypeDef AdcHandleTab[ADC_NUM];
+static DMA_HandleTypeDef DmaHandleTab[ADC_NUM] = {NULL};
+static PinName g_current_pinTab[ADC_NUM];
 static PinName g_current_pin = NC;
+void (*cb[ADC_NUM])(void *user_data);
+static bool use_dma = false;
```
LMESTM: Maybe it would be useful to have a global ADC structure starting which would contain the ADC handle, the DMA handle,  cb and also use_dma I guess. 

+ADC_HandleTypeDef AdcHandleTab[ADC_NUM];
 +static DMA_HandleTypeDef DmaHandleTab[ADC_NUM] = {NULL};
 +static PinName g_current_pinTab[ADC_NUM];
 +void (*cb[ADC_NUM])(void *user_data);
 +static bool use_dma = false;

If you define your structure in such way that ADC_HandleTypeDef is the first structure parameter, then you can also access the global structure based on the handle. 

something like 
typedef struct {
  ADC_HandleTypeDef AdcHandle;
  DMA_HandleTypeDef DmaHandle;
  PinName current_pin;
  void(*cb)(void *user_data);
  void* user_data;
  bool use_dma;
} adc_context;

static adc_context context(ADC_NUM];

then in MSp_Init/Deinit and also in HAL_ADC_ConvCpltCallback you can get all information like cb: 

void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef *AdcHandle) {
  adc_context* context = (adc_context *) AdcHandle;
  if(context->cb != NULL)
    context->cb(context->user_data);
}

cores/arduino/stm32/dma.c
damienWi6labs: If you don't have the peripheral type, how did you want to cast your void pointer ? 
hadc = (ADC_HandleTypeDef*)periph_handle; Here I need to know i'm working on a ADC

cores/arduino/stm32/analog.c
damienWi6labs: DMA_Handle is part of the struct ADC_HandleTypeDef. You can see it in stm32f3xx_hal_adc.h for example.

cores/arduino/stm32/dma.c
damienWi6labs: This is done by this function, when your link the DMA to a peripheral : 
__HAL_LINKDMA(hadc, DMA_Handle, DmaHandleTab[indice]);

cores/arduino/stm32/dma.c
LMESTM: Why do you want to cast it?
You just need to compare 2 void* Instances 
Use void* as entries in the DmaMap_DMA and use the same Instance address as void* type for comparing.

damienWi6labs: By doing this, you make the assumption that instance is always at the beginning of the peripheral struct. Is it always true for all peripheral? Can we be sure it will not change in the future?

cores/arduino/stm32/analog.c
LMESTM: ok - I got confused by the way __HAL_LINKDMA macro is built. thanks

cores/arduino/stm32/dma.c
LMESTM: Ok same as previous - I have lloked at __HAL_LINKDMA and this is now clearer.

cores/arduino/stm32/dma.c
LMESTM: No. Here I'm proposing to change the function to:
dma_allocate(void *Instance, void **channel, IRQn_Type *irqHandler, uint8_t *indice)
So the caller will pass the Instance as a parameter. 
As far as I see, you are only using hadc->Instance in this function, so you don't need the hadc handle, you just need the instance.

damienWi6labs: OK my bad, i didn't see you change the prototype. Yes ok, with that it can work, only pass the instance and not the entire struct.

cores/arduino/stm32/dma.c
damienWi6labs: For dma_free, i try to remove the peripheral type, but I've got a problem for that call : 
HAL_DMA_DeInit(hadc->DMA_Handle);
I can't DeInit the DMA. When i think about it, it's not a big deal, because we stop peripheral interruption by doing this : 
HAL_NVIC_DisableIRQ(irqHandler);
So this peripheral will not generate other interruption, and the DMA will not have other wok to do but it is always initialize. 
What do think about that? Is it ok for you?

cores/arduino/stm32/analog.c
damienWi6labs: It's used for DAC function, I don't touch to this part. See for example function HAL_DAC_MspInit() or dac_write_value().

cores/arduino/stm32/analog.c
damienWi6labs: Here we don't need a table. I explain myself. When you want to use ADC peripheral with or without DMA, the same function was called : HAL_ADC_MspInit(). In this function I do the DMA part only if the call come from the dma function (adc_read_value_dma). The boolean dma_use fulfilled this role.

cores/arduino/stm32/analog.c
damienWi6labs: dma_allocate set the indice, so it's ok.

cores/arduino/stm32/dma.c
LMESTM: I think that you should pass dma_free(Instance, DmaHandle) as both are known at driver level.

cores/arduino/stm32/analog.c
LMESTM: hmm ok - can we use the same DAC in DMA then without DMA ? Isn't it supposed to be the same pin .

cores/arduino/stm32/analog.c
damienWi6labs: dma.c will assign channel, yes this is correct. For all microcontroller,  we talk about channel excepted for STM32F2xx, STM32F4xx, STM32F7xx. For this we talk about stream, that with Irqhandler name are different for this families. A channel in STM32F1xx is a stream on STM32F2xx. 
Moreover, this 3 families have also channel, but not at the same way of other microcontroller. This can be confusing, I took a while to understand but I invite you to consult the datasheets.

cores/arduino/stm32/analog.c
LMESTM: so does it cover the case where ADC_1 is used with DMA and ADC_2 is used without DMA ?

cores/arduino/stm32/analog.c
LMESTM: that's another comment I made, dma_allocate should not set the ADC indice. This is ADC specific and should be done in this function - don't you think ?

cores/arduino/stm32/analog.c
LMESTM: ok thanks  I will need to have a look at datasheets thanks.
Also I will check again when I review your next update

cores/arduino/stm32/analog.c
damienWi6labs: yes

cores/arduino/stm32/analog.c
damienWi6labs: The indice depend on how many ADC we have, so yes it's more specific to ADC than DMA. I can move it to analog file. So prototype of dma_allocate() becomes :
void dma_allocate(void *instance, void **channel, IRQn_Type *irqHandler)

cores/arduino/stm32/analog.c
damienWi6labs: Do you want to use g_current_pinTab for DAC too? I prefer to leave as it is because DAC and ADC are two different peripherals.

cores/arduino/stm32/analog.c
LMESTM: indeed this looks better to me

cores/arduino/stm32/analog.c
LMESTM: Ok - no please keep them separated if they're used for different purpose. But would be great to have at a least a comment to explain that one is used for ADC and the other one is used for DAC because as of now they have exactly the same name.

cores/arduino/stm32/dma.c
damienWi6labs: Done

cores/arduino/stm32/dma.c
damienWi6labs: done

cores/arduino/stm32/dma.c
damienWi6labs: Done, indice manage in ADC

cores/arduino/stm32/analog.c
damienWi6labs: Done. New release available.

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,360 @@
...
+  {ADC1, 2, DMA2_Stream0, DMA2_Stream0_IRQn, 0},
+  {ADC1, 2, DMA2_Stream4, DMA2_Stream4_IRQn, 4},
+  #if defined(STM32F2xx) || defined(STM32F429xx) || defined(STM32F407xx)
+  {ADC2, 2, DMA2_Stream2, DMA2_Stream2_IRQn, 2},
+  {ADC2, 2, DMA2_Stream3, DMA2_Stream3_IRQn, 3},
+  {ADC3, 2, DMA2_Stream0, DMA2_Stream0_IRQn, 0},
+  {ADC3, 2, DMA2_Stream1, DMA2_Stream1_IRQn, 1},
+  #endif
+  #endif
+};
```
LMESTM: isn't the table supposed to end with NP entry ?

cores/arduino/stm32/analog.c
```diff
@@ -110,11 +112,52 @@
...
   */
 
-/** @addtogroup STM32F4xx_System_Private_Variables
-  * @{
-  */
+/* Adc context                 */
+AdcContext context[ADC_NUM];
+
+/* DMA is use, or classic ADC  */
+static bool use_dma = false;
```
LMESTM: I still don"t understand why use_dma is not defined for each instance ... how does it work in case ADC1 is used in DMA mode wile ADC2 is used without DMA.

cores/arduino/stm32/analog.h
```diff
@@ -47,6 +57,15 @@
  extern "C" {
 #endif
 
+typedef struct {
+  ADC_HandleTypeDef AdcHandle;
```
LMESTM: We're making sure that AdcHandle is the first struct inside AdcContext => add a comment to make sure this is always true.

cores/arduino/stm32/analog.c
```diff
@@ -809,6 +1066,52 @@ void pwm_stop(PinName pin)
...
 
 
+/**
+  * @brief  Conversion DMA transfer callback in non blocking mode 
+  * @param  hadc: ADC handle
+  * @retval None
+  */
+void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef *AdcHandle)
+{
+  /* Report to main program that ADC sequencer has reached its end */
```
LMESTM: Because we're making sure that ADC_HandleTypeDef  is the first struct in the AdContext struct this is a trick to retrieve the AdcContext from the handle used in HAL.
you should be able to code as below:

AdcContext context = (AdcContext *) AdcHandle;
contex.cb(context.user_data);

cores/arduino/stm32/analog.c
damienWi6labs: Ok right, i see where is the problem. My thing works only for initialization, but I have lost the use_dma information when I wat Deinit it. So yes it's better to make a use_dma for each instance.

cores/arduino/stm32/dma.c
damienWi6labs: Yes. Done.

cores/arduino/stm32/analog.c
damienWi6labs: ok. Done. New release available.

LMESTM:
Thanks @damienWi6labs  
Can you provide also a few lines in the PR description of what was tested with this release  ?

LMESTM:
@fpistm - anything gating this PR ?

fpistm:
My review of DMA API...

cores/arduino/stm32/analog.c
```diff
@@ -434,25 +480,117 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
   HAL_GPIO_Init(port, &GPIO_InitStruct);
+  
+  indice = get_indice(hadc->Instance);
+  
+  if (context[indice].use_dma)
+  {
+    /*##-3- Ask DMA informations ###############################################*/
+    if (dma_allocate(hadc->Instance, &channel, &irqHandler))
+    {
+      /*##-4- Configure the DMA ##################################################*/
```
fpistm: DMA configuration have to be done by the driver  in dma.c
`dma_init()`

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,361 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+  ******************************************************************************
+  */
+
+#include "stm32_def.h"
+#include "dma.h"
+#include "analog.h"
+
+const DmaMap DmaMap_DMA[] = {
```
fpistm: ADC DMA mapping have to be in analog part:
ADC1 , 1 ,1  --> ADC1 use DMA1 channel 1

In dma: only the dma handle should be stored

cores/arduino/stm32/analog.c
```diff
@@ -543,6 +683,14 @@ void HAL_ADC_MspDeInit(ADC_HandleTypeDef* hadc)
 #ifdef __HAL_RCC_ADC_CLK_DISABLE
   __HAL_RCC_ADC_CLK_DISABLE();
 #endif
+
+  indice = get_indice(hadc->Instance);
+
+  if (context[indice].use_dma)
+  {
+    dma_free(hadc->Instance);
+    HAL_DMA_DeInit(hadc->DMA_Handle);
```
fpistm: dma_free should be called after deinit ?

fpistm:
Hi @fprwi6labs 
any news on this PR?

fprwi6labs:
Hi @fpistm 
This PR can be closed because the thread moved to PR #231.

fpistm:
Hi @fprwi6labs,
Thanks.
Closed replaced by #231

PR #177, opened 2017-12-13
Add RTC driver

fprwi6labs:
Driver to use the internal RTC:
- Enable/disable RTC
- Set/Get time & calendar
- Enable/disable alarm
- Attach/detach alarm callback

This driver is linked with the [STM32RTC](https://github.com/stm32duino/STM32RTC/pull/1) library.


variants/NUCLEO_F207ZG/stm32f2xx_hal_conf.h
```diff
@@ -175,7 +175,7 @@
 
 /* Section 2: PHY configuration section */
 
-/* DP83848 PHY Address*/ 
+/* DP83848 PHY Address*/
```
LMESTM: You're mixing up 2 things in the commit - clean-up of various white spaces with the actual changes for RTC.
In such cases, please try to make 2 separate commits instead ...

cores/arduino/stm32/rtc.c
```diff
@@ -0,0 +1,426 @@
...
+#endif // !defined(STM32F1xx)
+  }
+}
+
+/**
+  * @brief Attach alarm callback.
+  * @param func: pointer to the callback
+  * @retval None
+  */
+void attachAlarmCallback(void (*func)(void))
```
LMESTM: rather than 
void (*func)(void)
it could be useful from application to offer a callback of type
void (*func)(void* data)

fprwi6labs: What information would you like to pass in parameter? rtc handler? other?

LMESTM: This is up to application .  It could the time when alarm was set, or a pointer to the led to be turned on. The application provides a pointer and you simply pass it back to application. This is useful for callback mecanisms.  It would be particularly useful if setting several alarms.

fprwi6labs: It is not the application that choice this parameter. It is the driver which can return a parameter.
If you don't say me the parameter to return, I think return a void* is useless.
Or I must change the prototype of attachAlarmCallback to take account another parameter with the data pointer (also useless).

LMESTM: sorry if unclear, the user data is passed from application to driver, stored in driver and then passed back from driver to application. The driver doesn't know which informaition it contains.

This can be useful for coding generic callback where you don't need to store application data as globals or in context.
https://arduino.stackexchange.com/questions/23063/user-callback-functions
https://stackoverflow.com/questions/14270846/passing-a-parameter-in-a-callback-in-c

In the case of RTC alarm, this can be useful if application can set 2 different alarms in future, then application could pass the programmed time, or a different LED or whatever it needs

fprwi6labs: It is clearer thank you. And done.

variants/NUCLEO_F207ZG/stm32f2xx_hal_conf.h
fprwi6labs: Done!

cores/arduino/stm32/rtc.c
```diff
@@ -0,0 +1,446 @@
...
+
+#ifdef __cplusplus
+ extern "C" {
+#endif
+
+/* Private define ------------------------------------------------------------*/
+/* RTC prediv are set to the default values to match with the LSE clock source.
+  If the clock source is modified, those prescalers must be changed */
+/* Synchonuous prediv  */
+#define PREDIV_S                  0x00FF
```
fpistm: Main restriction of the driver is the RTC clock configuration.
The driver have to be able to dela with all possible source (HSE, LSE,LSI).
Refer to [AN4759 Overview of the STM32 MCU advanced RTC](http://www.st.com/resource/en/application_note/dm00226326.pdf)
§ 1.1.4         RTC         clock         configuration

fpistm: Main restriction of the driver is the RTC clock configuration.
The driver have to be able to dela with all possible source (HSE, LSE,LSI).
Refer to [AN4759 Overview of the STM32 MCU advanced RTC](http://www.st.com/resource/en/application_note/dm00226326.pdf)
§ 1.1.4         RTC         clock         configuration

cores/arduino/stm32/rtc.c
```diff
@@ -0,0 +1,446 @@
...
+  *        This function configures the hardware resources used. By default the
+  *        LSE clock is used as RTC clock source.
+  * @param hrtc: RTC handle pointer
+  * @note  Care must be taken when HAL_RCCEx_PeriphCLKConfig() is used to select
+  *        the RTC clock source; in this case the Backup domain will be reset in
+  *        order to modify the RTC Clock source, as consequence RTC registers (including
+  *        the backup registers) and RCC_CSR register are set to their reset values.
+  * @retval None
+  */
+void HAL_RTC_MspInit(RTC_HandleTypeDef *hrtc)
```
fpistm: ditto

cores/arduino/stm32/rtc.c
```diff
@@ -0,0 +1,446 @@
...
+}
+
+/**
+  * @brief RTC Initialization
+  *        This function configures the RTC time and calendar. By default, the
+  *        RTC is set to the 1st January 2017 0:0:0:00
+  * @param format: enable the RTC in 12 or 24 hours mode
+  * @retval None
+  */
+void RTC_init(hourFormat_t format)
```
fpistm: Have to mange RTC clock config

fprwi6labs:
RTC driver updated with clock source selection.
The RTC clock source can be set: LSE, LSI or HSE. 
User can set the prescaler values for his own configuration. The user clock configuration is not overwritten.
Default prescaler values are provided for the most common hardware clock configuration. 

Compilation OK. 
Tests OK for LSE and LSI. HSE not tested. 
Tested only on boards NUCLEO L0, L4, F0, F1, F3, F4.
L1, F2 & F7 missing.

cores/arduino/stm32/rtc.c
```diff
@@ -0,0 +1,492 @@
...
+  UNUSED(format);
+#else
+  if(format == HOUR_FORMAT_12) {
+    RtcHandle.Init.HourFormat = RTC_HOURFORMAT_12;
+  } else {
+    RtcHandle.Init.HourFormat = RTC_HOURFORMAT_24;
+  }
+  RtcHandle.Init.OutPut = RTC_OUTPUT_DISABLE;
+  if(source == LSE_CLOCK) {
+    RtcHandle.Init.AsynchPrediv = LSE_ASYNCH_PREDIV;
```
fpistm: [HSE|LSE|LSI]_A?SYNC_PREDIV should not be hard-coded. User should have a way to dynamically change them without add a dedicated define. At least an API to change the default value.
Anyway, as mentioned in rtc.h, it should be possible to define the best PREDIV

cores/arduino/stm32/rtc.h
```diff
@@ -0,0 +1,181 @@
...
+#define IS_RTC_HOUR12(HOUR)      IS_RTC_HOUR24(HOUR)
+#endif // defined(STM32F1xx) && !defined(IS_RTC_WEEKDAY)
+
+/* RTC prescalers are set to obtain the RTC clock to 1Hz. See AN4759.
+  Default prescalers are calculated for the following clock speed:
+  - LSI = 32, 37 or 40 kHz
+  - LSE = 32.768 kHz
+  - HSE = 8 MHz
+
+ Prescalers should be set by user in case the clock configuration is different.
```
fpistm: It should be possible to define [HSE|LSE|LSI]_A?SYNC_PREDIV as for each variant clock values are defined.
Ex: https://github.com/stm32duino/Arduino_Core_STM32/blob/2415879ab2472618918a8f2ca054996d54cb21d4/variants/NUCLEO_L476RG/stm32l4xx_hal_conf.h#L159

lacklustrlabs:
Have you considered adding an `attachAlarmCallback(std::function)` function?
I guess that will introduce an additional, potentially costly, C/C++ interfacing layer.
What is your opinion?

fpistm:
@lacklustrlabs 
I  do not considered this but it could. This has been done for the interrupt.
I'm not very aware about that and it's use/utility but why not.

lacklustrlabs:
@fpistm 
As you know std::function is a generic way of handling callable objects in C++ (functions, lambda expressions, bind expressions).

I guess std::function callbacks can be added at any time later on, the only thing that should be considered now is if std::function callbacks are good (fast and memory efficient) enough to replace C-style callbacks. If so, the switch-over should be made now, before the API is released.

If you want plain C-style callbacks in the API, the std::function callbacks can be added at any time later. The downside of this option is a, possible, duplicated callback handling.

fpistm:
Hi @fprwi6labs 
any news on this PR?

fpistm:
@lacklustrlabs 
It could be considered.
Maybe @fprwi6labs have an opinion on this

fprwi6labs:
@fpistm 
Prescalers are now dynamically calculated.
Compilation OK for all available boards but not tested with HSE clock source.

@lacklustrlabs 
Is it possible to use std::function<void(void *)>? I tried but without success and I didn't find example.
So I will not implement this feature for now.

PR #181, opened 2017-12-19
Added force rebuild menu

fpistm:
Allow to force a rebuild, useful with build_opt.h feature. 
When updating build_opt.h file, Arduino IDE does not detect change in the source then do not rebuild and new build options are ignored. This avoid to restart IDE to take those changes in account.

Required aggressively cache compiled core to be disabled
else no force rebuild.

Need to be tested on Linux/MacOSx

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -76,6 +76,7 @@ build.xSerial=
 build.enable_usb=
 build.flags.optimize=
 build.flags.ldspecs=
+build.force=0
 
 # Pre and post build hooks
 build.opt.name=build_opt.h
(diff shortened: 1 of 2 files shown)


platform.txt
```diff
@@ -87,6 +88,11 @@ recipe.hooks.prebuild.1.pattern.windows=cmd /c "if not exist {build.opt.sourcepa
 recipe.hooks.prebuild.1.pattern.linux=bash -c "[ -f {build.opt.sourcepath} ] || (mkdir -p {build.path}/sketch && touch {build.opt.path})"
 recipe.hooks.prebuild.1.pattern.macosx=bash -c "[ -f {build.opt.sourcepath} ] || (mkdir -p {build.path}/sketch && touch {build.opt.path})"
 
+# Check if force rebuild required (required aggressively cache compiled core to be disabled)
+recipe.hooks.sketch.prebuild.1.pattern.windows=cmd /c "if {build.force} equ 1 ( if exist {build.path}\core del /Q /F /S {build.path}\core {build.path}\libraries > nul 2>&1)"
+recipe.hooks.sketch.prebuild.1.pattern.linux=bash -c "[ {build.force} -eq 0 ] || [ ! -f {build.path}/core ] || rm -fr {build.path}/core {build.path}/libraries"
+recipe.hooks.sketch.prebuild.1.pattern.macosx=bash -c "[ {build.force} -eq 0 ] || [ ! -f {build.path}/core ] || rm -fr {build.path}/core {build.path}/libraries"
```
lacklustrlabs: For bash, use `[ ! -d {build.path}/core ]`
With this modification it works for OSx (High Sierra). Haven't tested linux yet.

fpistm: It seems logical. 
Anyway, I think about new way to force the rebuild without a menu. I need to test if it can work.
I do not like this kind of workaround/stub to  change way of IDE handle rebuild or not.
And yes when a menu option is changed a full rebuild is performed as menu parameters are changed.

platform.txt
```diff
@@ -76,6 +76,7 @@ build.xSerial=
 build.enable_usb=
 build.flags.optimize=
 build.flags.ldspecs=
+build.force=0
```
lacklustrlabs: Does this do anything?

fpistm:
Set on hold as not really proper to add this kind of IDE workaround.

PR #193, opened 2018-01-09
Added setXXX methods to change used pins of an instance

fpistm:
Example to change pins used by Wire instance (by default use defined SDA/SCL ):
```
Wire.setSDA(A4);
Wire.setSCL(PC2);
Wire.begin();
```

Example to change pins used by SPI instance (by default use defined MISO/MOSI/SCK):
```
SPI.setMOSI(22);
SPI.setMISO(PA3);
SPI.begin();
```

### DIFF
--- a/libraries/SPI/keywords.txt
+++ b/libraries/SPI/keywords.txt
@@ -17,7 +17,10 @@ transfer		KEYWORD2
 #setBitOrder	KEYWORD2
 setDataMode		KEYWORD2
 setClockDivider	KEYWORD2
-
+setMISO			KEYWORD2
+setMOSI			KEYWORD2
+setSCLK			KEYWORD2
+setSSEL			KEYWORD2
 
 #######################################
 # Constants (LITERAL1)
--- a/libraries/SPI/src/SPI.h
+++ b/libraries/SPI/src/SPI.h
@@ -112,6 +112,17 @@ class SPIClass {
     SPIClass();
     SPIClass(uint8_t mosi, uint8_t miso, uint8_t sclk, uint8_t ssel = (uint8_t)NC);
 
+    // setMISO/MOSI/SCLK/SSEL have to be called before begin()
+    void setMISO(uint32_t miso) { _spi.pin_miso = digitalPinToPinName(miso); };
+    void setMOSI(uint32_t mosi) { _spi.pin_mosi = digitalPinToPinName(mosi); };
+    void setSCLK(uint32_t sclk) { _spi.pin_sclk = digitalPinToPinName(sclk); };
+    void setSSEL(uint32_t ssel) { _spi.pin_ssel = digitalPinToPinName(ssel); };
+
+    void setMISO(PinName miso) { _spi.pin_miso = (miso); };
+    void setMOSI(PinName mosi) { _spi.pin_mosi = (mosi); };
+    void setSCLK(PinName sclk) { _spi.pin_sclk = (sclk); };
+    void setSSEL(PinName ssel) { _spi.pin_ssel = (ssel); };
+
     void begin(uint8_t _pin = CS_PIN_CONTROLLED_BY_USER);
     void end(void);
 
--- a/libraries/Wire/keywords.txt
+++ b/libraries/Wire/keywords.txt
@@ -17,6 +17,8 @@ endTransmission		KEYWORD2
 requestFrom		KEYWORD2
 onReceive		KEYWORD2
 onRequest		KEYWORD2
+setSCL			KEYWORD2
+setSDA			KEYWORD2
 
 #######################################
 # Instances (KEYWORD2)
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -67,6 +67,11 @@ class TwoWire : public Stream
   public:
     TwoWire();
     TwoWire(uint8_t sda, uint8_t scl);
+    // setSCL/SDA have to be called before begin()
+    void setSCL(uint32_t scl) { _i2c.scl = digitalPinToPinName(scl); };
+    void setSDA(uint32_t sda) { _i2c.sda = digitalPinToPinName(sda); };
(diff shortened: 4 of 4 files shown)


romainreignier:
That change seems very helpful but it there any mechanism to check if the selected pin has the right capability?

fpistm:
Pins capabilities are checked during the `begin()` and print an error message.
For SPI:
https://github.com/stm32duino/Arduino_Core_STM32/blob/cf03b789508776fcb36ca623b5874fbdddc27083/cores/arduino/stm32/spi_com.c#L205
For I2C:
https://github.com/stm32duino/Arduino_Core_STM32/blob/cf03b789508776fcb36ca623b5874fbdddc27083/cores/arduino/stm32/twi.c#L166

Unfortunately the standard API `begin()` from Arduino do not return any value.
Mainly up to the user to ensure pins used have the right capabilities.

mayukhIITD:
Yes I would agree to @romainreignier .. For example in my case I needed to set the SDA SCL pins to A4,A5.. But I am not able to do that.. Because I am using IKSA102 board for temperature,Humidity,Pressure sensing which uses I2C and is probably using either of A4,A5.. Hence when I set  SDA SCL pins to A4,A5 ,I am getting wrong values for Temperature and Humidity..

fpistm:
@mayukhIITD
If you need several pins to I2C you have to instantiate a new Wire instance.
Main issue could come from Arduino libraries which use the standard `Wire` instance.
I you use 2 libraries using this instance,  you have to modify the library to use a different instance.

romainreignier:
@fpistm Ah, that's great that the pins are being checked but it is at runtime and the user need a serial console opened, right?

It would be nice to check at compile time with [static_assert](http://en.cppreference.com/w/cpp/language/static_assert) but will prevent to change the pins on the fly :(
But if it is possible to instantiate several instances of the peripherals, it may be done.

fpistm:
Yes, right.
Core could not check all possible use case.
As the main goal of this core is to be generic for all STM32 MCU, it's quite hard to perform this kind of check at compile time. Peripherals linked to a pin are stored in several arrays in PeripheralPins.c

romainreignier:
Ok, I see.

But anyway, this change is great and will ease the usage of these peripherals :+1:

lacklustrlabs:
I was playing around with compile-time checks on Wire.setSDA() and Wire.SDL(), this is what I came up with:

Wire.h:
```
   ....
    void setSDA(PinName sda) { _i2c.sda = sda; };
   + 
   +template <uint32_t scl>
   +void setSCL() {
   +	  static_assert(digitalPinHasI2CSCL(scl), "SCL must be a supported pin");
   +	  _i2c.scl = digitalPinToPinName(scl); 
   +};
   +
   + template <uint32_t sda>
   + void setSDA() { 
   +	  static_assert(digitalPinHasI2CSDA(sda), "SDA must be a supported pin");
   +  	  _i2c.sda = digitalPinToPinName(sda); 
   +  };
        
    void begin();
    ... 
```

variant.h:
```
...
+constexpr bool digitalPinHasI2CSCL(uint32_t scl) {
+  return scl==1||scl==14||scl==11; // just an example
+}

+constexpr bool digitalPinHasI2CSDA(uint32_t sda) {
+  return sda==2||sda==15||sda==11; // just an example
+}

#endif /*  __cplusplus */


#endif /* _VARIANT_ARDUINO_STM32_ */

```

The.ino
```
...
void setup()
{
  +Wire.setSCL<1>();
  +Wire.setSDA<2>();
  Wire.begin(); // join i2c bus (address optional for master)
}
....
```
This would require some work to manually edit each variant.h (add the I2C & SPI constexpr functions)
 
Disclamer: I've only tested that it compiles

fpistm:
Nice code snippets ;)
I think it too much work for no real benefits.
pins are hardcoded, while the pin mapping allows to get capabilities dynamically qnd depending on the PeripheralPins config the capabilities could be enabled or not.
This is customizable by the user so the risk is to add some misalignement.
```
const PinMap PinMap_I2C_SDA[] = {
//  {PB_7,   I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
    {PB_9,   I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)}, // D14
//  {PB_11,  I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
//  {PC_9,   I2C3, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C3)},
//  {PD_13,  I2C4, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C4)},
//  {PF_0,   I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
//  {PF_15,  I2C4, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C4)},
//  {PH_5,   I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
//  {PH_8,   I2C3, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C3)},
//  {PH_12,  I2C4, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C4)},
    {NC,    NP,    0}
};
```
anyway, User has to know which pins are used for the wiring of it's I2C device.

LMESTM:
agree with @fpistm  - compile time check looks a bit overkill to me. What really matters is that user correctly check schematics and wiring of the hardware board - that's usually were most encountered errors comes from ...

Issue #140, opened 2017-10-25
NUM_DIGITAL_PINS and NUM_ANALOG_INPUTS breaks cpp conditionals


bperrybap:
The definitions of NUM_DIGITAL_PINS and NUM_ANALOG_INPUTS below:
```
#define NUM_DIGITAL_PINS            ((uint32_t)DEND)
#define NUM_ANALOG_INPUTS ((uint32_t)(AEND-A0))
```
breaks code that use cpp conditionals like this:
`#if NUM_ANALOG_INPUTS > 4`

I've not run into this issue in any other core.

The cpp processor is not type aware and so the uint32_t cast is what breaks things.
When the cast is removed, the cpp conditionals that use the macros will work.

I'm not sure why the cast would be necessary in the macros.
enums can normally be used as integers, and the NUM_ANALOG_INPUTS is already depending on that in the calculation.

If there are some places that get warnings because of some sort of int vs unsigned int comparisons, it seems like it should be fixed there vs in the actual macro definitions.

fpistm:
Right it was added to avoid warning.
So, I will remove this cast here and correct all warning raised.

fpistm:
Hi @bperrybap,
in fact use the NUM_DIGITAL_PINS and NUM_ANALOG_INPUTS is not possible for conditional preprocessing as they are enum values (A0, AEND, PEND).
Instead of use:
`#if NUM_ANALOG_INPUTS > 4`
you will have to use standard comparaison:
`if ( NUM_ANALOG_INPUTS > 4)`
I could remove the cast but the `#if NUM_ANALOG_INPUTS > 4` will be never evaluated `true`

bperrybap:
Bummer. when I did my simple test, I only tested that it compiled not that it worked.

The issue in my case is that I need  a test that works at compile time.
My code is looking to see if symbols A4 and A5 exist in the core and it uses the NUM_ANALOG_INPUTS to detect this. If the check is moved to run time, then while it would work for the STM core it, or any core that has the symbols,  would fail on any core that didn't define the symbols A4 and A5 since A4 and A5 would always be referenced.
And unfortunately those symbols are sometimes defines and sometimes const ints so there is no way to use the A4/A5 symbols themselves as conditionals to detect it.

I'm not sure how much other code out there does this type of compile time checking using these symbols.

Its a tough problem for something so simple.

[...]

PR #206, opened 2018-01-23
Assigning literal values to NUM_DIGITAL_PINS and NUM_ANALOG_INPUTS

lacklustrlabs:
It turned out that I needed the fix of #140 for a pet project of mine. 
So here are the changes I made.

I've compiled every board with this .ino:
```
#if NUM_DIGITAL_PINS > 4
#else
#error "NUM_DIGITAL_PINS not fixed"
#endif

#if NUM_ANALOG_INPUTS > 1
#else
#error "NUM_ANALOG_INPUTS not fixed"
#endif

void setup() {
}

void loop() {
} 
```

### DIFF
--- a/cores/arduino/pins_arduino_var.h
+++ b/cores/arduino/pins_arduino_var.h
@@ -25,14 +25,16 @@
 // Include board variant
 #include "variant.h"
 
+// Avoid pins number misalignment
+_Static_assert(NUM_DIGITAL_PINS==PEND, "NUM_DIGITAL_PINS and PEND differ!");
+_Static_assert(NUM_ANALOG_INPUTS==(AEND-A0), "NUM_ANALOG_INPUTS and (AEND-A0) differ!");
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 #define NOT_AN_INTERRUPT            NC // -1
-#define DEND                        PEND
-#define NUM_DIGITAL_PINS            ((uint32_t)DEND)
-#define NUM_ANALOG_INPUTS           ((uint32_t)(AEND-A0))
+#define DEND                        NUM_DIGITAL_PINS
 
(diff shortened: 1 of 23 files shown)


fpistm:
Thanks @lacklustrlabs 
First check seems ok. I need to review it more deeper😉
It is the first part of the #140 which require also Ax pins definition review

lacklustrlabs:
imho it would be easiest to add the first PIN_AX variant:
```
#if NUM_ANALOG_INPUTS > 0
#define PIN_A0 A0
#endif
#if NUM_ANALOG_INPUTS > 1
#define PIN_A1 A1
#endif
...
#if NUM_ANALOG_INPUTS > 18
#define PIN_A18 A18
#endif
```
I can do that, but the question is where do you want it?

fpistm:
Thanks @lacklustrlabs 
I have to check that before doing any change for Ax definition. I would like check how other variant do it.
Anyway, I could merge your PR before doing that.
I have only to review code and made some tests. (firmata, ....)

cores/arduino/pins_arduino_var.h
```diff
@@ -25,14 +25,16 @@
 // Include board variant
 #include "variant.h"
 
+// Avoid pins number misalignment
+_Static_assert(NUM_DIGITAL_PINS==PEND, "NUM_DIGITAL_PINS and PEND differ!");
+_Static_assert(NUM_ANALOG_INPUTS==(AEND-A0), "NUM_DIGITAL_PINS and (AEND-A0) differ!");
```
fpistm: Sorry made a wrong copy/paste
"~NUM_DIGITAL_PINS~NUM_ANALOG_INPUTS and (AEND-A0) differ!"

lacklustrlabs: No problem :)

PR #208, opened 2018-01-27
[Arduino compatibility] Ax pin definition

### LINKED ISSUE #140: NUM_DIGITAL_PINS and NUM_ANALOG_INPUTS breaks cpp conditionals

fpistm:
Ax definition is now inline with Arduino style
Variant has only to define the digital pin number of the first analog input  (i.e. which digital pin is A0)
Clean variant header include

Fix #140


lacklustrlabs:
Can't help feeling that converting from an enum to uint8_t for `Ax` is a step back, but that's the burden of legacy code i guess.

Somewhat unrelated to this PR, but in the same files:
In C++ we can have typed enums, is it possible to use an unsigned type for the `Dx` enum or will that break with the C-compatibility?

fpistm:
I think enum could be kept but at this time I will kept legacy code. About Dx, I will check about C compatibility.

lacklustrlabs:
Maybe this is the wrong place to discuss this, but here goes:

```
// int8_t is clearly the wrong type for this, but it's a likely typo 
int8_t p1 = (128);  // no warning, silent fail
int8_t p2 = (128u); // no warning, silent fail
int8_t p3 = D128;   // The current enum: no warning, silent fail
digitalWrite(p3, HIGH); // escalating the fail: pin = 4294967168
int8_t p4 = {128u}; // error: narrowing conversion of '128' from 'unsigned int' to 'int8_t {aka signed char}' inside { } [-Wnarrowing]
```
I prefer to make the compiler find my inevitable coding errors, so maybe something like this could work?
```
#ifdef __cplusplus
#define D0 {0u}
#define D1 {1u}
...
#else
#define D0 (0u)
#define D1 (1u)
...
#endif
```
I guess this is a rather low priority issue, something for the back burner..

Issue #212, opened 2018-01-30
I2C slave failed


larsstenberg:
BOARD: NUCLEO-F091RC

Connected a NanoPi Neo Air (master) with the nucleo when running i2cset-y 0 0x09 on the NPI the nucleo hangs. Serial stops, blinking stop. It just freezes.

Code running on the nucleo:
```
#include <Wire.h>

void setup()
{
  Wire.begin(0x15);                // join i2c bus with address #4
  Wire.onReceive(receiveEvent); // register event
  Wire.onRequest(requestEvent);
  Serial.begin(115200);           // start serial for output
}

void loop()
{
  delay(100);
}

void receiveEvent(int howMany)
{
  Serial.println(howMany);
  while(1 < Wire.available()) // loop through all but the last
  {
    int c = Wire.read(); // receive byte as a character
    Serial.print(c);         // print the character
  }
  int x = Wire.read();    // receive byte as an integer
  Serial.print(c);         // print the character
}

void requestEvent()
{
  Wire.write("hello");
}
```

1) It seems to be a problem in the Wire.c when running as a slave the allocateRxBuffer never gets called. So first fix seems to be to add:

```
  // copy twi rx buffer into local read buffer
  // this enables new reads to happen in parallel
+  if(rxBuffer == nullptr){
+    allocateRxBuffer(BUFFER_LENGTH);
+  }
  memcpy(rxBuffer, inBytes, numBytes);
```

and changing `allocateRxBuffer` to static.

Now the freeze stopped.

2) It also seems to be an error with the numBytes received from master.
Changing line 541 in twi.c
from: `nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferCount;`
to: `nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferSize;`

Seems to correct the problem.

Am I doing something wrong or this a actual bug?

fpistm:
Hi @larsstenberg 
yes seems to be a bug :) thanks to report.

fpistm:
Hi @larsstenberg 
I've tested Master/receiver Wire example and it is now ok.
Could you review the patch, please?
Thanks in advance

PR #216, opened 2018-02-02
Fix I2C slave issue

fpistm:
See #212

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -539,7 +539,7 @@ void HAL_I2C_ListenCpltCallback(I2C_HandleTypeDef *hi2c)
 
   if((obj->i2c_onSlaveReceive != NULL) &&
     (obj->slaveMode == SLAVE_MODE_RECEIVE)) {
-    nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferCount;
+    nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferSize;
     if(nbData != 0) {
       obj->i2c_onSlaveReceive(obj->i2cTxRxBuffer, nbData);
     }
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -381,13 +381,12 @@ void TwoWire::onReceiveService(uint8_t* inBytes, int numBytes)
     return;
   }
 
+  allocateRxBuffer(numBytes);
+  // error if no memory block available to allocate the buffer
   if(rxBuffer == nullptr){
-    allocateRxBuffer(numBytes);
-    // error if no memory block available to allocate the buffer
-    if(rxBuffer == nullptr){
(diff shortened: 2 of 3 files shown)


cores/arduino/stm32/twi.c
```diff
@@ -539,7 +539,7 @@ void HAL_I2C_ListenCpltCallback(I2C_HandleTypeDef *hi2c)
 
   if((obj->i2c_onSlaveReceive != NULL) &&
     (obj->slaveMode == SLAVE_MODE_RECEIVE)) {
-    nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferCount;
+    nbData = I2C_TXRX_BUFFER_SIZE - obj->handle.XferSize;
```
LMESTM: Aren't you using HAL_I2C_SlaveRxCpltCallback and HAL_I2C_SlaveTxCpltCallback ?

fpistm: no, only HAL_I2C_ListenCpltCallback

LMESTM: Ok. then it's not easy for me to understand the code. If this works - fine. But I think  ListenCpltCallback does not mean that data has been transmitted already, so I don't understand the lines above and why XferCount is replaced by XferSize

fpistm: Right, I think, that we missed that when @fprwi6labs do the #135.
While I2C is sequential it seems to be OK but this will need to be review to handle that properly.

libraries/Wire/src/Wire.cpp
```diff
@@ -380,6 +380,13 @@ void TwoWire::onReceiveService(uint8_t* inBytes, int numBytes)
   if(rxBufferIndex < rxBufferLength){
     return;
   }
+
+  allocateRxBuffer(numBytes);
```
lacklustrlabs: Wouldn't it be better to do as little as possible inside the ISR? - I.e do the RX buffer allocation in begin() when slave mode is detected.

Since the current implementation of onReceiveService() never appends data to the RX buffer (it always start writing at index 0) it should be safe to allocate the buffer in begin(). 
At least as long as I2C_TXRX_BUFFER_SIZE <= BUFFER_LENGTH (add a static_assert for safety)

fpistm: if numbytes is greater than the one allocated it need realloc

lacklustrlabs: yes, but it will never be greater.  I2C_TXRX_BUFFER_SIZE = BUFFER_LENGTH = 32  (at least for now)

lacklustrlabs: allocateRxBuffer(numBytes) can still be called in the ISR, just to make sure. 
In the normal case it will do nothing if the buffer is pre-allocated in begin()

fpistm: you're right in case of slave receive. I will check all those things when I will review #217

PR #224, opened 2018-03-08
Specify the ARM GCC version in platforms.txt

sandeepmistry:
I had a conflict with the version of the ARM GCC used by the official Arduino SAMD core, this changes specifies the version of GCC in the path, so that the arduino-builder/IDE can select the right version.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -19,7 +19,7 @@ compiler.warning_flags.default=
 compiler.warning_flags.more=-Wall
 compiler.warning_flags.all=-Wall -Wextra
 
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @sandeepmistry,
Thanks for this PR.
I didn't know it is possible to specifies the version of GCC in the path.

fpistm:
I've made some tests.
I reproduce the issue only when I use the STM32 core git repo.
With SAMD and STM32 cores installed thanks the board managers it is ok.
To avoid any issue, I think it's better to merge this PR.
Thanks @sandeepmistry

PR #231, opened 2018-03-20
Introduce new ADC API for DMA-based high sampling rate

fprwi6labs:
Continuation of the PR #175 
I had to open a new PR because @damienWi6labs is no longer on this project.

I followed @fpistm comments (see  #175).
DMA layer is now fully independent of the peripheral.
DMA has been tested only with ADC peripheral.

Wiki page about this method available [here](https://github.com/stm32duino/wiki/wiki/API).
Example provided [here](https://github.com/stm32duino/STM32Examples/pull/3).

### DIFF
--- a/cores/arduino/stm32/analog.c
+++ b/cores/arduino/stm32/analog.c
@@ -50,6 +50,7 @@
 #include "analog.h"
 #include "timer.h"
 #include "PinAF_STM32F1.h"
+#include "dma.h"
 
 #ifdef __cplusplus
  extern "C" {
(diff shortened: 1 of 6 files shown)


fpistm:
Hi @fprwi6labs,
thanks for this PR. One question: have you take in account my remarks on wiring_analog files on #175?

fprwi6labs:
Hi @fpistm 
This PR is fully based on your remarks. Did I miss something for wiring_analog files?

fpistm:
Currently not it seems ok. Thanks

ropa35:
How to use?

fpistm:
@ropa35, 
this is a PR, if you want to use it, you will have to merge it on the core source, best way is to use git clone of the core.

cores/arduino/stm32/analog.c
```diff
@@ -638,32 +790,161 @@ uint16_t adc_read_value(PinName pin)
...
+  ADCxContext->current_pin  = pin;                                             /* Needed for HAL_ADC_MspInit     */
+  ADCxContext->cb           = callback;                                        /* Register the callback function */
+  ADCxContext->user_data    = functionParameter;
+  AdcHandle                 = &(ADCxContext->ADCHandle);
+  AdcHandle->Instance       = peripheral;
+
+  if (HAL_ADC_DeInit(AdcHandle) != HAL_OK)
+  {
+    return 0;
+  }
```
tve: I find it pretty bothersome that for each call to `adc_read_value_dma` the ADC gets deinitialized, then initialized, finally sampled, and then deinitialized again. Yes, this is amortized over `lData` samples, but still... How lkong does all this de-init/init/de-init take? Why not create an API that lets the user init things, and then call a very lightweight function to collect a batch of samples?

tve:
IMHO this suffers from the same problems that the current analogRead implementation suffers, which is that the API only provides a single call which has to init everything from scratch, collect the samples, and then deinit everything.

(What benefit does using the HAL give over using LL?)

fpistm:
Yes you're right that's why this PR is not merged.
HAL is higher level API than LL. HAL function do several check and operations per function while in LL you have to code all by yourself (mainly direct access to register).
The main constraint to HAL is the code size.
Legacy, LL was not at the right level. Now they are fully deployed and that's why I've planned to move/use  LL when I get time.

tve:
I started an STM32ADC library here: https://github.com/tve/goobies/blob/master/libraries/STM32ADC/src/STM32ADC.h
It is adapted from the STM32ADC library in RogerClark's core.
I would appreciate if you could give it a look and let me know if you think this is more along the lines of what you are thinking. I'm hoping the implementation can be "just a bunch of LL_ calls".
I tried to define it such that simple usage requires very few calls. I don't have experience with DMA & ADC so those functions are pretty straight from RogerClark's core and I'll have to see when I get there...
My main constraint is that I pretty much only have STM32L0 boards to test with. I think I have an F401 nucleo somewhere I can perhaps dig out...

tve:
I can't do a PR until I have something that shows sign of life :-). I posted the link mainly to get feedback on the interface before I start coding too much and get attached to an implementation...

cores/arduino/stm32/analog.h
```diff
@@ -56,6 +79,8 @@ void dac_stop(PinName pin);
 uint16_t adc_read_value(PinName pin);
 void pwm_start(PinName pin, uint32_t clock_freq, uint32_t period, uint32_t value, uint8_t do_init);
 void pwm_stop(PinName pin);
+bool adc_read_value_dma(PinName pin, uint32_t *pData, uint32_t lData,
+                            void (*callback)(void *user_data), void *functionParameter);
```
ghent360: you can probably rename the last argument to user_data safely.
Can you allso add comment what happens if there is an error: does the CB get called? Do we need on_error callback?

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,493 @@
...
+  * @brief  Find DMA instance from DMA mapping information
+  * @param  DMAx    : DMA number (1 or 2)
+  * @param  channel : channel number (0 to 7)
+  * @retval DMA
+  */
+static void *DMA_find_instance(int8_t DMAx, int8_t channelx)
+{
+  void *DMAx_Instance = NULL;
+
+  if((DMAx == NC) || (channelx == NC)) {
```
ghent360: suggestion add: || DMAx < 1 || DMAx > 2 || channelx < 0 || channelx > 7

cores/arduino/stm32/dma.c
```diff
@@ -0,0 +1,493 @@
...
+
+  // Match with DMAHandle index
+  DMAx -= 1;
+
+  // Configure DMA
+  DMAHandle[DMAx][channel].Instance                 = DMAx_Instance;
+  DMAHandle[DMAx][channel].Init.Direction           = DMA_PERIPH_TO_MEMORY;
+  DMAHandle[DMAx][channel].Init.PeriphInc           = DMA_PINC_DISABLE;
+  DMAHandle[DMAx][channel].Init.MemInc              = DMA_MINC_ENABLE;
+  DMAHandle[DMAx][channel].Init.PeriphDataAlignment = DMA_PDATAALIGN_WORD;//DMA_PDATAALIGN_HALFWORD;
```
ghent360: isn't ADC halfword?

fpistm:
Close this one as it will never merged as it is.
This should be covered by #5.

PR #234, opened 2018-04-02
Enable IWDG module for Nucleo-64 F446RE board.

ghent360:
Not sure why it was not enabled in the first place. It would probably work fine for other boards
as well.


fpistm:
Hi, @ghent360 
There is no specific reason. 
At the beginning we targetted to be Arduino Uno "like", so only basic feature was enabled, I guess. 
User can enable HAL module, modifying the hal conf, or more simply adding a build_opt.h file at sketch level with `-DHAL_IWDG_MODULE_ENABLED`
Anyway, It could be enabled by default.

ghent360:
Hi, @fpistm 
Thank you for the consideration. Yes the build_opt.h file does work. 
How does this look for an API start:

https://github.com/ghent360/Marlin/blob/bugfix-2.0.x/Marlin/src/HAL/HAL_STM32F4/watchdog_STM32F4.h

Or would you rather have Arduino style C++ API?

Watchdog.begin(uint16_t reload_time /* seconds */); // starts the watchdog
Watchdog.reload(); // call periodically to reload and prevent reset

fpistm:
welcome. It's an community project. Any contribution are welcome. ;)
About API, it's open. C or C++ are welcome. I think the goal is to have something generic and user friendly.
Probably, C++ is not required to avoid use an instance or maybe both could be proposed...
C++ one using the C API? 
Do not hesitate to propose one, why not based on Marlin one.

ghent360:
I reverted the HAL config change since the build_opt.h file can serve the same purpose. I added a Watchdog library and an example to expose the IWDG component.

ghent360:
@fpistm any thoughts on the API proposal?

fpistm:
Hi @ghent360,
first look seems fine.
I will deeper review it in the coming weeks. Currently, I'm reviewing/testing RTC and Low Power  which is a huge task ;)

fpistm:
I need to do more tests before release it so I will not have time to validate it properly for 1.3.0. This will be part of the 1.3.1

fpistm:
I've reviewed this PR and done tests across several STM32 Series.
Basically, it works for series which do not have Window feature on those having this not worked.
Library format is not fully respected: src/ folder, library version have to be semver x.y.z.
I will rename it IWatchdog to differentiate with WWDG (when implemented).
New API could be added to know if the system has resumed from IWDG reset.
And probably handle also reset < 1ms (handle prescaler value 4/8/16) even if LSI accuracy is not very good. (I remember it is possible to use a timer connected to LSI to measure LSI freq but this could/will be an other enhancement, primary goal of this PR is to provide IWDG access :))
Also use LSI_VALUE typical value to define the best combo prescaler/reload value (example for F1, LSI is 40KHz).
Maybe using C++ could be fine to handle default method parameters and also using LL.
I need to think about all of this to handle it in the best way. 
Any suggestions are welcome.

fpistm:
#292 merged. This one can be closed.
Thanks @ghent360

PR #238, opened 2018-04-27
Add Blackmagic (BMP) support for Generic F103

evark:
A small addition to boards/platform files to support Blackmagic probe as uploading method for Generic F103 devices.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -172,3 +172,11 @@ tools.serial_upload.path.linux64={runtime.hardware.path}/tools/linux64
 tools.serial_upload.upload.params.verbose=-d
 tools.serial_upload.upload.params.quiet=n
 tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+
+# blackmagic upload for generic STM32
+tools.bmp_upload.cmd=arm-none-eabi-gdb
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc.path}/bin/
+tools.bmp_upload.upload.speed=230400
+tools.bmp_upload.upload.params.verbose=
+tools.bmp_upload.upload.params.quiet=-q --batch-silent
+tools.bmp_upload.upload.pattern="{path}{cmd}" -cd "{build.path}" -b {upload.speed} {upload.verbose} -ex "set debug remote 0" -ex "set target-async off" -ex "set remotetimeout 60" -ex "set mem inaccessible-by-default off" -ex "set confirm off" -ex "set height 0" -ex "target extended-remote {serial.port}" -ex "monitor swdp_scan" -ex "attach 1" -ex "x/wx 0x8000004" -ex "monitor erase_mass" -ex "echo 0x8000004 expect 0xffffffff after erase\n" -ex "x/wx 0x8000004" -ex "file {build.project_name}.elf" -ex "load" -ex "x/wx 0x08000004" -ex "tbreak main" -ex "run" -ex "echo \n\n\nUpload finished!" -ex "quit"
--- a/platform.txt
+++ b/platform.txt
@@ -175,8 +175,8 @@ tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.alt
 
 # blackmagic upload for generic STM32
 tools.bmp_upload.cmd=arm-none-eabi-gdb
-tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc.path}/bin/
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/
 tools.bmp_upload.upload.speed=230400
-tools.bmp_upload.upload.params.verbose=
-tools.bmp_upload.upload.params.quiet=-q --batch-silent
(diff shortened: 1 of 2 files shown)


platform.txt
```diff
@@ -172,3 +172,11 @@ tools.serial_upload.path.linux64={runtime.hardware.path}/tools/linux64
 tools.serial_upload.upload.params.verbose=-d
 tools.serial_upload.upload.params.quiet=n
 tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+
+# blackmagic upload for generic STM32
+tools.bmp_upload.cmd=arm-none-eabi-gdb
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc.path}/bin/
```
fpistm: I think you should use
`tools.bmp_upload.path={compiler.path}`

which is set on top of platform.txt
`compiler.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/`
This will avoid any issue if compiler version change

sabas1080:
I can try in a few days with my BMP

RickKimball:
It looks like you are using the original version of the BMP command line I wrote a long time ago.  I've since simplified the parameters so that it isn't doing a lot of extra work.  You might look at doing something like this:

```
# blackmagic probe upload
tools.bmp_upload.cmd=arm-none-eabi-gdb
tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc.path}/bin/
tools.bmp_upload.upload.speed=115200
tools.bmp_upload.upload.params.verbose=-batch
tools.bmp_upload.upload.params.quiet=--batch-silent
tools.bmp_upload.upload.pattern="{path}{cmd}" -cd={build.path} -ex "file {build.project_name}.elf" -nh -b {upload.speed} {upload.verbose} -ex "target extended-remote {serial.port}" -ex "monitor swdp_scan" -ex "attach 1" -ex "stop" -ex "mon erase_mass" -ex "load"
```
I started a post a while back http://www.stm32duino.com/viewtopic.php?f=37&t=2802 but I don't think it went anywhere.

Seeing as you are adding something new to this core, you might think about how to select the debug BMP port a different way so you can still use the serial console without having to switch the port between the debug port and the virtual com port.  I had some ideas in that post thread.

evark:
I tried the path
`tools.bmp_upload.path={compiler.path}`

but it seems that this is not evaluated correctly (why?), so I have to define the compiler path again 

`tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/`

The command line of gdb can be simplified, according [blackmagic wiki](https://github.com/blacksphere/blackmagic/wiki/GDB-Automation) and looking how the gdb is used in [platformio](https://github.com/platformio/platform-ststm32/blob/develop/builder/main.py#L157) as 

`tools.bmp_upload.upload.pattern="{path}{cmd}" -nx --batch -ex "set confirm off" -ex "target extended-remote {serial.port}"  -ex "monitor swdp_scan"  -ex "attach 1" -ex "load" -ex "compare-sections" -ex "kill" "{build.path}/{build.project_name}.elf"`

There is no need to full erase the chip on every download.

platform.txt
```diff
@@ -172,3 +172,11 @@ tools.serial_upload.path.linux64={runtime.hardware.path}/tools/linux64
...
 tools.serial_upload.upload.params.quiet=n
 tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+
+# blackmagic upload for generic STM32
+tools.bmp_upload.cmd=arm-none-eabi-gdb
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/
+tools.bmp_upload.upload.speed=230400
+tools.bmp_upload.upload.params.verbose=-batch
+tools.bmp_upload.upload.params.quiet=--batch-silent
+tools.bmp_upload.upload.pattern="{path}{cmd}" -nx {upload.verbose} -ex "set confirm off" -ex "target extended-remote {serial.port}"  -ex "monitor swdp_scan"  -ex "attach 1" -ex "load" -ex "compare-sections" -ex "kill" "{build.path}/{build.project_name}.elf"
```
fpistm: There are some extra space btw:
`{serial.port}"  -ex "monitor swdp_scan`
and
`-ex "monitor swdp_scan"  -ex "attach 1"`

platform.txt
```diff
@@ -172,3 +172,11 @@ tools.serial_upload.path.linux64={runtime.hardware.path}/tools/linux64
 tools.serial_upload.upload.params.verbose=-d
 tools.serial_upload.upload.params.quiet=n
 tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+
+# blackmagic upload for generic STM32
+tools.bmp_upload.cmd=arm-none-eabi-gdb
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/
+tools.bmp_upload.upload.speed=230400
```
fpistm: upload speed is not used in the cmd? It should be fine to add it, no?
` -b {upload.speed} `

fpistm:
I've tested under windows, it seems it could not expand properly `{runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}` at upload step.... I need to investigate this. You use it under Linux?

platform.txt
RickKimball: It is really a USB CDCACM so it doesn't really matter. It is going to go USB speed. But you can add it without issue

fpistm:
I've open an issue at Arduino side:
https://github.com/arduino/Arduino/issues/7616

platform.txt
```diff
@@ -172,3 +172,11 @@ tools.serial_upload.path.linux64={runtime.hardware.path}/tools/linux64
 tools.serial_upload.upload.params.verbose=-d
 tools.serial_upload.upload.params.quiet=n
 tools.serial_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+
+# blackmagic upload for generic STM32
+tools.bmp_upload.cmd=arm-none-eabi-gdb
+tools.bmp_upload.path={runtime.tools.arm-none-eabi-gcc-6-2017-q2-update.path}/bin/
```
fpistm: Based on @facchinm comment  in https://github.com/arduino/Arduino/issues/7616#issuecomment-390894992
`{runtime.tools.arm-none-eabi-gcc}` will be used by default in git repo.
I will update path when releasing the package to set the gcc version as it is done for other upload tools.

fpistm:
Thanks @evark 
It is merged with small fix.

Issue #240, opened 2018-05-04
va_start & va_end are not defined


Adminius:
i'm missing this two functions:

- `va_start`
- `va_end`

i had success with this workaround:

> `if defined(ARDUINO_ARCH_STM32)
> #include <stdarg.h>
> #define va_start(v,l)	__builtin_va_start(v,l)
> #define va_end(v)	__builtin_va_end(v)
> #endif

in which file i can make PR? :)

fpistm:
I think you could simply include stdio.h here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/bad5b56870ce311a4d591f789cbe5d47450097e4/cores/arduino/wiring.h#L23

Adminius:
if add my implementaion in wiring_privat.h it doesn't work. 
But it works in wiring.h...

P.S. im on vacation in paris until weekend, so i'm not available...

fpistm:
Hi @Adminius ,
could you share with me your use case to understand why it's not working with wiring_private.h

Adminius:
we use in our [KONNEKTING Library](https://github.com/KONNEKTING/KonnektingDeviceLibrary/tree/develop_beta4a) a Debug Class, that uses [this to functions](https://github.com/KONNEKTING/KonnektingDeviceLibrary/blob/develop_beta4a/src/DebugUtil.cpp#L59).

You can complie a [DemoSketch ](https://github.com/KONNEKTING/KonnektingDeviceLibrary/blob/develop_beta4a/examples/DemoSketch/DemoSketch.ino)with [KDEBUG enabled](https://github.com/KONNEKTING/KonnektingDeviceLibrary/blob/develop_beta4a/examples/DemoSketch/DemoSketch.ino#L9) to test it

fpistm:
@Adminius 
I've tested and build is now ok.
Let me know if it is ok then I will merge #253

Adminius:
Hi @fpistm 
it works, thx. 
I will accept your PR after new STM32 core will be released, deal?

Issue can be closed.

fpistm:
Of course ;)
New release is planned this week or the next one.

PR #241, opened 2018-05-05
va_start & va_end added

Adminius:
solves issue #240

### DIFF
--- a/cores/arduino/wiring.h
+++ b/cores/arduino/wiring.h
@@ -21,6 +21,7 @@
 #define _WIRING_H_
 
 #include <stdint.h>
+#include <stdarg.h>
 #include <stdlib.h>
 #include <stdbool.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


cores/arduino/wiring.h
```diff
@@ -52,4 +53,7 @@
 #define clockCyclesToMicroseconds(a) ( ((a) * 1000L) / (SystemCoreClock / 1000L) )
 #define microsecondsToClockCycles(a) ( (a) * (SystemCoreClock / 1000000L) )
 
+#define va_start(v,l)	__builtin_va_start(v,l)
```
fpistm: Those lines are not needed if you include also stdio.h.
Anyway, I will investigate to include them using wiring_private.h to be aligned with Arduino as mentioned in #240.

Adminius:
we have to look for better way...

PR #262, opened 2018-06-10
Add define for ARDUINO_CORE_STM32 that can be used to identify this Arduino core

ktand:
It would be nice if it was possible to identify this Arduino core. I have added a define for ARDUINO_CORE_STM32.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -29,7 +29,7 @@ compiler.c.elf.cmd=arm-none-eabi-gcc
 compiler.objcopy.cmd=arm-none-eabi-objcopy
 compiler.elf2hex.cmd=arm-none-eabi-objcopy
 
-compiler.extra_flags=-mcpu={build.mcu} -mthumb "@{build.opt.path}"
+compiler.extra_flags=-mcpu={build.mcu} -mthumb "@{build.opt.path}" -DARDUINO_CORE_STM32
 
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @ktand ,
in fact it already exist. Each arduino core define its own define: `-DARDUINO_ARCH_{build.arch}`
Use `ARDUINO_ARCH_STM32`

ktand:
Hi @fpistm.

Unfortunately the [Platformio framework](https://github.com/maichaell/platform-ststm32) that has been modified for [STM32GENERIC](https://github.com/danieleff/STM32GENERIC) defines the same `ARDUINO_ARCH_STM32`. I've created a HAL for STM32F4 for the [Marlin](https://github.com/MarlinFirmware/Marlin) project using STM32GENERIC but would like to be able use this core as well. Any idea how I could identify this framework without relying on the `ARDUINO_ARCH_STM32` define?

fpistm:
Unfortunately, I would not add new define, official one is `ARDUINO_ARCH_STM32`
This define is used in several third party libraries for this core and do not want multiply define to identify this core.

ktand:
@fpistm I realized that STM32GENERIC defines `STM32GENERIC `so I can use that define to switch on. I will close this PR.

PR #292, opened 2018-08-03
Add built-in IWatchdog library to expose the IWDG component.

fpistm:
Based on first implementation (#234) from @ghent360.

Use LL instead of HAL (no need to enable the HAL_IWDG_MODULE_ENABLED).
Applicable to all STM32 series.
Window mode is available for STM32 series which support the feature.
API  added to know if the system has resumed from IWDG reset.
Timeout range increase using LSI_VALUE of the target.

### DIFF
--- a/libraries/IWatchdog/README.md
+++ b/libraries/IWatchdog/README.md
@@ -1,0 +1,136 @@
+## IWatchdog Library V1.0.0 for stm32 Arduino core
+
+**Written by:** _Venelin Efremov_ and _Frederic Pillon_
+
+### Requirement
+* [Arduino_Core_STM32](https://github.com/stm32duino/Arduino_Core_STM32) version > 1.3.0
+
+### What is the IWatchdog library.
+
+Th IWatchdog library provides an interface to the independent watchdog module (IWDG) inside STM32 chips.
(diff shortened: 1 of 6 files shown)


hasenbanck:
Works as advertised. Tested with the custom F765VI board I have developed. Only tested the provided example project and tested including the library in my current Marlin branch.

libraries/IWatchdog/README.md
```diff
@@ -0,0 +1,136 @@
...
+**Written by:** _Venelin Efremov_ and _Frederic Pillon_
+
+### Requirement
+* [Arduino_Core_STM32](https://github.com/stm32duino/Arduino_Core_STM32) version > 1.3.0
+
+### What is the IWatchdog library.
+
+Th IWatchdog library provides an interface to the independent watchdog module (IWDG) inside STM32 chips.
+The IWDG module is used in production systems to generate a reset signal to the CPU in case some
+catastrophic event in causes the software to become "stuck" or unresponsive.
```
ghent360: s/event in causes/event causes

fpistm: done

PR #293, opened 2018-08-05
Added STM32F030K6 as Generic type

swissbyte:
Hello 

I have tried to add the STM32F030K6 as a generic part. 
This config compiles successfully. I will add more Parts if this is appreciated. 

Regards
Swissbyte


swissbyte:
I saw that the checks have failed. What is the idea behind the onboard led? A generic Chips does not have on board LEDs. Since there is no board at all.

fpistm:
Hi, thanks,
I will review it when got some time.
Any contributions are welcome.

You're right, Blink sketch it's probably not the best sketch to build for Travis.
Travis deployment was done since 2 Weeks, this will be enhance.

swissbyte:
Thanks for your response. If you have some guidlines for further contributions, please let me know so that i can follow them in the future. By the way, i tried to open an account on the stm32duino forum. Unfortunately i never received an activation email.

boards.txt
```diff
@@ -712,6 +753,34 @@ GenF103.menu.opt.ogstd=Debug (-g)
 GenF103.menu.opt.ogstd.build.flags.optimize=-g -Og
 GenF103.menu.opt.ogstd.build.flags.ldspecs=
 
+GenF030.menu.opt.osstd=Smallest (-Os default)
+GenF030.menu.opt.osstd.build.flags.optimize=-Os
```
fpistm: This line could be removed as it is the default in platform.txt

boards.txt
```diff
@@ -712,6 +753,34 @@ GenF103.menu.opt.ogstd=Debug (-g)
...
+GenF030.menu.opt.osstd.build.flags.ldspecs=
```
fpistm: This line could be removed as it is the default in platform.txt

boards.txt
```diff
@@ -439,6 +439,41 @@ Disco.menu.upload_method.STLink.upload.tool=stlink_upload
...
+GenF030.pid.0=0x0004
+
+GenF030.build.core=arduino
+GenF030.build.board=GenF030
+GenF030.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial}
+
+# Generic F030 board
+# Support: Serial1 (USART1 on PA10, PA9)
+GenF030.menu.pnum.GENERIC_F030K6=Generic F030K6
+GenF030.menu.pnum.GENERIC_F030K6.node="No_mass_storage_for_this_board_Use_STLink_upload_method"
```
fpistm: Not needed if you do not defined massstorage upload method

boards.txt
```diff
@@ -439,6 +439,41 @@ Disco.menu.upload_method.STLink.upload.tool=stlink_upload
...
+GenF030.vid.0=0x1EAF
+GenF030.pid.0=0x0004
+
+GenF030.build.core=arduino
+GenF030.build.board=GenF030
+GenF030.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial}
+
+# Generic F030 board
+# Support: Serial1 (USART1 on PA10, PA9)
+GenF030.menu.pnum.GENERIC_F030K6=Generic F030K6
```
fpistm: If the idea is to have only one entry and then select the proper RAM/flash size and appropriate linker script, then it should be rewrote:

Something like that:
GenF030.menu.pnum.GENERIC_F030xx=Generic F030xx

and adding a new menu entry:
menu.var=Variant

Then: 
GenF030.menu.var.K6=K6
GenF030.menu.var.K6.upload.maximum_size=32768
GenF030.menu.var.K6.upload.upload.maximum_data_size=4096
GenF030.menu.var.K6.build.board=GENERIC_F030K6
GenF030.menu.var.K6.build.ldscript=ldscriptK6.ld
GenF030.menu.var.K6.build.product_line=STM32F030x6

So new var could be added like that:
GenF030.menu.var.XY=XY
GenF030.menu.var.XY.upload.maximum_size=xxxxx
GenF030.menu.var.XY.upload.upload.maximum_data_size=xxxx
GenF030.menu.var.XY.build.board=GENERIC_F030XY
GenF030.menu.var.XY.build.ldscript=ldscriptXY.ld
GenF030.menu.var.XY.build.product_line=STM32F030xY

I did not test but the philosophy is here, I guess

variants/GENERIC_F030K6/variant.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+#include "variant.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Pin number
+const PinName digitalPin[] = {
+ //P1 connector Right side (bottom view)
```
fpistm: I guess this one is a copy of the Disco F030.
So not aligned for F030K6.
Here simply list all pin in order:
PA_0,
PA_1,....

List all analog at  end of the enum to avoid duplicated entries

variants/GENERIC_F030K6/variant.h
```diff
@@ -0,0 +1,192 @@
...
+#endif // __cplusplus
+
+/*----------------------------------------------------------------------------
+ *        Pins
+ *----------------------------------------------------------------------------*/
+extern const PinName digitalPin[];
+
+enum {
+  //P1 connector Right side (bottom view)
+  PC13, //D0
```
fpistm: Same comment than before

swissbyte:
Thank you for your review. 
I will do the necessarry changes as soon as possible!

boards.txt
swissbyte: I have tried to understand your idea. It would be great if we could choose STM32F030xx as Chip Series and then slecet at the "Board part Number" the appropriate chip. 
Unfortunately i was not able to get this working... 

i see that you want to replace the menu.var with the variant. But how can you choose the variant variable?

swissbyte:
I have changed all files according to your review requests. 
Very stupid that i forgot to modifie the pins in the variant.cpp and h file. 
Now everything should be correct. 

I only havent figured out how to implement the board.txt in such a way that we can shoose the specific chip in the "board part number" dropdown instead of the "board" drop down.

boards.txt
fpistm: Arduino menu is not very convenient.
The menu var is only a new submenu where you can select the chip version (K6, K8, R8,...)

swissbyte: Ok i think i see your goal. 
I will try to implement it. 
Thanks

Aulakiria:
I made a breakout board.




But, ST-Link v2 does not connect k6 chips.
It is not connect at all in Arduino.
It must be tried more than 10 times in the ST-Link Utility and it will be connect only once.
- Both the Aduno and Utility connect F103 board well.

My Aduino 1.8.7 has a problem.
The first compilation works well.
The second compilation fails.

Version 1.6.5 has no problem compiling, but it shows a lot of error messages. ToT

Aulakiria:
k6-sch
I made a schematic as shown in the picture.
Is there anything I have done wrong?

Aulakiria:
I make Break board.
It is Good working. ^^

Aulakiria:
oled test... ^^;

Aulakiria:
I wrote a sketch that displays the input of PA0.
It does not display the input of PA0.
PA5 input is displayed.

#include <Arduino.h>
#include <U8g2lib.h>

const int AnalogPin = PA0;

void setup(void) {
  pinMode(AnalogPin, INPUT);
  u8g2.begin();
}

void loop(void) {
  u8g2.firstPage();
  do {
    u8g2.setFont(u8g2_font_ncenB14_tr);
    u8g2.setCursor(0, 15);
    u8g2.print(analogRead(AnalogPin));
    //u8g2.drawStr(0,15,analogRead(0));
  } while ( u8g2.nextPage() );
}

fpistm:
Since this PR was submit, several update of the core has been added which would allow to be more generic.
(LD script preprocessing, common attribute...)
Moreover, after thinking about it and based on users feedback, variant should be only for a dedicated or family board else this bring too much support around customization, configuration,...
It is enough simple to create his own variant with all file provided (PeripheralPins, PinMaps, ldscript and clock config generated by STM32CubeMX). 
One thing I will do is it to provide a way to generate the variant.[h|cpp] with a default pin mapping.

Issue #298, opened 2018-08-09
I2C hardware or Software not working on STM32F103C8T6 with "generic STM32F103 series" selected


nikisalli:
using a normal I2C scanner code, i keep getting random addresses detected. i tried with another core (roger's core) with the same hardware and code and it works but when i switch to the stm32duino core it doesn't work anymore. i tried both hardware i2c and software i2c with the same results.

code:
`#include <Wire.h>

void setup() {
  Serial1.begin(115200);
  Wire.begin();
  Serial1.println("\nI2C Scanner");
}

void loop() {
  byte error, address;
  int nDevices;

  Serial1.println("Scanning...");

  nDevices = 0;
  for(address = 1; address < 127; address++) {
    // The i2c_scanner uses the return value of
    // the Write.endTransmisstion to see if
    // a device did acknowledge to the address.

    Wire.beginTransmission(address);
    error = Wire.endTransmission();
    
    if (error == 0) {
      Serial1.print("I2C device found at address 0x");
      if (address < 16) 
        Serial1.print("0");
      Serial1.println(address, HEX);

      nDevices++;
    }
    else if (error == 4) {
      Serial1.print("Unknown error at address 0x");
      if (address < 16) 
        Serial1.print("0");
      Serial1.println(address, HEX);
    }    
  }
  if (nDevices == 0)
    Serial1.println("No I2C devices found");
  else
    Serial1.println("done");

  delay(5000);           // wait 5 seconds for next scan
}`

fpistm:
Linked #217

nikisalli:
Isn't there a temporary fix i can use? Because from what i see it's like six months it is broken

LMESTM:
@nikisalli Hi - please find a proposed patch to fix the scan issue - let me know if you can test on your side.

nikisalli:
@LMESTM i just tested your patch on the bluepill (stm32f103c8t6) and it works flawlessly

LMESTM:
@nikisalli thanks for the quick feedback and testing !

PR #301, opened 2018-08-10
Handle I2C scan with specific call to HAL_I2C_IsDeviceReady

LMESTM:
# Description 

When size is 0, the objective is to check if device is there and ready.
A typical use case is to scan through all addresses to check for
available slaves.

So we're checking the various error cases to report to application.

# Tests

I've run successfully the example sketch from #298 on:
NUCLEO_F429ZI
NUCLEO_F411RE
DISCO L475 IOT


nikisalli:
works on the bluepill board (stm32f103c8t6) i tested the i2c scanner and 5 different sensors on both i2c1 and i2c2, works flawlessly

nikisalli:
works on stm32f103vet6 too (i made a custom board in the selection menu)

Testato:
Please add to this PR also the scanner sketch  that you are used (under Wire library examples).

Thanks a lot for your work

LMESTM:
@nikisalli would you mind pushing your proposed scanner sketch ?

Testato:
PR with the scanner , tested an all three I2c for a Bluepill
https://github.com/stm32duino/Arduino_Core_STM32/pull/303

MauroMombelli:
hi @LMESTM , thanks for the patch. I just tested, using an arduino UNO as slave (standard test sketch), and on a discovery stm32f407
Result: if while scanning i connect the arduino, it does NOT get seen. If i reboot the STM will see the arduino without problem. At least during one disconnection  the i2c hanged.

LMESTM:
> PR with the scanner , tested an all three I2c for a Bluepill
#303

Thanks @Testato

LMESTM:
> hi @LMESTM , thanks for the patch. I just tested, using an arduino UNO as slave (standard test sketch), and on a discovery stm32f407
Result: if while scanning i connect the arduino, it does NOT get seen. If i reboot the STM will see the arduino without problem. At least during one disconnection the i2c hanged.

@MauroMombelli  Have you added pull up resistors on the I2C lines ?

fpistm:
This PR Will Fix  #298

MauroMombelli:
@LMESTM for sure the arduino is using its internal by default, not sure about this core. I can add some 4.7kohm resistor instead if you think this is the issue, but i don't see why it should work when i reboot the stm; technically any new scan should be completely independent from previous status of the bus, and leaving the system connected for a couple of scan should see the slave without need to reboot.
My idea is once in "error 1" it never really reset completely the communication and will always return error.

LMESTM:
@MauroMombelli without pull-up resistors, plugging / unplugging the boards will lead to  falling edges of the I2C lines which could lead to undesired behavior or possible lock-up of the bus.

MauroMombelli:
>possible lock-up of the bus

aren't the timeout there to avoid this issue?

LMESTM:
I meant this probably leads to HW lock up ... so the need to reset the STM32.
You may try out using internal pull-up on I2C lines of STM32 as well ... even though this is not the recommended configuration (recommendation is external pull-up resistors)

cores/arduino/stm32/twi.c
```diff
@@ -355,6 +355,11 @@ i2c_status_e i2c_master_write(i2c_t *obj, uint8_t dev_address,
   uint32_t tickstart = HAL_GetTick();
   uint32_t delta = 0;
 
+  /* When size is 0, this is usually a I2C scan / ping to check if device is there and ready */
```
fpistm: > an I2C scan

LMESTM: ok done

Issue #296, opened 2018-08-09
Buffered read / writes for the EEPROM emulation


hasenbanck:
Right now the EEPROM emulation inside the core (stm32_eeprom.c/h) provides two functions for reading and writing to an EEPROM emulated via flash (size of a page).

Both functions are touching the whole page in each function call, which is especially wasteful when writing to the flash.

To stay backward-compatible, I would propose to add four new functions to the core library:

```C
void eeprom_buffer_fill();
void eeprom_buffer_flush();
uint8_t eeprom_buffered_read_byte(const uint16_t __p);
void eeprom_buffered_write_byte(uint16_t __p, uint8_t __value);
```

The developer would then to first fill the static, private buffer with the fill function, interact with it and when he wants to save the content of the buffer to flash, flush the buffer to flash.

Implementing these changes shouldn't take much effort, since all pieces are already in place. I would like to do the implementation.

Comments? Suggestions?

fpistm:
The EEPROM is not really optimized. 
This feature is mainly to provide easiest way to save/restore some user settings which don't required high data rate and access else a real storage (SD,...) should be used.
Anyway, your suggestions seems a good improvement.

hasenbanck:
Since I'm currently adapting Marlin for your core, this problems popped up. Currently Marlin would safe the configuration (around 600 bytes) one byte at a time and would use 600 erase cycles of the whole page just for one save operation.

I implemented the usage of the Backup SRAM as an alternative, but I want to correct the current default operation mode non the less.

I will open a PR once I finished and tested the implementation.

PR #302, opened 2018-08-10
Rework eeprom emulation for buffered access

hasenbanck:
The current eeprom emulation didn't had any buffered access to the
data on the flash. Every access to a byte would trigger a whole
page read / write. For writes it would mean, that the whole page
would be erased. For backward compatibility we will keep
the current functions as they are, but will add functions
to fill/flush the buffer and read from it.

The buffer size is static and is one page in size.

This commit also cleans up old unused code comments and
documentation stubs. I also cleaned up the old references
to the original author / version / date, since we already
refactored this code quite well. The copyright notice of
course is still intact.

@fpistm If I reworked too much from the file, then I can of course pull back the RC and just change the affected parts, but I thought it would be best if I clean up the code while working on it anyhow.

Fixes: #296

### DIFF
--- a/cores/arduino/stm32/stm32_eeprom.c
+++ b/cores/arduino/stm32/stm32_eeprom.c
@@ -1,11 +1,7 @@
 /**
   ******************************************************************************
-  * @file    eeprom.c
-  * @author  WI6LABS
-  * @version V1.0.0
-  * @date    01-August-2016
-  * @brief   provide emulated eeprom from flash
-  *
+  * @file    stm32_eeprom.c
+  * @brief   Provides emulated eeprom from flash
   ******************************************************************************
   * @attention
   *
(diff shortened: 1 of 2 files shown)


fpistm:
Thanks @hasenbanck 
I will try to review it before my vacation.

hasenbanck:
@fpistm Are you back from your vacation? Could my work be merged sooner or later?

fpistm:
Hi @hasenbanck 
Yes. I'm back.
I've started to review it .
My first feeling is that  I preferred to split in 2 commits one for the clean and the second for effective implementation.

fpistm:
@hasenbanck 
I've splitted your commits:
https://github.com/fpistm/Arduino_Core_STM32/tree/pr-302-review

I've reviewed the code and it looks fine.
I will made some tests next week.
Thanks

fpistm:
Maybe an example using the new api could be added.
Maybe updating EEPROM class to support this buffering could be added.
Example by adding private member to set buffering access on and add a way to flush

hasenbanck:
Were would you like to have an example for the C API? Maybe in the STM32Example repository?

We could of course rework the EEPROM class to support this buffering, but for now I would separate it from this pull request (or your branch of it).

fpistm:
Yes STM32Example is a good candidate.
OK for EEPROM Class, this is just an idea ;)

fpistm:
#333 merged

PR #303, opened 2018-08-10
I2C Scanner examples

Testato:
Tested on Bluepill for all three I2c (TWI 1 - TWI 1 alternative pin - TWI 2)

### DIFF
--- a/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
+++ b/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
@@ -1,0 +1,89 @@
+// --------------------------------------
+// i2c_scanner
+//
+// Version 1
+//    This program (or code that looks like it)
+//    can be found in many places.
+//    For example on the Arduino.cc forum.
+//    The original author is not know.
+// Version 2, Juni 2012, Using Arduino 1.0.1
+//     Adapted to be as simple as possible by Arduino.cc user Krodal
+// Version 3, Feb 26  2013
+//    V3 by louarnold
+// Version 4, March 3, 2013, Using Arduino 1.0.3
+//    by Arduino.cc user Krodal.
+//    Changes by louarnold removed.
+//    Scanning addresses changed from 0...127 to 1...119,
+//    according to the i2c scanner by Nick Gammon
+//    http://www.gammon.com.au/forum/?id=10896
+// Version 5, March 28, 2013
+//    As version 4, but address scans now to 127.
+//    A sensor seems to use address 120.
+//
+// This sketch tests the standard 7-bit addresses
+// Devices with higher bit address might not be seen properly.
+//
+
+
+
+// Pinmap for Bluepill I2Cs (by Testato)
+//
+// I2C-1 standard pins: PB7(sda) PB6(scl)
+// Use it by "Wire" without pin declaration
+//
+// I2C-1 alternative pins: PB9(sda) PB8(scl)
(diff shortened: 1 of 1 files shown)


fpistm:
Thanks @Testato

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
```diff
@@ -0,0 +1,89 @@
...
+//    As version 4, but address scans now to 127.
+//    A sensor seems to use address 120.
+//
+// This sketch tests the standard 7-bit addresses
+// Devices with higher bit address might not be seen properly.
+//
+
+
+
+// Pinmap for Bluepill I2Cs (by Testato)
```
fpistm: I wonder if it's really needed to add specific board pins?

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
```diff
@@ -0,0 +1,89 @@
...
+//
+// I2C-1 standard pins: PB7(sda) PB6(scl)
+// Use it by "Wire" without pin declaration
+//
+// I2C-1 alternative pins: PB9(sda) PB8(scl)
+// TwoWire Wire_alt(PB9,PB8);
+//
+// I2C-2: PB9(sda) PB8(scl)
```
fpistm: If kept, this line need to be updated:
` // I2C-2: PB11(sda) PB10(scl) `

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
```diff
@@ -0,0 +1,89 @@
...
+
+
+
+#include <Wire.h>
+
+
+void setup() {
+
+  Serial.begin(9600);
+  Wire.begin();
```
fpistm: Other method to use another I2C pin or instance, it to update the Pins before call Begin().
This allow to kept Wire instance and declare a new one.
Example for the BP:
```
 Wire.setSDA(PB9);
 Wire.setSCL(PB8);
 Wire.begin();
```
or
```
 Wire.setSDA(PB11);
 Wire.setSCL(PB10);
 Wire.begin();
```

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
Testato: yes, i think it is very useful add at least one examples for a board, and the Bluepill is the most famous ST variant

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
Testato: ok

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
Testato: yes, i also think is better, but you cannot use two I2c in the sametime, so the second must have a second declaration.
I will modifie the example, look if you like it

fpistm: I know :)
I only gave several example to use only one but with different pins possibilities.

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
fpistm: ok

Issue #217, opened 2018-02-02
[I2C] Review I2C SLAVE IT Rx/Tx complete callback


fpistm:
Refers to #216 
Since #135 by @fprwi6labs  (I2C IT mode)
`HAL_I2C_SlaveRxCpltCallback` and `HAL_I2C_SlaveTxCpltCallback` should be used instead of `HAL_I2C_ListenCpltCallback`

Several issue raised around I2C and related to IT mode implementation:
http://stm32duino.com/viewtopic.php?f=29&t=3210&start=10
http://stm32duino.com/viewtopic.php?f=48&t=3612

~http://stm32duino.com/viewtopic.php?f=48&t=3243&p=41486#p41486~ Addressed by #301 

~Will also include #203~  Addressed by #305

MauroMombelli:
Is there any ETA? this to me seems a pretty critical issue, i have been extremely lucky to read this issue before starting to use this core.

Testato:
I2C is one of the fundamental libraries for an Arduino core.
It can not remain damaged for so long.
If there are no resources to dedicate, at least go back to the previous version that was less faulty than this.

fpistm:
I know this is a drawback issue. In several case I2C is working but some use case are broken (mainly master/slave use case). 
I've planned to fix this issue in September. I've several stuff to perform before and it's hard to mitigate btw all those stuff. I try to do my best to answer/improve/fix/review/support all inputs.

This is an open source project, any contribution are welcome and really appreciated.

IMHO, rollback is not an option as it will break other use case. 
If you want you can do it in your own.

MauroMombelli:
>  In several case I2C is working but some use case are broken 

"some user case" is a problem, I would like to start a project with F411 as master, how can i know if i will be affected? Is there any pre-analisis or patch working on or it is still all in the "todo" list?

> IMHO, rollback is not an option as it will break other use case.
> If you want you can do it in your own.

guess now is too late, is there any particular release known as working?

>This is an open source project, any contribution are welcome and really appreciated.

Im implementing a new board and started using this core to create a variant; but i2c is a main core component, and sincerely as this project is version 1.2 and officially backed by ST i was expecting a stable product.
This will basically force me to use some alternative implementation, and im lucky i catch it early and if this is the case i will have "only" a couple of days of precious free time to throw away.

Testato:
> sincerely as this project is version 1.2 and officially backed by ST i was expecting a stable product.
+1

[...]

PR #305, opened 2018-08-13
Configure I2C Fast Mode when I2C freq is more than 100kHz

### LINKED ISSUE #203: I2C: extEEPROM test failed on Nucleo L152RE
Since this #135, extEEPROM test failed on Nucleo L152RE

LMESTM:
##Description

For STM32 families that have a DutyCycle init parameter, we need to
select the Fast Mode dedicated configuration which will set the
fast mode bit in CCR register.

Fixes #203 
so #203 can be removed from the list to be addressed in #217 

## Tests

Tested on NUCLEO_L152RE and NUCLEO_F429ZI with eepromTest example from extEEPROM library.


ghent360:
I'll try it an let you know by the end of the week.

ghent360:
It worked fine on my project.

PR #306, opened 2018-08-13
I2C Slave : rework slave receive data sequence

LMESTM:
## Description 

This work includes some rework of the Slave RX path, which is based on
below principle:
- we don't know in advance how many bytes will be sent by the I2C master
so we'll be listening to bytes 1 by 1
- in order to get them one by one, we're programing the I2C with
HAL_I2C_Slave_Sequential_Receive_IT and 1 byte at a time and we're using
the HAL_I2C_SlaveRxCpltCallback to store the byte then programing again
for next byte.
- this sequence is ended when the HAL_I2C_ListenCpltCallback is called,
which happens when the master ends the ongoing sequence. We can then
prepare for the next one.

In order to implement this mecanism, we're introduced a local counter
slaveRxNbData where we store the number of received bytes, as well as a
new slave mode SLAVE_MODE_LISTEN which allows for extra checks.

i2c_s structure members that can be modified from main context or under
interrupt context have been marked as volatile.

Addresses #217 

## TESTS
Example used: master_reader_writer + slave_sender_recevier
Has been tested with NUCLEO boards F411RE, L152RE, F091RC, L476_RG in various combinations 1 as master, the other as slave, with a X-NUCLEO-IKS01A1 connected on one board to provide the required pull-up resistors on the I2C lines.

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -75,6 +75,8 @@
 
 #define SLAVE_MODE_TRANSMIT     0
 #define SLAVE_MODE_RECEIVE      1
+#define SLAVE_MODE_LISTEN       2
+
 
 /**
   * @}
@@ -283,6 +285,9 @@ void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, u
   HAL_I2C_Init(handle);
 
   obj->isMaster = master;
+  /* Initialize default values */
+  obj->slaveRxNbData = 0;
+  obj->slaveMode = SLAVE_MODE_LISTEN;
 }
 
 /**
@@ -395,8 +400,9 @@ i2c_status_e i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint16_t size)
   // Check the communication status
   for(i = 0; i < size; i++) {
     obj->i2cTxRxBuffer[i] = *(data+i);
-    obj->i2cTxRxBufferSize++;
   }
+
+  obj->i2cTxRxBufferSize = size;
 
   return I2C_OK;
 }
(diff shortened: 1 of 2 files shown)


cores/arduino/stm32/twi.c
```diff
@@ -534,19 +541,56 @@ void HAL_I2C_AddrCallback(I2C_HandleTypeDef *hi2c, uint8_t TransferDirection, ui
   */
 void HAL_I2C_ListenCpltCallback(I2C_HandleTypeDef *hi2c)
 {
-  uint8_t nbData = 0;
   i2c_t *obj = get_i2c_obj(hi2c);
 
+  /*  Previous master transaction now ended, so inform upper layer if needed
+   *  then prepare for listeing to next request */
```
fpistm: listening

cores/arduino/stm32/twi.c
```diff
@@ -534,19 +541,56 @@ void HAL_I2C_AddrCallback(I2C_HandleTypeDef *hi2c, uint8_t TransferDirection, ui
...
+/**
+  * @brief Slave TX complete callback
+  * @param  hi2c Pointer to a I2C_HandleTypeDef structure that contains
+  *                the configuration information for the specified I2C.
+  * @retval None
+  */
+void HAL_I2C_SlaveTxCpltCallback(I2C_HandleTypeDef *hi2c)
+{
+  i2c_t *obj = get_i2c_obj(hi2c);
+  /*  reset transmit buffer size */
```
fpistm: Please, remove one space and use capital letter for first letter.
/* Reset

cores/arduino/stm32/twi.c
LMESTM: done

cores/arduino/stm32/twi.c
LMESTM: done

PR #319, opened 2018-09-13
Issue43

chwon64:
Remove "USB interface" menu and add "USB Serial" under "Serial interface" menu.
Add "System clock" menu.
Fix PLLCLK setting when wakeup from stop and standby mode.


chwon64:
Make wrong pull request. Please ignore it.

PR #331, opened 2018-09-19
Remove GSM built-in library and related code

fpistm:


KJansun:
Will the GSM library added to the standard core?

fpistm:
GSM shield is a retired product, so this is not planned to add a library.
Since 1.7.0, there is a SoftwareSerial built-in library. Maybe it could be port more easily.
Any contributions are welcome.

PR #336, opened 2018-09-21
Fix input string const-ness to avoid warnings with literal strings

ppescher:
I guess nobody at Arduino ever bothers to compile with full warnings enabled, which in fact is not the default compiler setting.

### DIFF
--- a/cores/arduino/Stream.cpp
+++ b/cores/arduino/Stream.cpp
@@ -90,20 +90,20 @@ void Stream::setTimeout(unsigned long timeout)  // sets the maximum number of mi
 }
 
  // find returns true if the target string is found
-bool  Stream::find(char *target)
+bool  Stream::find(const char *target)
 {
   return findUntil(target, strlen(target), NULL, 0);
 }
 
 // reads data from the stream until the target string of given length is found
 // returns true if target string is found, false if timed out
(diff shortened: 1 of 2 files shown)


fpistm:
Hi  @ppescher and thanks for your contribution
I always build with all warnings enabled and try to remove all related to core.
I do not fix HAL/LL which should be fixed across new HAL/LL release.
I never see warning about this class.
Please, could you share an example which rise those warnings?

ppescher:
Hi Frederic,

I know you do, because the build output of the STM32 core is very clean compared to other Arduino cores. I remember that was a pleasant surprise on my first build.

I know this code comes from various sources and contributions, so I do not expect the same quality everywhere. This is just my small contribution to make it a bit better.

My complaint was about the decision to turn all warnings off by default in the Arduino IDE. But this is not the place to argue about that.

One example of code that triggers a warning is as simple as:
```
Serial.find("pippo");
```
With avr-gcc (AVR core) the output message is:
> warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]

With arm-none-eabi-gcc (STM32 core) the message is:
> warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]

PR #337, opened 2018-09-24
Menu options to support floating point formatting functions

ppescher:
When you want to use sscanf() or sprintf() functions to work with floating point numbers the default Newlib Nano runtime will silently fail.

I added a new sub-menu to select the C runtime you want to use:
* Newlib Nano, without floating point support (as before)
* Newlib Nano + Float Printf (only printf supports fp numbers)
* Newlib Nano + Float Scanf (only scanf supports fp numbers)
* Newlib Nano + Float Printf/Scanf (both support fp numbers)
* Newlib Standard (not Nano)

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -6,6 +6,7 @@ menu.xserial=Serial interface
 menu.usb=USB interface
 
 menu.opt=Optimize
+menu.rtlib=C Runtime Library
 menu.upload_method=Upload method
 menu.flash=Flash Memory Size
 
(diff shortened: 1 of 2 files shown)


fpistm:
Hi @ppescher 
Thanks for this contribution, I think it's a good one to provide this kind of capabilities.
There is some issue mainly when using arduino-builder.
In fact the build.flags.ldspecs is overwritten.
Anyway, after a quick check I've found the current implementation show a duplicated argument when using lto: `-Os -flto -flto`
while this should be `-Os -flto --specs=nano.specs`
I guess, It also required to check order of inclusion.

ppescher:
Hi @fpistm 

`arduino-builder` gets called from the IDE so any issue on the command line should come up in the build output window too. I'm not sure what you are referring to. Could you make an example I can reproduce?

`build.flags.ldspecs` is not overwritten. I have checked and the build properties coming from different menu options are merged together.
For example:
`RemRam.menu.opt.o3lto.build.flags.ldspecs=-flto`
Is merged with:
`RemRam.menu.rtlib.nanofs.build.flags.ldspecs=--specs=nano.specs -u _scanf_float`
And the result is:
`build.flags.ldspecs=-flto --specs=nano.specs -u _scanf_float`
Which is correct.

There was in fact an issue with this line of platform.txt:
`compiler.ldflags={build.flags.ldspecs}`
That could end up with duplicated linker flags, but I removed it. 

`compiler.c.elf.flags` does already include `build.flags.ldspecs` as the first arguments, that should be the correct place. I found some reference about usage here: http://distribute.atmel.no/tools/opensource/Atmel-ARM-GNU-Toolchain/4.8.4/readme.txt
> e.g. to output a float, the command line is like: 
> $ arm-none-eabi-gcc --specs=nano.specs -u _printf_float $(OTHER_LINK_OPTIONS)

fpistm:
I've reproduced by inverting optimize menu and the new rtlib one.
Anyway, initially there is an issue in the current implementation.
I will provide soon a version which correct all of those issue.

fpistm:
Close this one as #339 merged

PR #339, opened 2018-09-25
[review #337] + fix on build.flags.ldspec

fpistm:
This PR fix usage of `build.flags.ldspec` to avoid `-flto` option used twice at link time.
This include the #337 Menu options to support floating point formatting functions for @ppescher with some fix:
* Removed default `*.menu.rtlib.nano.build.flags.ldspecs=--specs=nano.specs` as it is the default in platform.txt
* Rename `*.menu.rtlib.full=Newlib Standard` to `*.menu.rtlib.full=Standard` as this is the default libc not the nano one.

~I will still have to investigate the remove of `-Dprintf=iprintf`~


ppescher:
I had read somewhere that `iprintf` is just an alias to `printf` in newlib-nano, so it's not necessary anymore (I guess it's some left-over from the AVR core or old versions of ARM GCC).
Moreover, if you want to be able to use the full `printf` function, that macro definition have to be removed.

Some refs here: https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob_plain;f=newlib/README
>3) Integer-only versions of the formatted I/O functions (the iprintf/
	 iscanf family) simply alias their regular counter-parts.
	 The affected functions are:
	  diprintf vdiprintf
	  siprintf fiprintf **_iprintf_** sniprintf asiprintf asniprintf
	  siscanf fiscanf iscanf ...

fpistm:
That's what I've seen.
Let me know if you are ok with this PR.

ppescher:
Sure, it's fine for me. Thanks

PR #348, opened 2018-10-04
Enhance memory management

### LINKED ISSUE #307: malloc hooks
I am trying to make progress in porting an esp32 arduino sketch to use stm32duino.

In the short term, my progress was impeded by a freertos memory management Issue https://github.com/stm32duino/STM32FreeRTOS/issues/8.

As a temporary workaround  because malloc()/realloc() were failing quite badly, I changed my sketch to use heap_4.c along with a statically-allocated memory buffer.

This worked quite well with one exception: my I2C I/O was failing.

After debugging it, I found that the i2c core driver was doing malloc/realloc() for its buffer management.

This is problematic for two reasons:
- it is nonreentrant, in the case where freertos is operating.  
- It makes the stm32 arduino core dependent upon dynamically allocated memory, even in an environment in which the developer may wish to just use static

As a workaround, I added the code below to my sketch.  I would suggest that the correct workaround, however, is to do something so that the STM32 FreeRTOS code can hook all core allocator requests and reroute them through its own appropriate per-port allocator.

```
#if configAPPLICATION_ALLOCATED_HEAP && !defined(ESP32DUINO)

// Define the static heap                                                                                               
uint8_t ucHeap[configTOTAL_HEAP_SIZE];

/* Prototypes for our hooks, decorated to make sure they are C bindings */
extern "C" {
extern void *malloc(size_t size);
extern void free(void *p);
extern void *realloc(void *p, size_t size);
}

void *malloc(size_t size) {
    return pvPortMalloc(size);
}

void free(void *p) {
    vPortFree(p);
}

void *realloc(void *p, size_t size) {
    void *pnew = pvPortMalloc(size);
    memcpy(pnew, p, size);
    vPortFree(p);
    return pnew;
}

#endif  // configAPPLICATION_ALLOCATED_HEAP                   
```

fpistm:
This PR include:
 * Fix possible issue with realloc usage
 * Define `_sbrk` as weak (Required by https://github.com/stm32duino/STM32FreeRTOS/issues/9)
 * Avoid user heap to override the minimum stack size (Fix #307)
 * Fix Nucleo-L432KC and Nucleo-L476RG SRAM usage

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -282,7 +282,7 @@ Nucleo_64.menu.pnum.NUCLEO_L152RE.build.cmsis_lib_gcc=arm_cortexM3l_math
 Nucleo_64.menu.pnum.NUCLEO_L476RG=Nucleo L476RG
 Nucleo_64.menu.pnum.NUCLEO_L476RG.node=NODE_L476RG
 Nucleo_64.menu.pnum.NUCLEO_L476RG.upload.maximum_size=1048576
-Nucleo_64.menu.pnum.NUCLEO_L476RG.upload.maximum_data_size=131072
+Nucleo_64.menu.pnum.NUCLEO_L476RG.upload.maximum_data_size=98304
 Nucleo_64.menu.pnum.NUCLEO_L476RG.build.mcu=cortex-m4 -mfpu=fpv4-sp-d16 -mfloat-abi=hard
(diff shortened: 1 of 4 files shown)


rayozzie:
I've tested these fixes on the Nucleo-L432KC, and they work perfectly.  It would be helpful to get them released.

Thank you for doing this work!

PR #349, opened 2018-10-04
Allow setting PWM freq

edogaldo:
introduced new function PWMWrite() which allows generating PWM signals with specified freq

### DIFF
--- a/cores/arduino/wiring_analog.c
+++ b/cores/arduino/wiring_analog.c
@@ -84,13 +84,22 @@ void analogWrite(uint32_t ulPin, uint32_t ulValue) {
       dac_write_value(p, ulValue, do_init);
     } else
 #endif //HAL_DAC_MODULE_ENABLED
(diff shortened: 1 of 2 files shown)


edogaldo:
**Note: being the frequency shared among all channels of a specified timers, setting the freq for one channel (pin) will impact all others.**

edogaldo:
No, maybe I'm not..
@fpistm: could you please confirm that PWM_FREQUENCY represents the frequency of the PWM signal?

Thanks in advance, E.

fpistm:
Yes but depending of its value this will not be the case due to the computation done. (If I well remember).
I would not like introduce this PR while the HardwareTimer is not released.

fpistm:
@edolgado, I've do a new PR to provide a way to update Freq.
As I do not want introduce new API naming ( `PWMxxxx()` ) while not review Timer feature.
This only introduce a way to change de he default freq using a Arduinoish API liked `analogWriteResolution`

PR #353, opened 2018-10-09
Allow use of a custom startup file

### LINKED ISSUE #352: Question - How to include custom startup code?
I'm adding support for a couple of boards by Malyan (the M200 V1 and V2, STM32F103 and 070, respectively). I have these working with STM32GENERIC, but these have some odd startup code requirements. The booloader doesn't set SP, so on the 103, I need to load the SP in the startup assembler, and in both cases, there's some form of interrupt that watches a few key values in RAM that cause the printer to fault. If I disable interrupts and faults until after variable initialization, it works fine (at that point, the core's code is in control 100%). 

Right now, I've modified the startup_stm32f103xb.s file to include the required changes, but is there a way we can extend the core to allow a variant to include a startup file instead? I don't mind making such a change, but it's not clear to me yet how the .s files are selected for compliation. I've looked at platform.txt and see the -I settings, but not how we're matching the files.

fpistm:
Syntax in the board.txt:
`xxx.build.startup_file=-DCUSTOM_STARTUP_FILE`

Example to use a custom startup file for Nucleo_L476RG:
`    Nucleo_64.menu.pnum.NUCLEO_L476RG.build.startup_file=-DCUSTOM_STARTUP_FILE`
Then add a "*.S" file in the "variant/NUCLEO_L476RG/" folder
**Important note: extension must be '.S' not '.s'**

`CUSTOM_STARTUP_FILE `could also be defined thanks the build_opt.h.
`CMSIS_STARTUP_FILE `could now be overridden using the build_opt.h, ex:
`-DCMSIS_STARTUP_FILE=\"startup_stm32l476xx.s\"`

Fix #352

### DIFF
--- a/cores/arduino/stm32/startup_stm32yyxx.S
+++ b/cores/arduino/stm32/startup_stm32yyxx.S
@@ -1,4 +1,5 @@
 #include "stm32_def_build.h"
 
+#if defined(CMSIS_STARTUP_FILE)
 #include CMSIS_STARTUP_FILE
-
+#endif
--- a/cores/arduino/stm32/stm32_def_build.h
+++ b/cores/arduino/stm32/stm32_def_build.h
@@ -1,6 +1,7 @@
 #ifndef _STM32_DEF_BUILD_
 #define _STM32_DEF_BUILD_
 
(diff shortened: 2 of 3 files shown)


xC0000005:
I see now how to set it. Thanks.

PR #354, opened 2018-10-09
Allow boards to specify a custom startup file. 

xC0000005:
For boards such as the Malyan M100, M200 V1 and M200 V2 (and pro) the startup assembler must account for some oddities of the bootloader which is already running. Rather than modify the core itself, allow a board to include in its -D defines a CMSIS startup file name which will be included instead.


xC0000005:
I didn't actually want to merge the package.json, but I'm not sure how to remove the commit. I'm sorry.

fpistm:
Hi @xC0000005 ,
Have you seen #353 ?

xC0000005:
I did not see this, but it seems we had identical ways of going about it. Feel free to reject/close mine, your solution works perfectly well.

PR #368, opened 2018-11-05
GPIO enhancement

### LINKED ISSUE #274: Enhance GPIO usage using LL
Using LL instead of HAL will increase GPIO access.

fpistm:
Fix #274 

* Clean up the digital code part
* Use LL instead of HAL for read/Write
* Add `digitalToggle `API
* Add `digital(Read|Write|Toggle)Fast` API (using PinName PY_n)
* Enhance `get_GPIO_Port` (macro instead of a function)


cores/arduino/stm32/PortNames.h
```diff
@@ -73,7 +75,9 @@ typedef enum {
 
 #define MAX_NB_PORT (LastPort-FirstPort+1)
 
-GPIO_TypeDef *get_GPIO_Port(uint32_t port_idx);
+/* Return GPIO base address */
+#define get_GPIO_Port(p) ((p < MAX_NB_PORT) ? GPIOPort[p] : (GPIO_TypeDef *)GPIOA_BASE)
```
ktand: Isn't it better to return `NULL` instead of `GPIOA_BASE` if port is out of bounds?

fpistm: yes. I wonder which value can be the best as if not properly used this would access to 0x00000000. Anyway this could allow to test if it is ok. I will update that.

fpistm: done

PR #388, opened 2018-12-12
[USB] Generic implementation providing  HID composite (Mouse and Keyboard) and CDC in FS/HS

fpistm:
This PR includes HID composite (Keyboard and Mouse) and CDC and will replace #91 and #344.
This new PR will be more generic, no more needs to have `usb/` directory in the variant.
USB pins are in `PeripheralPins.c`, `usbd_desc` and `usbd_conf` are common and handle USB base and USB OTG FS/HS (also FS in HS). 
This means all MCU with USB capabilities could be handled.
PR include your remarks and improvements (I hope).
Several define can be redefined in `variant.h` or thanks the `build_opt.h` including timer which  by default is set to `TIM6 `as well the `TIMx_IRQn` (`CDC_TIM` and `CDC_TIM_IRQn`)
LPM is currently not managed as well the BOS descriptor.

Several tests will be required and also some other optimisation (ex: @edogaldo comments https://github.com/stm32duino/Arduino_Core_STM32/pull/344#issuecomment-427192572 )
`USBSerial::write(const uint8_t *buf, size_t len)` has been added but this have to be rewrote properly.
`readBytes()` could probably be added.

Anyway first tests are promising, see below.
### Before Speed improvement from @hasenbanck and @ktand:
* Disco F746 
  * FS : Average bytes per second = 31983 (Linux)
  * HS: Average bytes per second = 254233 (Linux)

### After Speed improvement from @hasenbanck and @ktand:
 * Disco F746
   * FS: Average bytes per second = 786305 (Windows) / 959592 (Linux)
   * HS: Average bytes per second = 3205058 (Windows) / 2972262 (Linux)
 * Disco F407
   * FS: Average bytes per second = 790577 (Windows) / 845907 (Linux)

Note: Compare to previous PR, USB is not provided as a built-in library.
So, I did not add the `reenumerate()` as the usbd init is not the same. Any comments are welcome.
Updated implementation of `reenumerate()` could be something like that (don't know how manage for USB HS):
```
void USBSerial::reenumerate() {
  /* Re-enumerate the USB */
  volatile unsigned int i;
#ifdef USE_USB_HS_IN_FS
  uint32_t pinDP = pinNametoDigitalPin(USB_OTG_HS_DP);
#elif defined(USB_OTG_FS)
  uint32_t pinDP = pinNametoDigitalPin(USB_OTG_FS_DP);
#else /* USB */
  uint32_t pinDP = pinNametoDigitalPin(USB_DP);
#endif

  pinMode(pinDP, OUTPUT);
  digitalWrite(pinDP, LOW);
  for (i = 0; i < 1512; i++) {
  };
  pinMode(pinDP, INPUT);
  for (i = 0; i < 512; i++) {
  };
}
```


hasenbanck:
Good work. But sadly I can't test the branch right now, since my only working STM32 prototype with USB support is right now at a beta testers place.

Could you maybe provide an example in the example project on how to use the features provided in thid PR?

fpistm:
There is no special example. 
Thanks the menu select the use desired use case. HID (FS or HS) or CDC (FS or HS).
Then for HID you can try USB keyboard or mouse from Arduino sketches.
For CDC use SerialUSB instead of Serial (I will update later to have Serial linked to SerialUSB)

hasenbanck:
Ah okay, I got confused them by your comment about the ```void USBSerial::reenumerate()``` method. If you can just use the Menu inside Arduino, that's even easier. PlatfromIO seem to get support for this core this week, so I'm interested to see how use this feature there.

fpistm:
oh no. My concerns was about the` reenumerate()` provided by @ktand. I did not merge this part of his improvement as the usb init is not the same than you PR. (in fact USB is not moved as a builtin library).
So any comment on this part of code are welcome.

fpistm:
@AnHardt 

> Several define can be redefined in `variant.h` or thanks the `build_opt.h` including timer which by default is set to `TIM6 `as well the `TIMx_IRQn` (`CDC_TIM` and `CDC_TIM_IRQn`)

So use the build_opt.h and `-DCDC_TIM=TIMx -DCDC_TIM_IRQn=TIMx_IRQn` wit x the number of the timer you want

BennehBoy:
How is CDC enabled within a sketch?  Tried a few things but still get an unrecognized USB device in Windows.

fpistm:
You enable it thanks the USB interface menu. Then it depends of your board. HS or FS.

BennehBoy:
Ha, OK so I need to hack boards.txt for the BLACK407VET, thanks.

BennehBoy:
So compiled with USB CDC FS option I can see the board as CDC FS & a Serial USB COM port in USBDeview but it is 'purple' and I cannot select the COM port in the IDE Serial monitor, is there some internal glue that needs to happen to get Serial talking via USB? (I noticed the re-enumerate comments above).

fpistm:
> 
> 
> Ha, OK so I need to hack boards.txt for the BLACK407VET, thanks.

If you get the full PR, the menu should be available.


> 
> 
> So compiled with USB CDC FS option I can see the board as CDC FS & a Serial USB COM port in USBDeview but it is 'purple' and I cannot select the COM port in the IDE Serial monitor, is there some internal glue that needs to happen to get Serial talking via USB? (I noticed the re-enumerate comments above).

Nice it is recognized, anyway you should be able to select it.
To use it in Arduino sketch simply use `SerialUSB `instead of `Serial `or `Serialx`
Could you tell me which host OS version you used (I guess windows) ?
I will make a try as I have the black F407.
As said I need to perform mote tests before merge it so any help are welcome ;) Thanks

BennehBoy:
Host OS is Windows 10.

I have applied the full PR, there are no USB options for GenF4 in the included boards.txt, I just copied the appropriate lines.

Is there anyway to overload Serial to SerialUSB as in Roger's core?

BennehBoy:
Windows 10

Still cannot use the COM port, simple sketch of:

void setup() {
```
SerialUSB.begin(57600);
}

void loop() {
SerialUSB.println("Hello World!");
}
```

Tried also on WIndows 8.1 - OS fails to install drivers for the devices.

BennehBoy:
Have also tried Ubuntu 16.04 LTS, I can See the CDC FS device but not the Serial device via 'sudo lsusb -v'

fpistm:
@BennehBoy 
I've reproduced your issue.
1/ I've forgot to add the USB menu to GENF4. I will add it.
2/ By default, same issue than you. USB recognized but COM port not available.
In fact this is due to the VID/PID. VID is set to `0xleaf` in the `boards.txt`.
As I've hardcoded the PID in the core, it is not well recognized. As I guess for the `0xleaf` PID, the PID for VCP is not the same. 

Set PID in `board.txt` for GENF4 to `0x0483` solve this issue. `0x483 `is for STMicroelectronics

Then VCP works fine.

I've made the benchmark:
```
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
Bytes per second = 960077
Bytes per second = 960015
Bytes per second = 960015
Bytes per second = 960031
Bytes per second = 960031
Bytes per second = 960046
Bytes per second = 960031
Bytes per second = 960077
Bytes per second = 960000
Bytes per second = 768049
Bytes per second = 959985
Bytes per second = 960046
Bytes per second = 960031
Bytes per second = 960077
Bytes per second = 959985
Average bytes per second = 947233
```

So this seems to be a good results for USB FS.

BennehBoy:
Made this change but I still cannot open the COM port :(

I will try on another Win10 machine and report back

BennehBoy:
Here's what device manager says:

`The drivers for this device are not installed. (Code 28)

There are no compatible drivers for this device.


To find a driver for this device, click Update Driver.`

fpistm:
In this case it seems to be a driver issue.
Do you have this driver installed?
https://www.st.com/en/development-tools/stsw-stm32102.html
If yes then try to uninstall it as generic driver should be ok.

BennehBoy:
This is required for Windows 10 also?

BennehBoy:
I don't believe I've ever installed that driver on my dev system, and the other system I tried definitely hasn't had the driver installed.

fpistm:
> 
> 
> I replaced the driver using Zadig and can now use the comport, however two devices are shown 1 of which is still inoperable. Getting there +1

Fine.
When you told you see 2 devices, is it under USBDeview ?

BennehBoy:
It's actually 3, a composite, a Communication Device, and a CDC Data device.  Don't get this on a Windows 8.1 machine.  I'll try on a couple of Windows 10 machines later in case I've goosed this one.

Ubuntu 116.04 LTS Server I see a USB device (interface 0), but no tty gets mapped.

I have a spare machine that I'll boot from Another distro later to see how that copes.

ktand:
@fpistm Just tried the PR and it works just fine in FS mode on the STM32F407 based Armed 3d printer board. Shows up as a "STMicroelectronics Virtual COM port" device in Windows 10, as expected.

romainreignier:
I have tested the PR on Blue F407VE mini board.

I have noticed that I need to unplug then re-plug the USB after flashing in order to make the board recognized by Linux (upload with STLink). But It may be expected.

It works in FS mode but not HS. But may be limited by the hardware?

On the *Standard* pjrc benchmark, I get 141295 bytes/sec. Much lower than your results on the other 407 board (I have tried in both -Os and -O3).

The *ReadBytes* benchmark do not work (nothing seem received by the PC) but as I understand your first post, it is not yet implemented.

> readBytes() could probably be added.

kaofishy:
> I have tested the PR on Blue F407VE mini board.
> 
> It works in FS mode but not HS. But may be limited by the hardware?

The STM32F407VE needs an external HS PHY for the OTG HS, and in any case it looks like only the OTG FS controller is exposed as the micro-USB connector on the Blue mini board (the chip has two USB controllers).

fpistm:
> It works in FS mode but not HS. But may be limited by the hardware?

Of course, it is mainly dependents of the PCB. Some MCU can support FS and HS but as mentioned by @kaofishy this is linked how USB connector are wired

> On the _Standard_ pjrc benchmark, I get 141295 bytes/sec. Much lower than your results on the other 407 board (I have tried in both -Os and -O3).

Yes this is not very high. This could probably be enhance. Anyway this PR is a first step to unlock USB feature. It probably contains some drawbacks or bug.

> The _ReadBytes_ benchmark do not work (nothing seem received by the PC) but as I understand your first post, it is not yet implemented.

Right, this is not implemented.

romainreignier:
@kaofishy and @fpistm Thank you for the interesting information about the need of an external PHY.

@fpistm About pinout, I have noticed that `LED_BLUILTIN` is not defined for this board, should be `PB9`.

> Yes this is not very high (note this also depends of Host OS). This could probably be enhanced. Anyway this PR is a first step to unlock USB feature. 

Yes, don't get me wrong. This PR is awesome. Thanks! And the current performances are more than enough for most of the use cases.
I mentioned the benchmark because I've found my result much lower than yours on the Black F407 board that uses the same processor as far as I know.

fpistm:
@makarenya 
honestly, I don't know. I've just updated to latest STM32 USBD library. I will investigate internally why this has been added.

makarenya:
You don't need to use TIM1 at all.
just move this one
`
do {
       if (lineState == 0) { // Device disconnected
         status = USBD_OK;
 	  } else {
         status = USBD_CDC_TransmitPacket(&hUSBD_Device_CDC);
       }
     } while (status == USBD_BUSY);
`
into write function.

It doesn't change anything, but TIM1 not be used!

makarenya:
About STALL: just delete this line, and the firmware will start working with OpenVRT, and continue to work with the others!

edogaldo:
> 
> 
> You don't need to use TIM1 at all.
> just move this one
> `do { if (lineState == 0) { // Device disconnected status = USBD_OK; } else { status = USBD_CDC_TransmitPacket(&hUSBD_Device_CDC); } } while (status == USBD_BUSY);`
> into write function.
> 
> It doesn't change anything, but TIM1 not be used!

This is +/- what I proposed here: https://github.com/stm32duino/Arduino_Core_STM32/pull/344#issuecomment-427192572

makarenya:
> > You don't need to use TIM1 at all.
> > just move this one
> > `do { if (lineState == 0) { // Device disconnected status = USBD_OK; } else { status = USBD_CDC_TransmitPacket(&hUSBD_Device_CDC); } } while (status == USBD_BUSY);`
> > into write function.
> > It doesn't change anything, but TIM1 not be used!
> 
> This is +/- what I proposed here: [#344 (comment)](https://github.com/stm32duino/Arduino_Core_STM32/pull/344#issuecomment-427192572)

Yes! this is same. 
USBD_CDC_TransmitPacket already have own buffer and it hardware-based one. No need to use any other...

makarenya:
> @makarenya
> honestly, I don't know. I've just updated to latest STM32 USBD library. I will investigate internally why this has been added.
> Referring to Release note:
> `Core: Fix unexpected stall during status OUT phase`
> But this could be also an issue ;)

I am using the latest version of STM32CubeMX. And it includes the same version of the firmware that is used here, but the USB Middleware version V2.4.2 from 2015 is used! It seems to me that I know why.

BennehBoy:
Ran the benchmark on my WIndows 10 Machine - Black407VET6

```
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
Bytes per second = 673265
Bytes per second = 696581
Bytes per second = 690473
Bytes per second = 669785
Bytes per second = 626586
Bytes per second = 652543
Bytes per second = 637687
Bytes per second = 652415
Bytes per second = 667119
Bytes per second = 606146
Bytes per second = 564844
Bytes per second = 578274
Bytes per second = 641698
Bytes per second = 670953
Bytes per second = 674294
Average bytes per second = 646844
```

Slower than Frederic's but not bad.

ktand:
Tried this PR on the STM32F103 (Blue pill). Had to change VID from 0x1eaf to 0x0483 to get Windows to recognize the COM-port:

ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
ignoring startup buffering...
Bytes per second = 450075
Bytes per second = 447081
Bytes per second = 450248
Bytes per second = 445600
Bytes per second = 448119
Bytes per second = 448139
Bytes per second = 450708
Bytes per second = 445822
Bytes per second = 448072
Bytes per second = 448669
Bytes per second = 447808
Bytes per second = 448317
Bytes per second = 450447
Bytes per second = 447942
Bytes per second = 446266
Average bytes per second = 448221

Not that bad...

ktand:
> I have tested the PR on Blue F407VE mini board.
> 
> I have noticed that I need to unplug then re-plug the USB after flashing in order to make the board recognized by Linux (upload with STLink). But It may be expected.

Maybe this is because of the removed `reenumerate()` function?

ktand:
> > > You don't need to use TIM1 at all.
> > > just move this one
> > > `do { if (lineState == 0) { // Device disconnected status = USBD_OK; } else { status = USBD_CDC_TransmitPacket(&hUSBD_Device_CDC); } } while (status == USBD_BUSY);`
> > > into write function.
> > > It doesn't change anything, but TIM1 not be used!
> > 
> > 
> > This is +/- what I proposed here: [#344 (comment)](https://github.com/stm32duino/Arduino_Core_STM32/pull/344#issuecomment-427192572)
> 
> Yes! this is same.
> USBD_CDC_TransmitPacket already have own buffer and it hardware-based one. No need to use any other...

@edogaldo @makarenya @fpistm 

I might be wrong but removing the `UserTxBuffer` and calling `USBD_CDC_TransmitPacket` directly in `USBSerial::write(uint8_t ch)` would cause the next call to `USBD_CDC_TransmitPacket` to fail with `USBD_BUSY` if the previous packet has not been sent. This would also be VERY inefficient as we would be sending 1-byte packets...

edogaldo:
In my version I proposed to put the implementation in the buffer version of
the write and make the byte version call the other with size 1.
This would allow to send a whole buffer in a single call.

Il giorno ven 21 dic 2018 alle 21:08 Karl Andersson <
notifications@github.com> ha scritto:

fpistm:
> > > You don't need to use TIM1 at all.
> > > just move this one
> > > `do { if (lineState == 0) { // Device disconnected status = USBD_OK; } else { status = USBD_CDC_TransmitPacket(&hUSBD_Device_CDC); } } while (status == USBD_BUSY);`
> > > into write function.
> > > It doesn't change anything, but TIM1 not be used!
> > 
> > 
> > This is +/- what I proposed here: [#344 (comment)](https://github.com/stm32duino/Arduino_Core_STM32/pull/344#issuecomment-427192572)
> 
> Yes! this is same.
> USBD_CDC_TransmitPacket already have own buffer and it hardware-based one. No need to use any other...

@edogaldo , @makarenya  could you propose a patch based on this PR ? This would be great.
Thanks in advance

> I am using the latest version of STM32CubeMX. And it includes the same version of the firmware that is used here, but the USB Middleware version V2.4.2 from 2015 is used! It seems to me that I know why.

@makarenya, in fact this is not right. STM32CubeMX use the version included in the Cube FW package.
So, if you try F7 then you will have v 2.5.0. 

| STM32 | USB device version |
| --- | :---: |
| F0 | 2.4.2 |
| F1 | 2.4.2 |
| F2 | 2.4.2 |
| F3 | 2.4.2 |
| F4 | 2.4.2 |
| F7 | 2.5.0 |
| H7 | 2.4.2 |
| L0 | 2.5.1 |
| L1 | 2.4.2 |
| L4 | 2.5.0 |

Those with old version will probably be updated if new release are done.
Anyway about the stall issue, I will raise this issue internally to ST guy supporting this.

> > I have tested the PR on Blue F407VE mini board.
> > I have noticed that I need to unplug then re-plug the USB after flashing in order to make the board recognized by Linux (upload with STLink). But It may be expected.
> 
> Maybe this is because of the removed `reenumerate()` function?

I think yes. that's why I would like to review how USB is init. Maybe, add it to `begin()` method should be fine.

> Tried this PR on the STM32F103 (Blue pill). Had to change VID from 0x1eaf to 0x0483 to get Windows to recognize the COM-port:

@ktand  interesting, I'd never success to have the BP properly working with this PR. I've always have `unrecognised USB device` even with VID set to `0x0483`. Maybe this is due to wrong resistor but with Roger's core this is well recognized :(

BennehBoy:
On bluepill I get a USB composite device reported in USBDeview but no COM port

ktand:
> In my version I proposed to put the implementation in the buffer version of the write and make the byte version call the other with size 1. This would allow to send a whole buffer in a single call.

You can only transmit up to 64 bytes in each call to `USBD_CDC_TransmitPacket`. Writing more than 64-bytes in `USBSerial::write(const uint8_t *buffer, size_t size)` would require a loop which would block until the previous packet is transmitted. Again not very efficient.

Also, this way, sending single byte characters would cause one packet to be sent for each character.

But again, I might be wrong. As @fpistm suggested, please try your solution and then we can compare performance.

Maybe there is some other way to get notified when the USB packet has been sent so we can push a new packet from the UserTxBuffer without using a timer. But without using a timer I don't think there is a way to get multiple calls to write(uint8_t ch) to be sent in the same packet.

ktand:
@fpistm 

> @ktand interesting, I'd never succeed to have the BP properly working with this PR. I've always have `unrecognised USB device` even with VID set to `0x0483`. Maybe this is due to wrong resistor but with Roger's core this is well recognized :(

I have replaced the R10 resistor with 1.5k.

ktand:
@BennehBoy 

> @ktand I'll test on a bluepill here, I have a 1.8k resistor soldered between 3.3v and PA12.

What is the value of the R10 resistory on your BP? 10K or 4.7K? I think a 1.8k resistor between 3.3v and PA12 is the right value when R10 is 10K.

fpistm:
> 
> 
> On bluepill I get a USB composite device reported in USBDeview but no COM port

Do you replace VID by 0x0483  instead of 0x1eaf ?

> I have replaced the R10 resistor with 1.5k.
 
@ktand  ok I will try but if this work I could not explain why it worked with Roger's core .... timing issue?

ktand:
> On bluepill I get a USB composite device reported in USBDeview but no COM port

I get the same when using the 0x1eaf VID, a DFU and a COM-port device. Both in error state. When using the 0x0483 VID I get a single STM Virtual Comport that is working.

ktand:
> I thought you said you never had it working.... confused.

@BennehBoy It was @fpistm that said that. I did get it working (see speed results above)

BennehBoy:
> > I thought you said you never had it working.... confused.
> 
> @BennehBoy It was @fpistm that said that. I did get it working (see speed results above)

Sorry misread the post!

makarenya:
Yes. I make my [own implementation](https://github.com/makarenya/Arduino_Core_STM32), without any timers. But I really don’t know what to do next? Maybe make a request to this repository?

makarenya:
I tried to achieve the following goals: remove the dependency on the timer, and reduce the number of memcpy operations. 

For example, the STM32F103C8T6 doesn't have so many timers, and using one of them for the CDC, is not so good. 

And memcpy operations. 
In fact, they occur too often. First, in the USBSerial::write method the data is copied from the caller buffer into the UserTxBuffer, then the CDC_TIM_PeriodElapsedCallback the same data is copied into the StackTxBuffer, this data is transferred to USBD_CDC_SetTxBuffer without copying, and from where they are copied to PMA. 3 copy operations, and I managed two.

Same situation with the receive.

edogaldo:
> @edogaldo , @makarenya could you propose a patch based on this PR ? This would be great.
How to propose it? I'm not that strong with github...

romainreignier:
> Another question: where can I find the benchmarking sketch?

In the zip in that page https://www.pjrc.com/teensy/benchmark_usb_serial_receive.html#code

edogaldo:
That's for receiving speed, is there any benchmark for sending speed?

romainreignier:
> That's for receiving speed, is there any benchmark for sending speed?

I am not aware of such benchmark.

ktand:
@fpistm 

I just noticed that one of the PWM-pins has been incorrectly commented out in variants/ARMED_V1/PeripheralPins.c in this PR:

`//  {PC_6,  TIM3,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF2_TIM3, 1, 0)}, // TIM3_CH1`

fpistm:
> How to propose it? I'm not that strong with github...

To propose a patch to this PR, you have to propose a PR on my branch PR from my fork. I know this is a bit confuse. This could be done easily when doing some practice.
Anyway you could post a zip then I will be able to try it.

> Another question: where can I find the benchmarking sketch?

Currently, I've only used the one from PJRC which test send speed.

About, win10 issue with Blue F4. You should have VID to 0x0483. This should be ok.
Could you use USBDeview to see all info about it. Maybe there is a previous driver installed for this device with the driver for 0x1EAF ?

variants/ARMED_V1/PeripheralPins.c
```diff
@@ -26,168 +26,170 @@
...
+//  {PB_10, TIM2,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM2, 3, 0)}, // TIM2_CH3 - Mic CLK
+//  {PB_11, TIM2,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM2, 4, 0)}, // TIM2_CH4
+//  {PB_13, TIM1,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM1, 1, 1)}, // TIM1_CH1N
+//  {PB_14, TIM1,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM1, 2, 1)}, // TIM1_CH2N
+//  {PB_14, TIM8,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF3_TIM8, 2, 1)}, // TIM8_CH2N
+//  {PB_14, TIM12,  STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF9_TIM12, 1, 0)}, // TIM12_CH1
+//  {PB_15, TIM1,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM1, 3, 1)}, // TIM1_CH3N
+//  {PB_15, TIM8,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF3_TIM8, 3, 1)}, // TIM8_CH3N
+//  {PB_15, TIM12,  STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF9_TIM12, 2, 0)}, // TIM12_CH2
+//  {PC_6,  TIM3,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF2_TIM3, 1, 0)}, // TIM3_CH1
```
fpistm: Need to be uncomment. (Thanks @ktand https://github.com/stm32duino/Arduino_Core_STM32/pull/388#issuecomment-450177885)

edogaldo:
> Anyway you could post a zip then I will be able to try it.

here you are: 
[USB_patch.zip](https://github.com/stm32duino/Arduino_Core_STM32/files/2712911/USB_patch.zip)

> About, win10 issue with Blue F4. You should have VID to 0x0483. This should be ok.

I have VID = 0x0483 but it's still not working..

> Could you use USBDeview to see all info about it. Maybe there is a previous driver installed for this device with the driver for 0x1EAF ?

Have to check anyway if I run the CDC Cube samples (which use VID = 0x0483 and PID = 0x5740), the VCP is correctly recognised and working fine so it doesn't seem to me really a driver issue..

fpistm:
I did not check all variant USB mapping. Maybe not all the pins are relevant for this one? Only DP and DM are useful? https://github.com/stm32duino/Arduino_Core_STM32/pull/388/files#diff-1a87a375c2193f692f7ee494ba87b465R361

BennehBoy:
@edogaldo, the only way I could make it work was to use zadig to substitute its USB Serial CDC driver for the 'interface 1' portion of the FS device.

I 'think' this may be due to having previously installed an incorrect driver, but I've not had time to check on a fresh system.

kaofishy:
I tried this on a NUCLEO-L432KC with a USB breakout connected to the DP and DM pins (which worked for the Cube examples) and it compiles and uploads properly with the CDC Full Speed option enabled, but the USB device is not enumerated. The rest of the sketch runs fine. I suspect it's something to do with clock setup, since the L4 has a plethora of clock sources for the USB.

cores/arduino/stm32/usb/usbd_desc.c
```diff
@@ -1,315 +1,396 @@
...
+  USBD_CDC_InterfaceStrDescriptor,
+};
+
+/* USB Standard Device Descriptor */
+__ALIGN_BEGIN uint8_t USBD_CDC_DeviceDesc[USB_LEN_DEV_DESC] __ALIGN_END = {
+  0x12,                       /* bLength */
+  USB_DESC_TYPE_DEVICE,       /* bDescriptorType */
+  0x00,                       /* bcdUSB */
+  0x02,
+  0x00,                       /* bDeviceClass */
```
edogaldo: this should be "0x02" to fix the driver issue (ref std Cube implementation)

cores/arduino/stm32/usb/usbd_desc.c
```diff
@@ -1,315 +1,396 @@
...
+  0x00,                       /* bDeviceSubClass */
```
edogaldo: this should be "0x02" to fix the driver issue (ref std Cube implementation)

makarenya:
Maybe the reason is that the re-numbering is not performed during the STM32 reboot?
Deasserting DP line to zero for 100ms after rebooting...

edogaldo:
> @edogaldo, the only way I could make it work was to use zadig to substitute its USB Serial CDC driver for the 'interface 1' portion of the FS device.
> 
> I 'think' this may be due to having previously installed an incorrect driver, but I've not had time to check on a fresh system.

No, this is due to the configuration of "USBD_CDC_DeviceDesc" in cores/arduino/stm32/usb/usbd_desc.c, in particular the values for bDeviceClass and bDeviceSubClass: making both to "0x02" instead of "0x00" makes the driver working fine.

BennehBoy:
@edogaldo thanks I'll give it a try.

BennehBoy:
I applied your patch and recompiled with a minimal sketch outputting continuously to SerialUSB for for Black407VET6 + USB CDC FS + Serial Interaced Disabled.

On 4 different Windows 10 machines I still get 3 USB devices, a composite, a 'Communications Device' & a USB Serial device.  The latter does not function, the 'Communications Device' will function if a Serial USB driver is associated with it.

USBDeview info below:
```
BLACK_F407VE CDC in FS Mode	USB Composite Device	Unknown	Yes	Yes	No	No		9y4i36	03/01/2019 14:56:04	03/01/2019 14:23:10	0483	5740	2.00	00	00	00		DESKTOP-V44DKFD			7&d6c0f4b&2	usbccgp	@usb.inf,%GenericParent.SvcDesc%;Microsoft USB Generic Parent Driver	usbccgp.sys		(Standard USB Host Controller)		100 mA	2.00	USB Composite Device	10.0.17134.1	Composite.Dev.NT	usb.inf	USB\VID_0483&PID_5740\397934693336	Removable, UniqueID, SurpriseRemovalOK	
0003.0000.0000.011.000.000.000.000.000	USB Serial Device	Communication	Yes	No	No	No	COM27		03/01/2019 14:37:40	03/01/2019 14:37:40	0483	5740	2.00	02	02	01		DESKTOP-V44DKFD				usbser	@usbser.inf,%UsbSerial.DriverDesc%;Microsoft USB Serial Driver	usbser.sys		Microsoft	USB Serial Device (COM27)			USB Serial Device	10.0.17134.1	UsbSerial_Install	usbser.inf	USB\VID_0483&PID_5740&MI_00\7&d6c0f4b&2&0000		
0003.0000.0000.011.000.000.000.000.000	USB Serial Device	CDC Data	Yes	Yes	No	No	COM28		03/01/2019 14:56:04	03/01/2019 14:35:44	0483	5740	2.00	0a	00	00		DESKTOP-V44DKFD				usbser	@usbser.inf,%UsbSerial.DriverDesc%;Microsoft USB Serial Driver	usbser.sys		Microsoft	USB Serial Device (COM28)			USB Serial Device	10.0.17134.1	UsbSerial_Install	usbser.inf	USB\VID_0483&PID_5740&MI_01\7&d6c0f4b&2&0001	SurpriseRemovalOK	
```

edogaldo:
Pay attention to apply the change to USBD_CDC_DeviceDesc and not to USBD_HID_DeviceDesc: to me this change solved the driver issue with Win10.

BennehBoy:
*blush*

I changed the values but at lines 104 & 105.... Ooopsie!

Working now!

uzi18:
@fpistm does hid bootloader still supported with this PR?

fpistm:
> 
> 
> @fpistm does hid bootloader still supported with this PR?

You talk about this one:
https://github.com/Serasidis/STM32_HID_Bootloader

So no, I did not test with this bootloader. This one was mainly used by Roger's core.

uzi18:
Yes it just checks for magic sequence on serial buffer and launch bootloader.

uzi18:
In facts there are some stm32 cores available and did not found comparision.
What do you think about this hid stuff?

BennehBoy:
> In facts there are some stm32 cores available and did not found comparision.
> What do you think about this hid stuff?

I think it's probably straightforward to add support for many boards, and should either be a seperate PR or an issue raised as a feature request.  PR #351 talks about using a dfu bootloader & Frederic mentioned that he's looking a the options.  My 2 cents would be to go with dfu as it's much faster...

uzi18:
Hid does not need any special drivers installed. Is it possible for you to check speed?
From code it is look like it is just copy flash blocks over into device by libusb.

fpistm:
Some update:
* Fix STALL issue in STM32 USB device as mentioned by @makarenya. Thanks ;) I will report internally.
* Some small fixes.
* Tested on F1/F2/F3/F4/F7 and L0/L1/L4. HID composite and CDC are up and running. See [usb_serial_receive_results.xlsx](https://github.com/stm32duino/Arduino_Core_STM32/files/2746352/usb_serial_receive_results.xlsx)

Next Step:
* Timer usage
* Serial definition. How to manage? If USB then Serial == SerialUSB ? (Any comments on this are welcome)
* reenumerate()
* ReadBytes()
* ... ?

Maybe it should be better to merge this and then provide those steps in other PR? Any comments on this are also welcome :)

BennehBoy:
> Maybe it should be better to merge this and then provide those steps in other PR? Any comments on this are also welcome :)

I think you should merge and enhancements can come on other PR's

Getting this base functionality across all the variants makes life a lot easier.

edogaldo:
I really don't like a timer based solution but I see the point for having a quick win solution to be improved/reviewed later..

fpistm:
I don't like too I just think to merge to ease next step and follow better improvements. It's hard to follow all comments and do not forget one 😉 I will start tomorrow review and test of your solutions to avoid TIM usage ans would like also test send speed.

ktand:
> Maybe it should be better to merge this and then provide those steps in other PR? Any comments on this are also welcome :)

Yes, I think this is a good idea.

BennehBoy:
Not sure if this is an issue/optimisation opportunity for future PR's but compiling _with_  CDC enabled results in ~10K more SRAM usage than without.  That's a big hit for smaller boards...

fpistm:
Right, this is mainly due to the buffer, hereafter .map extraction.

| Source code | Name | Size dec | Size hex | Section | Address |
| --- | --- | --- | --- | --- | --- |
core.a(usbd_cdc_if.c.o) | StackTxBuffer | 2048 | 0x800 | .bss | 0x20000464
core.a(usbd_cdc_if.c.o) | UserTxBuffer | 2048 | 0x800 | .bss | 0x20000c64
core.a(usbd_cdc_if.c.o) | UserRxBuffer | 2048 | 0x800 | .bss | 0x20001464
core.a(usbd_conf.c.o) | g_hpcd | 1008 | 0x3f0 | .bss | 0x20001ca4
core.a(usbd_interface.c.o) | hUSBD_Device_CDC | 668 | 0x29c | .bss | 0x20002194
core.a(usbd_desc.c.o) | USBD_StrDesc | 256 | 0x100 | .bss | 0x20002094

This could probably optimized

BennehBoy:
Ok I'll do a quick compare with Roger's core later to see what buffer size
it uses.

BennehBoy:
Not sure if I'm barking up the wrong tree and these are sub buffers, but Serial TX & RX buffers for CDC are 256 bytes in rogers core (on F1 at least):

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/6a9c7956538892fa6fbe4a9ffc90d3c4bfab9441/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c#L263

makarenya:
Pardon, but what about my PR (into this PR). I use smaller buffers (only 64b for transmitting and 128b for sending, when this PR uses 2048b buffer for receiving and TWO 2048b buffers for sending). My PR not used timers to determine the moment, when it can continue sending data. Instead, it uses a USB interrupt to detect this moment (which is somewhat

BennehBoy:
I think your PR has conflicts - perhaps rebase now that this one is merged, and just include the changes - would be v welcome from me, or split into two?  One for timer removal, one for buffer size reduction...

makarenya:
Buffer sizes can be reduced in the current PR. They are simply chosen thoughtlessly large. Maximum size of USB packet in full speed mode can't be greater when 64b. So 128b is enough. For the high speed devices, maximum size of the buffer is 512b. so you need only 1k, not 2k. 

Just write this define:
```
#if USE_USB_HS
#define CDC_MAX_PACKET_SIZE (USB_HS_MAX_PACKET_SIZE * 2)
#else
#define CDC_MAX_PACKET_SIZE (USB_FS_MAX_PACKET_SIZE * 2)
#endif
```

And you will have extra small and compact buffers for small devices and two time smaller buffers in devices with HS support (but they have more RAM, and it's not will a problem for it)

Timer reduction and extreme memory optimization PR i make in few days.

BennehBoy:
I think you just need to rebase your existing PR against master - 388 has been merged already.

makarenya:
Yes, give me few days, please!

BennehBoy:
I'm not pressuring you :smiley:

roc2:
@fpistm It is not working when add Keyboard and Mouse in one sketch, but it work well when using Keyboard or Mouse standard-alone. They can't work together?? I test this with version 1.5.0, the testing code:

#include <Mouse.h>
#include <Keyboard.h>

void setup()
{
  Keyboard.begin();
  
  Mouse.begin();
}

void loop()
{
  Keyboard.write('w');

  Mouse.move(100,100);
  delay(1000);
  Mouse.move(200,300);
  
  delay(2000);
  
}

fpistm:
@roc2 normally yes. I've tested with both on the same sketch.
Which host OS. Which board?...

roc2:
@fpistm My board is Black STM32F407VE, on win10 OS.

fpistm:
Tested with 1.5.0 on Win10 it works. 
Anyway it fails on repo master. Linked to F4 HAL update.

PR #391, opened 2018-12-17
Do not provide some default macro's

### LINKED ISSUE #389: Overly-generic defines in utils.h break some compilations
This core contains a few very generic defines in utils.h: https://github.com/stm32duino/Arduino_Core_STM32/blob/fa0dcc8b0575e4142178f6511557dad9abe45cb3/cores/arduino/utils.h#L6-L13

Especially the `str` and `xstr` defines are problematic, since you might be using those as function or method names, which triggers the preprocessor. I ran into this issue when using the boost library, which has a `str` method somewhere, but this can be easily reproduced with this sketch:

```
void str() { }
void setup() { }
void loop() { }
```
Which gives:

```
In file included from /home/matthijs/.arduino15/packages/STM32/hardware/stm32/1.4.0/cores/arduino/wiring.h:31:0,
                 from /home/matthijs/.arduino15/packages/STM32/hardware/stm32/1.4.0/cores/arduino/Arduino.h:32,
                 from /tmp/arduino_build_gdfirmware/sketch/sketch_dec13a.ino.cpp:1:
sketch_dec13a:1:10: error: expected unqualified-id before string constant
 void str() { }
          ^
/home/matthijs/.arduino15/packages/STM32/hardware/stm32/1.4.0/cores/arduino/utils.h:13:17: note: in definition of macro 'str'
 #define str(s) #s
```

Here, it points to the problematic macro, but in my original problem case the second message was not present, making this particularly tricky to diagnose.

As for resolving this:
 - A grep suggests that the `str()` and `xstr()` macros are not actually used in the entire core, so I suggest removing them.
 - The `CONCAT` and `CONCATS` macros could perhaps be rename as well, but better would be to not expose them from header files. grep shows that these macros are only used in .c files, so simply including `utils.h` from those .c files, and dropping the include from `wiring.h` sounds like the best approach to me.

fpistm:
Include utils.h manually will provide them
Note: `str()` and `xstr()` are now capitalized.

Fix #389 

@matthijskooijman : any comment are welcome ;)

### DIFF
--- a/cores/arduino/utils.h
+++ b/cores/arduino/utils.h
@@ -1,15 +1,13 @@
 #ifndef __UTILS_H
 #define __UTILS_H
 
-#include "avr/dtostrf.h"
-
-// Concatenate 2 strings
+/* Concatenate 2 strings */
 #define CONCAT(s1, s2) (s1 s2)
-// Concatenate 2 strings separated by space
+/* Concatenate 2 strings separated by space */
 #define CONCATS(s1, s2) (s1" " s2)
 
-// Stringification
-#define xstr(s) str(s)
-#define str(s) #s
+/* Stringification */
+#define XSTR(s) STR(s)
+#define STR(s) #s
 
 #endif
--- a/cores/arduino/wiring.h
+++ b/cores/arduino/wiring.h
@@ -26,9 +26,9 @@
 #include <string.h>
 #include <math.h>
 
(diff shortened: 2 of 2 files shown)


fpistm:
Luckily `utils.h` was already include in c files (mainly `usbd_desc.c`).

matthijskooijman:
Great!

PR #415, opened 2019-01-21
Hid/DFU boot loader (superseded by #525)

BennehBoy:
This PR enables the use of the STM32 HID Bootloader 3.0 for Generic F103 & Generic F4 variants

Tested on Bluepill, Blackpill (using BP menu option), & Black407VET6.

**Limitations:**
* boards.txt glue probably needs to be a bit more elegant
* F1 boards must use the latest 2K HID bootloader -> https://github.com/Serasidis/STM32_HID_Bootloader  (hid_flash tool will warn if an older version is installed)

**Usage:**

First apply this PR to your STM32 tools -> https://github.com/stm32duino/Arduino_Tools/pull/33.

**HID:**
Select your variant in the boards menu
Set the upload method to "HID Bootloader 3.0 (0x0800)" for F1 or "HID Bootloader 3.0 (0x4000)" for F4
Ensure that "USB Interface (if available):" is set to "CDC Full Speed" (required for DTR toggling)

**DFU:**
_The PR applies the magic word & byte code recognition to USB_CDC_IF.cpp which also enables STM32Duino Bootloader 2.0 to work.  ~~Also included is DISC pin handling for Maple Mini~~. This is the same logic used in Roger Clarke's core and in STM32GENERIC_
Select your variant in the boards menu
Set the upload method to "Maple DFU Bootloader 2.0" or "Maple DFU Bootloader original (0x5000)" depedning on the BL installed.
Ensure that "USB support (if available):" is set to either CDC option (required for DTR toggling)

Note - edited due to changes to HID Bootloader 3.0 on 21/1/19. & addition of DFU 27/1/2019

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -638,6 +638,24 @@ GenF1.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
 GenF1.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
 GenF1.menu.upload_method.bmpMethod.upload.tool=bmp_upload
 
+GenF1.menu.upload_method.hidMethod=HID Bootloader 3.0 (0x0800)
+GenF1.menu.upload_method.hidMethod.upload.protocol=hid3
+GenF1.menu.upload_method.hidMethod.upload.tool=hid_upload
+GenF1.menu.upload_method.hidMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x0800 -DHIDBL_F1
+
+GenF1.menu.upload_method.dfu2Method=Maple DFU Bootloader 2.0
+GenF1.menu.upload_method.dfu2Method.upload.protocol=maple
+GenF1.menu.upload_method.dfu2Method.upload.tool=maple_upload
+GenF1.menu.upload_method.dfu2Method.upload.usbID=1EAF:0003
+GenF1.menu.upload_method.dfu2Method.upload.altID=2
+GenF1.menu.upload_method.dfu2Method.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x2000 -DDFUBL2_F1 -DUSB_DISC_PIN=PB9
+
+GenF1.menu.upload_method.dfuoMethod=Maple DFU Bootloader original (0x5000)
+GenF1.menu.upload_method.dfuoMethod.upload.protocol=maple
+GenF1.menu.upload_method.dfuoMethod.upload.tool=maple_upload
+GenF1.menu.upload_method.dfuoMethod.upload.usbID=1EAF:0003
+GenF1.menu.upload_method.dfuoMethod.upload.altID=1
+GenF1.menu.upload_method.dfuoMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x5000 -DDFUBLO_F1  -DUSB_DISC_PIN=PB9
 ################################################################################
 # Generic F4
 
@@ -734,6 +752,11 @@ GenF4.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
 GenF4.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
 GenF4.menu.upload_method.bmpMethod.upload.tool=bmp_upload
 
+GenF4.menu.upload_method.hidMethod=HID Bootloader 3.0 (0x4000)
+GenF4.menu.upload_method.hidMethod.upload.protocol=hid3
+GenF4.menu.upload_method.hidMethod.upload.tool=hid_upload
+GenF4.menu.upload_method.hidMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x4000 -DHIDBL_F4
+
 ###############################
 # Sparky Flight Controllers
 
@@ -761,7 +784,7 @@ Sparky.menu.upload_method.STLinkMethod=STLink
 Sparky.menu.upload_method.STLinkMethod.upload.protocol=STLink
 Sparky.menu.upload_method.STLinkMethod.upload.tool=stlink_upload
 
-################################################################################
+###############################
 # RAK boards
 
 RAK.name=RAKWireless
--- a/boards.txt
+++ b/boards.txt
@@ -641,21 +641,24 @@ GenF1.menu.upload_method.bmpMethod.upload.tool=bmp_upload
 GenF1.menu.upload_method.hidMethod=HID Bootloader 3.0 (0x0800)
 GenF1.menu.upload_method.hidMethod.upload.protocol=hid3
 GenF1.menu.upload_method.hidMethod.upload.tool=hid_upload
+GenF1.menu.upload_method.hidMethod.upload.bl_size=0x800
 GenF1.menu.upload_method.hidMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x0800 -DHIDBL_F1
 
 GenF1.menu.upload_method.dfu2Method=Maple DFU Bootloader 2.0
 GenF1.menu.upload_method.dfu2Method.upload.protocol=maple
 GenF1.menu.upload_method.dfu2Method.upload.tool=maple_upload
+GenF1.menu.upload_method.dfu2Method.upload.bl_size=0x2000
 GenF1.menu.upload_method.dfu2Method.upload.usbID=1EAF:0003
 GenF1.menu.upload_method.dfu2Method.upload.altID=2
-GenF1.menu.upload_method.dfu2Method.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x2000 -DDFUBL2_F1 -DUSB_DISC_PIN=PB9
+GenF1.menu.upload_method.dfu2Method.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x2000 -DUSB_DISC_PIN=PB9
 
 GenF1.menu.upload_method.dfuoMethod=Maple DFU Bootloader original (0x5000)
 GenF1.menu.upload_method.dfuoMethod.upload.protocol=maple
 GenF1.menu.upload_method.dfuoMethod.upload.tool=maple_upload
+GenF1.menu.upload_method.dfuoMethod.upload.bl_size=0x5000
 GenF1.menu.upload_method.dfuoMethod.upload.usbID=1EAF:0003
 GenF1.menu.upload_method.dfuoMethod.upload.altID=1
-GenF1.menu.upload_method.dfuoMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x5000 -DDFUBLO_F1  -DUSB_DISC_PIN=PB9
+GenF1.menu.upload_method.dfuoMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x5000 -DUSB_DISC_PIN=PB9
 ################################################################################
 # Generic F4
 
@@ -755,6 +758,7 @@ GenF4.menu.upload_method.bmpMethod.upload.tool=bmp_upload
 GenF4.menu.upload_method.hidMethod=HID Bootloader 3.0 (0x4000)
 GenF4.menu.upload_method.hidMethod.upload.protocol=hid3
 GenF4.menu.upload_method.hidMethod.upload.tool=hid_upload
+GenF4.menu.upload_method.hidMethod.upload.bl_size=0x4000
 GenF4.menu.upload_method.hidMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x4000 -DHIDBL_F4
 
 ###############################
--- a/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
+++ b/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
@@ -187,6 +187,7 @@ static int8_t USBD_CDC_Control(uint8_t cmd, uint8_t *pbuf, uint16_t length)
         transmitStart = 0;
       }
 	  dtr_pin++; //DTR pin is enabled
+
       break;
 
     case CDC_SEND_BREAK:
@@ -220,31 +221,31 @@ static int8_t USBD_CDC_Receive(uint8_t *Buf, uint32_t *Len)
 {
   /* USER CODE BEGIN 6 */
   /* Four byte is the magic pack "1EAF" that puts the MCU into bootloader. */
-  if(*Len >= 4){
-	/**
-	 * Check if the incoming contains the string "1EAF".
-	 * If yes, check if the DTR has been set, to put the MCU into the bootloader mode.
-	 */
-    if(dtr_pin > 3){
-      if((Buf[0] == '1')&&(Buf[1] == 'E')&&(Buf[2] == 'A')&&(Buf[3] == 'F')){
+  if (*Len >= 4) {
+    /**
+     * Check if the incoming contains the string "1EAF".
+     * If yes, check if the DTR has been set, to put the MCU into the bootloader mode.
+     */
+    if (dtr_pin > 3) {
+      if ((Buf[0] == '1') && (Buf[1] == 'E') && (Buf[2] == 'A') && (Buf[3] == 'F')) {
 
-#if defined (HIDBL_F1)  
-      RTC_HandleTypeDef hrtc;
-      __HAL_RCC_PWR_CLK_ENABLE();
-      __HAL_RCC_BKP_CLK_ENABLE();
-      HAL_PWR_EnableBkUpAccess();
-      HAL_RTCEx_BKUPWrite(&hrtc,RTC_BKP_DR10,0x424C);  //Write the magic number 0x424C  
-        
+#if defined (HIDBL_F1)
+        RTC_HandleTypeDef hrtc;
+        __HAL_RCC_PWR_CLK_ENABLE();
+        __HAL_RCC_BKP_CLK_ENABLE();
+        HAL_PWR_EnableBkUpAccess();
+        HAL_RTCEx_BKUPWrite(&hrtc, RTC_BKP_DR10, 0x424C); //Write the magic number 0x424C
+
 #elif defined (HIDBL_F4)
-      __HAL_RCC_PWR_CLK_ENABLE();
-      HAL_PWR_EnableBkUpAccess();
-      __BKPSRAM_CLK_ENABLE();
-      *(__IO uint32_t *)(BKPSRAM_BASE) = 0x424C; //Write the magic number 0x424C at Backup SRAM address 0x40024000    
-        
-#endif       
-      HAL_NVIC_SystemReset();
+        __HAL_RCC_PWR_CLK_ENABLE();
+        HAL_PWR_EnableBkUpAccess();
+        __BKPSRAM_CLK_ENABLE();
+        *(__IO uint32_t *)(BKPSRAM_BASE) = 0x424C; //Write the magic number 0x424C at Backup SRAM address 0x40024000
+
+#endif
+        HAL_NVIC_SystemReset();
       }
-    dtr_pin = 0;
+      dtr_pin = 0;
     }
   } /* USER CODE END */
   UNUSED(Buf);
--- a/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
+++ b/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
@@ -186,7 +186,7 @@ static int8_t USBD_CDC_Control(uint8_t cmd, uint8_t *pbuf, uint16_t length)
       if (lineState) { // Reset the transmit timeout when the port is connected
         transmitStart = 0;
       }
-	  dtr_pin++; //DTR pin is enabled
+      dtr_pin++; //DTR pin is enabled
 
       break;
 
--- a/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
+++ b/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
@@ -55,6 +55,10 @@ __IO bool receivePended = true;
 static uint32_t transmitStart = 0;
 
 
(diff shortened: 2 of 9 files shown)


BennehBoy:
The F4 issue is due to BOOT0 being set low, remove the jumper and the device resets into the sketch.

RickKimball:
You probably want to adjust the total available flash:
```
#ifdef HIDBL_F1
FLASH (rx)      : ORIGIN = 0x8000800, LENGTH = 64K
#else
```
might be:
```
#ifdef HIDBL_F1
FLASH (rx)      : ORIGIN = 0x8000800, LENGTH = 62K
#else
```
goes for all .ld scripts

boards.txt
```diff
@@ -555,6 +556,9 @@ GenF103.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
 GenF103.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
 GenF103.menu.upload_method.bmpMethod.upload.tool=bmp_upload
 
+GenF103.menu.upload_method.bmpMethod=HID Bootloader 2.0 (0x0800)
```
fpistm:
```suggestion
GenF103.menu.upload_method.HIDBootloaderMethod=HID Bootloader 2.0 (0x0800)
```

BennehBoy: I'll fix it shortly - it still works however... as if by magic.

BennehBoy:
@RickKimball yeah very good call, will update.

fpistm:
Tips: you can define new switch and use it in ld script. Ex: LD_SIZE_OFFSET

BennehBoy:
Think I found a simpler way:

FLASH (rx)      : ORIGIN = 0x8004000, LENGTH = LD_MAX_SIZE - 0x4000

Seems to work - ie doesn't throw linker error

BennehBoy:
I've left out the change to the upload method for now since 1) it works as is (no idea why) 2) it doesn't work when the correct naming is used.

BennehBoy:
I've attached precompiled bootloader bins, labelled with the boards they match.


[firmwares.zip](https://github.com/stm32duino/Arduino_Core_STM32/files/2784040/firmwares.zip)

edogaldo:
What's the advantage of using this bootloader vs using the DFU versions?

Thanks in advance, E.

uzi18:
- not need to shuffle with jumpers
- can jump from user code to bootloader automatically
- no use of any drivers during upload

something more?

BennehBoy:
Pretty much that.

BennehBoy:
I'll upload a different formware for zet6 later, BOOT1 is pulled down on that board and it needs to be pulled up for this BL (so as an alternate it can be compiled to use PD4 which is an SDIO pin with a pullup... won't be active when the BL is running).

uzi18:
maybe just try to read K0 or K1

BennehBoy:
After seeing the above ref, I started to play with adding maple DFU support.

I can successfully upload to a maple mini with the original 0x5000 dfu bootloader - linking works well. BUT, CDC does not function - not sure why.

Will test against Bluepill & Blackpill

BennehBoy:
@fpistm of course...

I had a quick look through the code but the LL stuff is beyond me at present...

Here's what STM32GENERIC does:
```

void USBDeviceClass::reenumerate() {
    /* Re-enumerate the USB */
      volatile unsigned int i;

    #ifdef USB_DISC_PIN
      pinMode(USB_DISC_PIN, OUTPUT);
      digitalWrite(USB_DISC_PIN, HIGH);
        for(i=0;i<512;i++);
      digitalWrite(USB_DISC_PIN, LOW);
    #else
      //pinMode(USBDP_PIN, OUTPUT);
      //digitalWrite(USBDP_PIN, LOW);
        //for(i=0;i<512;i++);
      //digitalWrite(USBDP_PIN, HIGH);


      pinMode(PA12, OUTPUT);
      digitalWrite(PA12, LOW);
      //HAL_Delay(1000);
      for(i=0;i<1512;i++){};
      pinMode(PA12, INPUT);
      //digitalWrite(PA12, HIGH);
      //HAL_Delay(1000);
      for(i=0;i<512;i++){};
    #endif
}
```
reenumerate gets called when CDC begins.

I tried 'frigging' it by issuing the same code in sketch startup, but CDC has already started at that point and things don'tr go well.

On a side note, Bluepill seems to get the BL 2.0 overwritten even though the linker & vector offset are 0x2000

BennehBoy:
OK, fixed bluepill...

So have working DFU BL 2.0 for BP 😃

Just need the DISC pin handling to get Maple Mini working

BennehBoy:
I still can't figure out why anything other than bmpMethod fails for the naming in boards.txt

BennehBoy:
I've pushed the DFU changes - Bluepill functional, Maple Mini semi functional but needs the USB_DISC_PIN change.

BennehBoy:
I think the DISC+PIN handling needs inserting here -> https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/stm32/usb/usbd_conf.c#L58

I'm struggling a bit on the LL code to initialise and toggle the pin though.

BennehBoy:
Works a charm with #410

fpistm:
I will provide a PR tomorrow based on #410 with USB DISC PIN management and reenumerate

BennehBoy:
> just in case, please add #undefine for HiGH and LOW before #endif

I removed those anyway - was a leftover from before conversion to HAL.

BennehBoy:
I just rebased all my PR's is this good or bad practice?

fpistm:
While there is no rebase conflict no.

BennehBoy:
OK, so now that the genf1 boards have been consolidated, my workaround for the upload method failing and causing a java error is broken because only the first bmpMethod is listed.

Could do with another set of eyes on this to try and figure out why the alternate method names fail even though they are defined in platform.txt

fpistm:
I will have a look when I've got a slot. Later on monday.

BennehBoy:
Although I guess this will now conflict with 

> I will have a look when I've got a slot. Later on monday.

No need, I've fixed it - needed a .protocol entry for each method.

BennehBoy:
@fpistm  is it worth me adding in native DFU like in #351?  I'm not sure which way you will want to go with bootloader support so will hold off if you say so.

fpistm:
This is open but I think I will provide soon the support of STM32CubeProgrammer.
As this tools is Multi OS and provide several interfaces (JTAG, SWD, USB, UART, SPI, CAN, I2C)
Moreover, this will allow to support new STM32  MCU and series.
The only thing that bothers me is that I could not provide it via the board manager. It will be up to the user to install it.

BennehBoy:
@fpistm This is working pretty consistently for me, I guess the only questions is around the DISC pin handling, my HAL hackjob probably isn't the best approach, happy to rebase it against your upcoming solution in https://github.com/stm32duino/Arduino_Core_STM32/pull/419

Would be nice to have some feedback from other users...

BennehBoy:
I'll fix this tonight - I think it's just due to the bluepill folder rename.

fpistm:
right.
Now that you're an expert on git, this will not be an issue :laughing:
One question is the contributing.md helps on that ?

BennehBoy:
> One question is the contributing.md helps on that ?

Yes, a lot.  Now that I understand how to set up tracking and how to rebase correctly and squash commits I think my PR's will be much neater.

However.... the previous rebase attempt has somehow ended up with me stepping on the MALYANM200_F103CB & EEXTR_F030_V1 ldscripts - I've no idea how this happened other than perhaps at some point I mistakenly dragged and dropped the bluepill ldscript into those folders.

Is there a way that I can prune any changes to those specific filed from my commits?  Or do i just go with reverting them as I did in https://github.com/stm32duino/Arduino_Core_STM32/pull/415/commits/e61a344666af240130710706a098f03b15bf86f2 ?

fpistm:
Tou can use git rebase -i then amend the commit which add this issue or simply squash the fix to this commit. Tips: During an interactive commit you can change order of the commit and so squash it.

BennehBoy:
OK, so during the rebase for the bluepill folder change, if I put those ldscripts back to how they should be, and simply skip the 2nd reversion commit all should be good...

fpistm:
right in theory :rofl:

BennehBoy:
I've locally implemented adding BL_SIZE ld preprocessing, this simplifies the linker scripts considerably, but the downside is that each non BL upload method needs BL_SIZE setting to 0 .

I tried setting this globally but was not successful, any ideas? 

boards.txt extract:
```
GenF1.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
GenF1.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
GenF1.menu.upload_method.bmpMethod.upload.bl_size=0
GenF1.menu.upload_method.bmpMethod.upload.tool=bmp_upload

GenF1.menu.upload_method.hidMethod=HID Bootloader 3.0 (0x0800)
GenF1.menu.upload_method.hidMethod.upload.protocol=hid3
GenF1.menu.upload_method.hidMethod.upload.tool=hid_upload
GenF1.menu.upload_method.hidMethod.upload.bl_size=0x800
GenF1.menu.upload_method.hidMethod.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET=0x0800 -DHIDBL_F1
```

platform.txt extract
```
# preprocess linker file
preprocld.name=preproc.ld
preprocld.filepath={build.path}/{preprocld.name}
preprocld.flags=-DLD_MAX_SIZE={upload.maximum_size} -DLD_MAX_DATA_SIZE={upload.maximum_data_size} -DBL_SIZE={upload.bl_size}
recipe.hooks.linking.prelink.1.pattern="{compiler.path}{compiler.c.cmd}" -x c -E -P {preprocld.flags} {compiler.c.flags} {build.info.flags} {compiler.c.extra_flags} {build.extra_flags} {compiler.arm.cmsis.c.flags} "{build.variant.path}/{build.ldscript}" -o "{preprocld.filepath}"

```

fpistm:
I thnik, you have to  define in platform.txt a default bl_size:
https://github.com/stm32duino/Arduino_Core_STM32/blob/cfb6fb7f77a91ea6663487522ad3dc47ce4150d3/platform.txt#L79

`upload.bl_size=0`

Then I guess this become useless:
`GenF1.menu.upload_method.*.upload.bl_size=0x800`

BennehBoy:
I placed 'upload.bl_size=0' at the top of board.txt and here's what happens....

`c:/users/bennehboy/appdata/local/arduino15/packages/stm32/tools/arm-none-eabi-gcc/6-2017-q2-update/bin/../lib/gcc/arm-none-eabi/6.3.1/../../../../arm-none-eabi/bin/ld.exe:C:\Users\BENNEH~1\AppData\Local\Temp\arduino_build_152172/preproc.ld:9: syntax error`

BennehBoy:
After re-reading your comment _correctly_ that works 👍

uzi18:
@BennehBoy @fpistm   nice idea, now it is readable and more elastic.
F407 have got some memory dedicated for Core, do we use it for something?

fpistm:
> 
> F407 have got some memory dedicated for Core, do we use it for something?

What did you mean? Could you explain your thought?

BennehBoy:
I think he's talking about the CCM memory...

It's way off topic for this thread tbh, but I guess there are probably things which could live in CCM rather than SRAM.

uzi18:
@BennehBoy exactly, do we waste this memory? Maybe we can just move stack there?

BennehBoy:
In roger's core you can select for CCM ram to be used like this:

uzi18:
nice hint, but for sure it is nice to have stack in separate mem ;)

BennehBoy:
@fpistm this PR needs reworking, the F1 version of HID has just changed design principal, it now loads into high memory, some linker script changes required.

uzi18:
@BennehBoy not so fast: https://github.com/Serasidis/STM32_HID_Bootloader/pull/7#issuecomment-460208138

BennehBoy:
@uzi18 yeah I just saw that too, thanks.

Let's see how it goes for now.  DFU at least is very useful for people with Bluepill/Maple Mini - and there are a LOT of those out there in the wild.

BennehBoy:
BTW regarding other bootloaders, these should now be super easy to add in.  If anyone wants any specific ones adding just shout or raise a PR against my repo.

I'll add a generic ROM DFU option sometime this week too.

BennehBoy:
Will rebase this later - I guess the magic word PR for USB re-enumeration is lost now that it got merged, so will keep it within this.

BennehBoy:
@fpistm any chance the Astyle check could give details of the failure - not sure what exactly is wrong.

fpistm:
@BennehBoy, you have to launch astyle.
on the root of the repo:
`python CI/astyle/astyle.py `
astyle have to be installed. You can use` -p <astyle install path>` to specify install directory if it is not in your path.

BennehBoy:
OK thanks

fpistm:
You can find astyle here
https://sourceforge.net/projects/astyle/files/astyle/astyle%203.1/

BennehBoy:
On windows there's a supplied Astyle.exe

usage format is:
`Astyle.exe --project=.astylerc <your target dir/file>`

.astylerc must be in the current dir, or you can reference the install path in the cmd line.

fpistm:
@BennehBoy  small update in order you don't think I've forgot it ;)
I will have to rewrite some part of code as I've added backup interface which could be used here.
Moreover, I would deploy STM32CubeProgrammer before and update properly tools menu and options.

BennehBoy:
OK np - I've been busy with work anyway.

BennehBoy:
Regarding STM32CubeProgrammer, the requirement to register with ST will likely discourage some people.

BennehBoy:
Perhaps a driver issue?

Here's what I can see when I look at the device in device manager:

image

BennehBoy:
dfu-util 0.9 also doesn't recognise the device :/

PS, I'm trying to get native DFU working.

BennehBoy:
OK, so this looks to be some weird USB3 issue, works fine connected to a USB2 port.

BennehBoy:
Thinking more about STM32CubeProgrammer, I can successfully upload (manual cmd line) using STM32_Programmer_CLI.exe

However, this is not in the system path - I'm also not sure if there's a 32 bit version of this tool?  So it's not immediately apparent as to how we would reference the exe?  Can it not be included in the core distribution?

fpistm:
@BennehBoy I'm currently study "all" possible options.
About 32 bits, as this is JAVA based, it should be ok.
Moreover it is mentioned in the UM:

> Supported operating systems and architectures:
•Linux® 32-bit and 64-bit (tested on Ubuntu 14.04)
•Windows® 7/8/10 32-bit and 64-bit
•macOS® (minimum version OS X® Yosemite

BennehBoy:
I was wondering whether it would be worth building some native dfu handling into the magic word checking, this way given the correct inbound code the board could be forced into the native bootloader, it would only need a small utility to be run prior to dfuu-util or stmcubeprogrammer...  I can't imagine it would be too difficult other than each board permutation possibly having a different rom address for the dfu code?  Thoughts?  It would make life a lot simpler for folks to not have to shuffle jumpers after the first upload.

fpistm:
Right. STM32CubeProgrammer has API to code our own programmer. Maybe it's possible

BennehBoy:
I got Native DFU working, but I'm not happy with the mish mash of tools required, I had to use dfu-util 0.9+.  This is fine but Roger's maple mini java wrapped uploader requires version 0.6 - I've not invested any time to see what exactly the extra bits are which that is doing to understand how it might be superceded by either the newer dfu-util or stm32cubeprogrammer.

More time required....

fpistm:
Thanks @BennehBoy for all your work.
That's why I would like introduce unified tools (STM32CubeProgrammer) to avoid mismatch btw tools accross all host OS.

uzi18:
@BennehBoy @fpistm only we not know if CubeProgrammer work as expected, maybe we need time to remove old tools.

fpistm:
I will give a try but my first test seems promising. Moreover, if I found an issue I can submit easily request to fix it😉

uzi18:
STM says they are opensource, so they should public their work of HAL/LL/Cmsis etc. in git repo so it will be more easy to PRs back to STM

BennehBoy:
Yup cubeprogrammer seemed to work quite well, what it won't be able to do is the custom 'magic word' triggering/DTR toggling required for the maple dfu bootloader, so the java that does that will need looking at - happy to do so when I get some time.

neilbalch:
Oh, how I _NEED_ this PR.

BennehBoy:
The latest CDC fixes seem to have broken the DTR toggling/re-enumeration somehow... will have to fiddle with it.

ag88:
a newer dfu-util e.g. 0.9 is a good thing, but i think accordingly there are some dependent usb-lib driver stacks on windows which may need to be updated as well
http://dfu-util.sourceforge.net/
the more recent dfu-util supports dfuse, stm's variant of dfu so it can be used to install sketches e.g. on the generic f407 black board as the f407 has a native dfu bootloader on chip. setting boot0 is necessary

fpistm:
About dfu 0.9: 
https://github.com/stm32duino/Arduino_Tools/pull/29#issuecomment-428385185

Moreover, I will deploy STM32CubeProgrammer to unified tools and allows support recent STM32. #445

ag88:
just for myself the magic word DTR toggle on f103 isn't too pressing a need, though it is good to have.
the original libmaple implementing DTR toggle as i found could conflict with using usb-cdcadm with all the functionalities i.e. the control signals, e.g. the emulated DTR, CTS/RTS, set baud rates etc.
hence when i tried to do a usb-serial dongle sketch, i did not implement the rs232 simulated control signals, e.g. i can't really set baud, stops, parity etc directly from the host, doing so would need to change the usb-serial comms code and may break the magic code DTR signalling. so i chose to set the downstream uart parameters baud, bits, stop, parity hard coding them in my usb-serial dongle sketch.

ag88:
about that 'newer' dfu-util 0.9, there are some old threads when the f4 black vet6 board is discussed
http://www.stm32duino.com/viewtopic.php?f=39&t=1976&start=10#p26203
here is how pito kind of made it work with zadig
http://www.stm32duino.com/viewtopic.php?f=39&t=1976&start=10#p26207

ag88:
thanks for the link to [445](https://github.com/stm32duino/Arduino_Core_STM32/issues/445)
STM32CubeProgrammer seem to be a good choice given the multitude of options

ag88:
oh but a main theme is programming it directly with a usb cable, no extra dongles

BennehBoy:
@ag88 yeah, the tools need an overhaul to be honest, there's a bit of sprawl at the moment.  Right now I have zero time to really exert on this and the potential move to cubeprogrammer:  I have a land rover engine stripped to the bar block that needs building back up 🤣

ag88:
hi no worries, i think cubeprogrammer, DFU, HID can *co-exist* in a sense it is kind of a variant where the start of flash and vtab address is different that's all. in terms of tools, i'd think it's alright for ST to make a decision. users can always use 'external' tools like dfu-util or hid-flash to install the sketch say on the command line separately. or some instructions in a wiki can be placed to show how to add the tools. so non issues :)

rtek1000:
Hi,

In case the user reaches this repository (usage instructions):
https://github.com/Serasidis/STM32_HID_Bootloader/issues/11

BennehBoy:
@fpistm I noticed the 1.6.0 milestone, does this mean you're refactoring with stmcubeprogrammer?

fpistm:
> @fpistm I noticed the 1.6.0 milestone, does this mean you're refactoring with stmcubeprogrammer?

I was wondering how long you would take to ask 😄 
So, yes, I will review and test it after rebase/refactor it properly.

BennehBoy:
🤣

uzi18:
what about hid bootloader/programmer? it is not supported by stmcubeprogrammer.

fpistm:
I will add it

fpistm:
@BennehBoy, @Serasidis
I'm wondering if we should not keep only HID bootloader?
Is Maple DFU Bootloader 2.0 and Maple DFU Bootloader original (0x5000) really needed?
What is your thought about that?

BennehBoy:
Bluepill's & Maple Mini's come preloaded with Maple DFU BL 2.0 & Maple
Original BL respectively, there's no SWD on the Maple Mini, and BL2
disables SWD on the Bluepill.  This means that by not including support for
these bootloaders we would effectively orphan a lot of potential new users.

fpistm:
Right, forgot this use case.
Thanks for the quick answer ;)

fpistm:
@BennehBoy 

>     * F1 boards must use the latest 2K HID bootloader -> https://github.com/Serasidis/STM32_HID_Bootloader  (hid_flash tool will warn if an older version is installed)

I'm little bit lost with that. I know some users works on this but I think it was not the case here as it is not compatible with EEPROM emulation.
And reffering to your updated ld the BL is a the start of the flash.

BennehBoy:
Can you explain further?

When I was creating the PR the HID BL author updated his repo so that the tools would refuse to flash devices with an earlier version of the BL installed....

Then some more commits were made to move the F1 version of the BL to the end of flash, these were then backed out.

I've noticed some recent commits but have not had my eye close enough on it to see if the end of flash has been re-introduced (I doubt it would be due to some other issues).

fpistm:
Ok so your limitation on your OP is not correct.
Currently, the bootloader is flashed at the beginning?

BennehBoy:
The limitation is correct with the hid tool that I supplied in the tools repo.

fpistm:
Oh ok. Sorry, i've read that it was flash on the latest sector :rofl: 
So, I've clen up your PR (split commit, clean code, enhance boards.txt, add switch to enable it only for compatible series, use backup API, ...)
See #525

neilbalch:
Out of curiosity, could one of you please explain what is currently hindering this PR from being merged in relatively lay-person parlance? From what this looks to be, It would be of great use to me to have this feature in the STM32duino core.

fpistm:
> Out of curiosity, could one of you please explain what is currently hindering this PR from being merged in relatively lay-person parlance? From what this looks to be, It would be of great use to me to have this feature in the STM32duino core.

@neilbalch 
This PR is now superseded by #525 
This will be merged soon.
@BennehBoy made a great job on this PR. It only required some time to review, test and merge.

BennehBoy:
@fpistm I figure you've guessed my time has been limited lately - the original reason I got involved with microcontrollers, my car, has stolen most of my recent time...  Engine rebuilds, gearbox work etc.  Never buy a Land Rover!

I'll try and get some testing of PR #525 done soon.

Serasidis:
A few days ago I started working on adapting the HID Bootloader (HID-BL) to the Official STM Core. I know that @BennehBoy works also on that, but I didn't see the source code (any link is welcome). 
So, I started from scratch adding the HID-BL to this Core.

There is a new feature on Arduino-IDE 1.8.9 the _**Pluggable Discovery**_ that I am working on it.
This feature adds an extra Port menu that includes HID devices in addition to the available serial ports.
This is very useful in case someone want to reboot the STM32 to the bootloader but the onboard USB is set as HID device or any other USB device such as MIDI.

PR #416, opened 2019-01-25
New Variant - HY-TinySTM103T

BennehBoy:
Adding support for HY-Tiny103T board -> http://wiki.stm32duino.com/index.php?title=HY-TinySTM103T

I've added this to the Generic 103 section of boards.txt which means it inherits the 64/128KB flash selection, either selection will result in the variant being correctly linked for 128KB, but the Arduino IDE will incorrectly report max ram/flash sizes during build if 64K is selected.

No local testing done...

Could do with Arduino A0/PA0 equivalence testing & the usual checks.


BennehBoy:
I've purchased a couple of boards just out of interest.

fpistm:
About Flash menu, it could be removed and instead use the ld preprocess I guess

BennehBoy:
> About Flash menu, it could be removed and instead use the ld preprocess I guess

Wouldn't that mean multiple choices in the board entry though?  How else would it know which size had been selected?

BennehBoy:
OK, well I'll leave that bit to you 🤣

fpistm:
Done in #418 ;)

variants/HY_TinySTM103T/variant.h
```diff
@@ -0,0 +1,136 @@
...
+#define PA15 10
+// Left side
+#define PA3  11
+#define PA4  12
+#define PA5  13
+#define PA6  14
+#define PA7  15
+#define PB0  16
+#define PB1  17
+#define PB2 //BOOT1
```
fpistm: There is a mistake here
As PB2 is defined in  `digitalPin[]` at index 18, there is a shift here.

variants/HY_TinySTM103T/variant.h
```diff
@@ -0,0 +1,136 @@
...
+#define PA8  18
+#define PA9  19
+#define PA10 20
+#define PA11 21
+#define PA12 22
+
+// This must be a literal
+#define NUM_DIGITAL_PINS        33
```
fpistm: So after manage PB2 definition, this one need to be checked.

variants/HY_TinySTM103T/variant.h
```diff
@@ -0,0 +1,136 @@
...
+// This must be a literal with a value less than or equal to to MAX_ANALOG_INPUTS
+#define NUM_ANALOG_INPUTS       9
+#define NUM_ANALOG_FIRST        24
```
fpistm: So after manage PB2 definition, this one need to be checked.

variants/HY_TinySTM103T/variant.cpp
```diff
@@ -0,0 +1,162 @@
...
+// Left side
+  //3V3
+  PA_3,
+  PA_4,
+  PA_5,
+  PA_6,
+  PA_7,
+  PB_0,
+  PB_1,
+  PB_2, //BOOT1
```
fpistm: PB2 definition need to be reviewed as in this array but index 18 in variant.h  is defined as PA8

fpistm:
Moreover, we will wait that some tests can be performed.

BennehBoy:
No problem I'll fix the pin map tonight.

I've asked the requester to test - he may have issues until I commit later :rofl:

variants/HY_TinySTM103T/variant.cpp
BennehBoy: OK I know what I did here,  skipped BOOT1 because of the comment, have fixed and checked the Dig/An pin counts.

variants/HY_TinySTM103T/variant.h
fpistm: So this should be 34 now ;)

fpistm:
@BennehBoy Still one issue if I'm not wrong ;)
As an extra you can try to rebase it properly as an exercise :stuck_out_tongue_winking_eye:

variants/HY_TinySTM103T/variant.h
BennehBoy: Yeah, sorry, was shattered when I did the update last night.

BennehBoy: Actually the count is correct, I had just numbered incorrectly...

BennehBoy:
I __think__ I've sorted this now....

variants/HY_TinySTM103T/variant.h
fpistm: Hum no, there is 34 pins in `digitalPin[]`.
as it is 0 index base.
23+1(for 0 index) + 10 analog.
Else I'm very tired. 🙄

fpistm:
Congratulations 👍 PR is cleaned.

variants/HY_TinySTM103T/variant.h
BennehBoy: No, you're right, I'm having a senior moment 👴

fpistm:
Even if not tested, I've merged it. If any issue is found then we will correct it.

BennehBoy:
My boards are coming airmail from Singapore so I should be able to test this soon if the original requester does not respond.

mitchmitchell:
I'm going to try this out on one of my test boards today and tomorrow -- if all goes well I'll let it soak running the mysensors application to see how it performs (found a problem in the SPI.cpp file in Roger's version).

PR #418, opened 2019-01-25
Update for Generic F1

fpistm:
See commit messages for details.
/CC @BennehBoy


BennehBoy:
OK got it, so is WoW to wait for this to be merged then refactor my open PR's to use this new format?

fpistm:
I think this will be good as this fix issue about the common flash menu which originally been added for BP.
Are you agreed with this PR?

PR #420, opened 2019-01-26
New Variant BluePill F103C6 (32K)

BennehBoy:
It looks like some manufacturers in China are cutting corners and re-laser etching F103C6's as F103C8's.

fredbox on the stm32duino forum has been on the receiving end of this and suggested this addition -> http://www.stm32duino.com/viewtopic.php?p=52804#p52804

These chips have only 32K flash & 10K SRAM, but should be useful for many lower end tasks.


fpistm:
This will require this 
https://github.com/BennehBoy/Arduino_Core_STM32/pull/1

BennehBoy:
OK, makes sense, I hadn't realised there were less peripherals - my bad.

fpistm:
No worries ;)

PR #424, opened 2019-02-01
Generic Serial

fpistm:
This PR allows to have `Serial `mapped on `SerialUSB `if USB CDC is enabled.

If USB CDC is enabled then the generic "`Serial`" used will be `SerialUSB`
else this is the `HardwareSerial `if one is defined.
    
 If `Serial `is defined before (`build_opt.h`, `variant.h`)
then this is the one used.
    
 ISO for generic "`serialEvent`".


BennehBoy:
Does this affect the naming sequence of the other Serial instances?

Ie will Serial3 become Serial 2 if Serial is remapped to USB?

fpistm:
@BennehBoy, no
Default instance naming is:
 * `SerialX `== U(S)ARTX
 * `SerialUSB `== CDC

`Serial `name is a defined to have a generic way to access default _Serial_.

fpistm:
I've updated/reword the menu to have better description and usage.
Let me know what you think.

BennehBoy:
Looks good

PR #419, opened 2019-01-25
USB enhancement

fpistm:
List:
 * [x] #410 merged (include some typos and warnings fixes)
 * [x] HID composite clean up 
 * [x] `USB_DISC_PIN` and re-enumeration --> Covered by #426
 * [x] Serial and Serial Event management --> Covered by #424
 * [x] `readBytes` and `readBytesUntil` thanks @makarenya (#410) 

Current Status:
FS well functional.
HS ok except ReadByte test.

PR #426, opened 2019-02-01
[USBD] Force re-enumeration

fpistm:
Enumeration delay could be redefined using `USBD_ENUM_DELAY`
To disable re-enumeration define `USBD_REENUM_DISABLED`
Use `USB_DISC_PIN` if defined.

Tested with BluePill, MapleMini and Disco F746.

Any comments are welcome.

### DIFF
--- a/cores/arduino/stm32/usb/usbd_conf.c
+++ b/cores/arduino/stm32/usb/usbd_conf.c
@@ -19,6 +19,7 @@
 #ifdef USBCON
 /* Includes ------------------------------------------------------------------*/
 #include "usbd_core.h"
+#include "usbd_if.h"
 #include "stm32yyxx_ll_pwr.h"
 
 #ifndef HAL_PCD_MODULE_ENABLED
@@ -421,6 +422,7 @@ void USBWakeUp_IRQHandler(void)
   */
 USBD_StatusTypeDef USBD_LL_Init(USBD_HandleTypeDef *pdev)
 {
+  USBD_reenumerate();
   /* Set common LL Driver parameters */
   g_hpcd.Init.dev_endpoints = 4;
   g_hpcd.Init.ep0_mps = DEP0CTL_MPS_64;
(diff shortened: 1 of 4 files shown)


BennehBoy:
@fpistm not sure if this is directly related to this PR, but on blackpill I've noticed that CDC enumeration doesn't occur unless the user sketch initialises the serial port, ie SerialUSB.begin

This means that bootloader tools such as hid_flash.exe cannot see the device without it being forced into perpetual bootloader.

I'll do some more testing later minus my bootloader PR to esnure I've not confused this somehow.

fpistm:
Sorry, for the delay, I'm sick so will not be very present this week.
Righ the reenumerate is done when the Serial.begin() is performed.
You can try to add it in the init after the hw_config call:
https://github.com/stm32duino/Arduino_Core_STM32/blob/cfb6fb7f77a91ea6663487522ad3dc47ce4150d3/cores/arduino/board.c#L11
or in the hw_config:
https://github.com/stm32duino/Arduino_Core_STM32/blob/4e1c22561ae15c976a42699c97e08c53dbb3f4b7/cores/arduino/stm32/hw_config.c#L57

do not forget to include: usbd_if.h

BennehBoy:
OK I'll give that a try, get well soon!

BennehBoy:
Haven't tested this yet @fpistm I've been a little bit distracted - will try do so today.

fpistm:
What I guess is that not only the re-enum should be done but also the CDC init to catch or not the magic number right ?

BennehBoy:
Yes

fpistm:
@BennehBoy 
Please, Could you test with this branch:
https://github.com/fpistm/Arduino_Core_STM32/tree/USB_reenum_Test

BennehBoy:
@fpistm sorry, had a very busy day yesterday - I should be able to test the above this evening.

fpistm:
I've merge this one as independent of issue met with #419.
This will ease test and fix.

BennehBoy:
Do I still need to make the test? we can submit a fix via PR if needed right?

fpistm:
I'm currently do an update of the last PR with my code to test.

BennehBoy:
I guess 1.5.0 is drawing close then :+1:

fpistm:
I hope. I really like release it. Anyway @ktand has a last blocking issue.

PR #444, opened 2019-02-18
USBSerial would not block on write if the USB inteface is not connected

ghent360:


fpistm:
Hi @ghent360 
Thanks for the PR.
I guess it solve issue with Marlin FW ?

ghent360:
Yep.

ktand:
Hi @ghent360

Will it handle the situation where the USB is first connected and then the
cable is disconnected without locking up? If I remember I had to check
lineState to prevent a lock up with the timer based CDC.

Karl

ghent360:
Let me write a little test to double check.

ghent360:
So I tried this:

void setup() {
  SerialUSB.begin(115200);
  Serial1.begin(115200);
}

void loop() {
  for(int idx = 0; true; idx++) {
    SerialUSB.println(idx);
    Serial1.println(idx);
  }
}

The Serial1 continues to print when the USB is disconnected. The USB
resumes after it gets re-connected.
https://photos.app.goo.gl/hiP9ZAKGDzBkLmBMA

ghent360:
I see where the problem is: this only handles the case where the cable is
yanked, but when connected there is something that reads the data on the PC.

If I close the terminal on the PC the HardwareSerial still stops printing.
Karl, what is it that you had to check?

~V

uzi18:
Is there a way to check if usb/cdc is connected?

ghent360:
Yeah, this is a complex issue. I added checking the DTR line value, which gets set when the terminal is open and reset when the terminal is closed.

That makes it work almost all the time. Now the edge case is when the terminal is open, but the program on the PC does not actually read the serial data, then it locks again.

In my Ubuntu I use minicom for serial port communication, when I try to exit, there is a confirmation dialog "Are you sure you want to exit", if I stay there long enough the STM32 lock again. If I exit quickly there is no issue, the DTL line is reset and the code stops sending data, but if I linger on the exit dialog, apparently minicom stops reading data from the virtual terminal, but the terminal port is not closed.

What a bummer :-(

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -37,6 +37,7 @@
 USBD_HandleTypeDef hUSBD_Device_CDC;
 
 static bool CDC_initialized = false;
+static bool hostPortOpen = false;
```
fpistm: I wonder if it is really need to define a new variable.

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -169,6 +170,7 @@ static int8_t USBD_CDC_Control(uint8_t cmd, uint8_t *pbuf, uint16_t length)
     case CDC_SET_CONTROL_LINE_STATE:
       lineState =
         (((USBD_SetupReqTypedef *)pbuf)->wValue & 0x01) != 0; // Check DTR state
+    hostPortOpen = lineState;
```
fpistm: Anyway, there is an indent issue:
```suggestion
      hostPortOpen = lineState;
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -245,6 +247,11 @@ void CDC_deInit(void)
   }
 }
 
+uint8_t CDC_connected()
+{
+  return hUSBD_Device_CDC.dev_state == USBD_STATE_CONFIGURED && hostPortOpen;
```
fpistm: I will use directly `lineState`here:
```suggestion
  return hUSBD_Device_CDC.dev_state == USBD_STATE_CONFIGURED && lineState;
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.h
```diff
@@ -50,6 +50,7 @@ void CDC_continue_transmit(void);
 void CDC_resume_receive(void);
 void CDC_init(void);
 void CDC_deInit(void);
+uint8_t CDC_connected();
```
fpistm: Maybe use a `bool `return should be better.
```suggestion
bool CDC_connected();
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -245,6 +247,11 @@ void CDC_deInit(void)
   }
 }
 
+uint8_t CDC_connected()
```
fpistm:
```suggestion
bool CDC_connected()
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
ghent360: as written, no it is the same as the value lineState
I was reading this thread:
https://stackoverflow.com/questions/5338875/detecting-open-pc-com-port-from-usb-virtual-com-port-device

and there it was suggested to trigger the hostPortOpen on SET_LINE_CODING, but it was not clear what would reset it, so I left it out.

cores/arduino/stm32/usb/cdc/usbd_cdc_if.h
ghent360: Sorry, I'm going to revert your suggestion. AFAIK bool is a C++ type and this is C file.
_Bool is available in C99 compilers and allegedly there is a macro defining bool as _Bool, but not in older C89/C90.

fpistm: bool is used in several other c files (including usbd_cdc_if.c) so this will not be an issue.

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
fpistm: As documented on many site:
> On many operating systems, including Windows, the DTR line is held low while the serial port is unused and not being controlled by any applications. 

What I wonder if the `bool ` operator should be updated to have the `while(!Serial)` functional.

ghent360: I was thinking that the bool() was intended to return isConnected().

fpistm: Right forgot isConnected ;)

ghent360:
This last variant works even if I linger in the minicom menus.
I added a simple timeout if our transmit packet drags for too long I would assume the host is not active.

ghent360:
If people can test this change with different boards and host environments. Look for comment #6 for an example test sketch.
You would need a board with physical serial port as well as USB connector.

cores/arduino/USBSerial.cpp
```diff
@@ -185,6 +187,11 @@ bool USBSerial::rts(void)
   return false;
 }
 
+bool USBSerial::isConnected()
```
fpistm: Sorry, could you remove it as in fact deprecated. bool operator is the right way.

ghent360:
I'm going on vacation tomorrow. I'll continue working on this PR when I come back at the end of the week.
Thank you all for the feedback.

ktand:
@ghent360 

I've tried the changes in this PR and this resolves the issue I had with Marlin where the firmware would hang until the serial port is opened on the host side.

On a side note, do you think there is any way that we could have a "serial disconnected" event that we could use within marlin to reset the board if the CDC port is closed/disconnected?

Thanks!

ghent360:
I was thinking about that for a bit, but  it is not elegant approach,
because when the board is restarted the usb device is
disconnected/reconnected and it can result in a circular board reset
behavior:
Board starts;
USB is connected;
Port is open;
Board is reset;
Go to 1

roc2:
why the serial assistant on PC can't receive data with follow sketch:
`#include <Arduino.h>
#include "USBSerial.h"

void setup()
{
    SerialUSB.begin();
    pinMode(LED_BUILTIN, OUTPUT);
}

void loop()
{
//    if (SerialUSB.available() > 0)
    {
       SerialUSB.println("testing read and wirte");
    }
    digitalToggle(LED_BUILTIN);
    delay(300);
}`

Only when I add this line "if (SerialUSB.available() > 0)", then send some data with serial assistant on PC, the PC can receive data continusly.

fpistm:
@roc2 this is not clear. Which board? Core version (1.5.0, master,...)? With this PR applied or not?....

roc2:
@fpistm My board is Black F407VE, core version 1.5.0.

fpistm:
As it is done in Arduino for the Due. 
https://www.arduino.cc/reference/en/language/functions/communication/serial/ifserial/
Try to add in the setup after `SerialUSB.begin();`
```
  while (!SerialUSB) {
    ; // wait for serial port to connect. Needed for native USB
  }
```

roc2:
@fpistm Add this 'while (!SerialUSB)', it's no working. I have tested with FS and HS mode, the code work fine with FS mode(select the USB speed Low/full speed), but it is not work well with HS mode(I added the USB3300 module on the ULPI interface)

ghent360:
The code works for me with the PR applied. The only trick is when you have
the arduino serial monitor connected and the board is reset the port name
is changed, so I had to select the new port from the tools menu.

ghent360:
while(!SerialUSB) delay(10);
works as well it waits until the serial monitor is open.

fpistm:
I've tested your original sketch on 1.5.0 and it works fine with BLACK F407. About HS, as I've no 
USB3300, this will be difficult to help.

roc2:
@fpistm  @ghent360  Thanks for your help. As you said, when I test with Arduino serial monitor, it's work well with "
while(!SerialUSB)"; But when test with other Serial Assistant on PC, it can't work. I think mabe the Arduino serial monitor have done something when opened.

fpistm:
Testing with putty and it works. Maybe this PR handling the DTR will allow to work with your term

roc2:
@fpistm  Yes, the Putty works fine, thank you . I have just tested another serial assistant.

roc2:
@fpistm There should be correct order with USBSerial, the serial monitor should be opend and connected before sending data, thus everything works well. Thanks for your help .

ghent360:
@fpistm Are there any more changes required for this PR?
I believe I addressed add requests, but the thing still shows red.

fpistm:
> 
> 
> @fpistm Are there any more changes required for this PR?
> I believe I addressed add requests, but the thing still shows red.

Currently not. I only need to review it and test deeply.

cores/arduino/USBSerial.cpp
```diff
@@ -83,6 +83,8 @@ size_t USBSerial::write(const uint8_t *buffer, size_t size)
       buffer += portion;
       // After storing data, start transmitting process
       CDC_continue_transmit();
+    } else if (!CDC_connected()) {
+      return size - rest;
```
fpistm: Should be better to use break in this case to exit while then `return size -rest`.
```suggestion
      break;
```

cores/arduino/USBSerial.cpp
```diff
@@ -83,6 +83,8 @@ size_t USBSerial::write(const uint8_t *buffer, size_t size)
...
     }
   }
   return size;
```
fpistm: Here
```suggestion
  return size - rest;
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -245,6 +249,17 @@ void CDC_deInit(void)
...
 }
 
+bool CDC_connected()
+{
+  uint32_t transmitTime = 0;
+  if (transmitStart) {
+    transmitTime = HAL_GetTick() - transmitStart;
+  }
+  return hUSBD_Device_CDC.dev_state == USBD_STATE_CONFIGURED
+         && transmitTime < USB_CDC_TRANSMIT_TIMEOUT
```
fpistm: USB_CDC_TRANSMIT_TIMEOUT is 3.
Is it not too low on slower board ?

Maybe ti could be fine to be able to redefine it if needed
```suggestion
#ifndef
#define USB_CDC_TRANSMIT_TIMEOUT 3 
#endif
```

fpistm:
@ghent360 I've made several test on several boards with different STM32 families and host OS. 
It's ok. 
Still my comments to discuss.

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
ghent360: The value depends on how fast/slow the host reads from the CDC endpoint. I added some comments and the #ifndef suggestion.

cores/arduino/USBSerial.cpp
ghent360: done

cores/arduino/USBSerial.cpp
ghent360: done

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -32,6 +32,15 @@
 #define CDC_MAX_PACKET_SIZE USB_MAX_EP0_SIZE
 #endif
 
+/*
+ * The value USB_CDC_TRANSMIT_TIMEOUT is defined in terms of HAL_GetTick() units.
+ * Typically it is 1ms value. The timeout determines when we would consider the
+ * host "too slow" and threat the USB CDC port as disconnected.
+ */
+#ifndef
+#define USB_CDC_TRANSMIT_TIMEOUT 3
```
fpistm: Astyle failed due to space at the end
```suggestion
#define USB_CDC_TRANSMIT_TIMEOUT 3
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -32,6 +32,15 @@
 #define CDC_MAX_PACKET_SIZE USB_MAX_EP0_SIZE
 #endif
 
+/*
+ * The value USB_CDC_TRANSMIT_TIMEOUT is defined in terms of HAL_GetTick() units.
+ * Typically it is 1ms value. The timeout determines when we would consider the
+ * host "too slow" and threat the USB CDC port as disconnected.
+ */
+#ifndef
```
fpistm:
```suggestion
#ifndef USB_CDC_TRANSMIT_TIMEOUT 
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
```diff
@@ -212,6 +222,7 @@ static int8_t USBD_CDC_Receive(uint8_t *Buf, uint32_t *Len)
 
 static int8_t USBD_CDC_Transferred(void)
 {
+  transmitStart = 0;
```
fpistm: Hum, I guess it should be 
```suggestion
  transmitStart = HAL_GetTick();
```

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
ghent360: fixed

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
ghent360: fixed.

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
ghent360: There was a line missing from the commit quashing. I added it.

ghent360: the line is correct, this call indicates end of a transfer so we stop the timer.
the timer would be reinitialized if there is something to transmit inside CDC_continue_transmit

fpistm: Ah right👍

cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
fpistm: Still not fixed in the updated PR

ghent360: how about now

fpistm: Thanks

fpistm:
Thanks @ghent360

fpistm:
Using your sketch https://github.com/stm32duino/Arduino_Core_STM32/pull/444#issuecomment-464629897
Works fine under Linux. 
Under Windows, Unable to recognized the device after unplug/plug. Enum failed.
Anyway this is better than before :)

cores/arduino/USBSerial.cpp
```diff
@@ -83,9 +83,11 @@ size_t USBSerial::write(const uint8_t *buffer, size_t size)
       buffer += portion;
       // After storing data, start transmitting process
       CDC_continue_transmit();
+    } else if (!CDC_connected()) {
```
fpistm: Doing further investigation.
I think it should be better to put the "CDC_connected() " in the while condition
` while (CDC_connected() && (rest > 0)) {`

fpistm:
>  Under Windows, Unable to recognized the device after unplug/plug. Enum failed.
Update:
Doing further investigations, if I didn't open a terminal on the COM port. Enumeration is correct each time.

ghent360:
Does my change affect the enumeration?

fpistm:
> Does my change affect the enumeration?

No @ghent360 , this is better with this PR.
Seems more like a driver issue but only after the COM has been used by Windows host (with a term).

cores/arduino/USBSerial.cpp
fpistm: @ghent360 did you see this comment?

ghent360: When you move it to the while. When I disconnect and re-connect the USB the USB serial sometimes stops transmitting. I believe it goes in perpetual timeout state, because we stop sending data and the timeout is never reset.

ghent360: I added core to reset the transmit timeout when the port is re-oppened.

fpistm: Make sense, thanks.

PR #470, opened 2019-03-12
[Superseded by #541] twi: enable ack'ing general calls

benwaffle:
In I2C, sending data to address 0 sends it to all devices on the bus. This is called a general call. This setting controls whether the slave device ACKs general calls or not.

I'm not sure if this will break existing code.

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -193,7 +193,7 @@ void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, u
   handle->Init.OwnAddress2     = 0xFF;
   handle->Init.AddressingMode  = addressingMode;
   handle->Init.DualAddressMode = I2C_DUALADDRESS_DISABLE;
-  handle->Init.GeneralCallMode = I2C_GENERALCALL_DISABLE;
+  handle->Init.GeneralCallMode = I2C_GENERALCALL_ENABLE;
   handle->Init.NoStretchMode   = I2C_NOSTRETCH_DISABLE;
 
   handle->State = HAL_I2C_STATE_RESET;


cores/arduino/stm32/twi.c
```diff
@@ -193,7 +193,7 @@ void i2c_custom_init(i2c_t *obj, i2c_timing_e timing, uint32_t addressingMode, u
   handle->Init.OwnAddress2     = 0xFF;
   handle->Init.AddressingMode  = addressingMode;
   handle->Init.DualAddressMode = I2C_DUALADDRESS_DISABLE;
-  handle->Init.GeneralCallMode = I2C_GENERALCALL_DISABLE;
+  handle->Init.GeneralCallMode = I2C_GENERALCALL_ENABLE;
```
LMESTM: Does it make sense as well if the STM32 is master ? or shouldn't we enable this feature only if master == false ?

benwaffle: I think this always should be enabled, because in multimaster mode, the masters can also act as slaves. I think this needs to be tested with 1 device doing lots of general calls and another device in master mode, and making sure that every general call is received.

LMESTM: Ok. but as you mentioned this requires a good testing session. An alternative would be to provide an option in the constructor to let application decide whether to answer General Calls or not.  up to @fpistm

fpistm: Reading some docs about general call, it's not so easy to tell that it should be enabled by default.
What I suggest is to provide a way to enable it if user really want to use it. This way, there is no regression risks.

fpistm:
This PR is superseded by #541.

fpistm:
#541 merged

PR #475, opened 2019-03-19
Allow to change default PWM frequency

fpistm:
Default is `PWM_FREQUENCY` (1000Hz).
Call `analogWriteFrequency(freq)` before `analogWrite(pin, val)`
to use the new frequency.

Note that no compute/check is performed to ensure the requested freq.
This is dependent on how the computation is currently done in analogWrite.
This limitation will be remove when Timer management will be done.

Should replace #349


romainreignier:
I was about to add this API to the wiki but you have already done it. Nice :+1:

PR #480, opened 2019-03-25
Fix issue #478.

### LINKED ISSUE #478: USBSerial loosing characters on Linux
**Describe the bug**
I'm using the Marlin 3d-printer firmware on the Armed 3D-printer board (STM32F407) together with Octoprint running on a Raspberry PI 3 A+. Octoprint request temperature status to be sent from the printer to the Octoprint host on a reqular basis (every 2nd second). A temperature status message looks like this:

` T:23.01 /0.00 B:22.88 /0.00 @:0 B@:0`

The problem is that sometimes one or more characters of this message is lost. I have tested without Octoprint, using minicom, and the character loss is there as well so this is not a problem with Octoprint.

```
 T:96.18 /0.00 B:61.56 /0.00 @:0 B@:0
 T:95.66 /0.00 B:61.47 /0.00 @:0 B@:0
 T:90.86 /0.00 B:60.41 /0.00 @:0 B@:0
 T:90.41 /0.00 B:60.34 /0.00 @:0 B@:0
 T:89.93 /0.00 :60.26 /0.00 @:0 B@:0 <-- character lost
 T:88.13 /0.00 B:.85 /0.00 @:0 B@:0 <-- characters lost
 T:87.70 /0.00 B:59.71 /0.00 @:0 B@:0
 T:87.20 /0.00 B:59.59 /0.00 @:0 B@:0
 T:85.52 /0.00 :59.22 /0.00 @:0 B@:0 <-- character lost
 :96.66 /0.00 B:61.65 /0.00 @:0 B@:0 <-- character lost
 T:85.10 /0.00 B:.88 /0.00 @:0 B@:0 <-- characters lost
 T:84.63 /0.00 B:58.76 /0.00 @:0 B@:0
```

The problem exists in the **master** branch but not in **release 1.5** (I'm currently testing with master @ commit 7feeb092bcaeed4d01f6de0e35b1666b2c4cefd3)

The message is composed by calls to both string and single character Stream functions 

_What is strange is that on Windows no characters are lost._

**To Reproduce**
I have tried, but failed, to reproduce this in a test program by sending the message above every 2nd second.

Could this be a timing issue? Marlin uses multiple timer interrupts.

I understand this is not easy to investigate but I'll be happy to help test if someone got an idea what could be causing this.

 - Name: Armed

ghent360:

* [ #478 ] Bug USBSerial loosing characters on Linux

In CDC_connected the core reads the value of the global variable transmitStart twice. This is a problem because the variable is modified inside ISR and the value can change between the two reads. It can lead to calculating the wrong value for transmitTime which can lead to loss of data.

This proposed fix would save the value of transmitStart in a local variable, so we read the value only once.

Fixes #478

### DIFF
--- a/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
+++ b/cores/arduino/stm32/usb/cdc/usbd_cdc_if.c
@@ -263,8 +263,9 @@ void CDC_deInit(void)
 
 bool CDC_connected()
 {
-  uint32_t transmitTime = 0;
-  if (transmitStart) {
+  // Save the transmitStart value in a local variable - fix #478
+  uint32_t transmitTime = transmitStart;
+  if (transmitTime) {
     transmitTime = HAL_GetTick() - transmitStart;
   }
   return hUSBD_Device_CDC.dev_state == USBD_STATE_CONFIGURED


ktand:
@ghent360 @fpistm The fix is incorrect.

The line:

`transmitTime = HAL_GetTick() - transmitStart;`

should be

`transmitTime = HAL_GetTick() - transmitTime;`

otherwise you'll be reading `transmitStart` twice, as before.

ghent360:
@ktand You are absolutely correct.

ghent360:
I'm the dunce of the week. Can't make a 3 line code fix right.

fpistm:
Thanks @ktand 
Now, fixed in master.
I think I read too much code 🙄

PR #500, opened 2019-04-16
Support custom Usb reenumerate

xC0000005:
**Summary**

* [ ] Supports USB custom reenumerate. The Malyan M200 V1 board has a pull up circuit that renders the DP pulldown method unreliable. 

making the renumerate function weak allows a variant to override it

**Validation**

Tested on Malyan M200 V1 rev 1, 3, and 4 (STM32F103) and on Malyan M200 V2 (STM32F070) to make sure it didn't affect others.

### DIFF
--- a/cores/arduino/stm32/usb/usbd_if.c
+++ b/cores/arduino/stm32/usb/usbd_if.c
@@ -16,7 +16,7 @@
   * @retval None
   */
 
-__weak void USBD_reenumerate(void)
+WEAK void USBD_reenumerate(void)
 {
 #ifndef USBD_REENUM_DISABLED
   /* Re-enumerate the USB */
--- a/cores/arduino/stm32/usb/usbd_if.c
+++ b/cores/arduino/stm32/usb/usbd_if.c
@@ -16,7 +16,7 @@
   * @retval None
   */
 
-void USBD_reenumerate(void)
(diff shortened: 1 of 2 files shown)


cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -16,7 +16,7 @@
   * @retval None
   */
 
-void USBD_reenumerate(void)
+__weak void USBD_reenumerate(void)
```
fpistm:
```suggestion
WEAK void USBD_reenumerate(void)
```

variants/MALYANM200_F103CB/variant.cpp
```diff
@@ -95,6 +95,19 @@ void initVariant()
   digitalWrite(PB9, 1);
 }
 
+extern "C"
```
fpistm: Already in extern "C" block.
```suggestion
```

xC0000005:
I'll update it once I've done that, thanks.

variants/MALYANM200_F103CB/variant.cpp
xC0000005: I'm working on fixing that.

fpistm: Is there an issue to simply remove it ?

Issue #494, opened 2019-04-05
Hardware serial port RX locks up after overrun error.


benwaffle:
**Describe the bug**
When you run the following code, and start spamming words (followed by the enter key) on your keyboard, eventually it stops reading data. You can also lock it up by just hitting enter without any letters. The USART1 interrupt stops firing, because the code seems to fail to set TXEIE. It also appears to only happen when there's an overrun error as verified by looking at USART_SR with gdb.

**To Reproduce**
```cpp
#include <Arduino.h>

void setup() {
    Serial.begin(9600);
    Serial.println("hello");
}

void loop() {
    if (Serial.available() > 0) {
        printf("Serial available: %d\n", Serial.available());
        char buf[128] = {0};
        char *line = buf;
        size_t nbytes;
        if ((nbytes = Serial.readBytesUntil('\n', buf, sizeof buf)) == 0) {
            printf("# error: no data read from serial\n");
            return;
        }

        printf("# read %d [%s]\n", nbytes, line);
        printf("rest = %d\n", Serial.available());
    }
}
```

1. Run code
2. Open a serial monitor (e.g. `pio device monitor`)
3. Spam stuff into the serial console. (Or just hit enter)
4. See it lock up and refuse to take more input.

**Expected behavior**
Continue receiving input, possibly by dropping the next character.

**Desktop (please complete the following information):**
 - OS: Mac OS 10.14.3
 - Platformio: 3.6.6
 - STM32 core version: 1.5.0
 - Upload method: STLink

**Board (please complete the following information):**
 - Name: Bluepill 128k

benwaffle:
looks similar to https://github.com/micropython/micropython/issues/3375

fpistm:
Hi @benwaffle 
Seems similar to #467.

benwaffle:
I tried the 2 patches from ppescher but that did not fix it

fpistm:
Ok. Thanks for the test.

uzi18:
sizeof(buf)

benwaffle:
@uzi18 The code compiles just fine. You only need the parenthesis for types, not variables.

ppescher:
Hi @benwaffle 
It may not work for you because you are using `printf()`. Try to only use `Serial.print()/println()`.

benwaffle:
I will try that, but `printf()` uses `uart_debug_write()` with uses the synchronous `HAL_UART_Transmit()`, without interrupts.

[...]

PR #496, opened 2019-04-09
Disable USART IRQ in uart_write and uart_debug_write

benwaffle:
This PR fixes issues #494 and #467 for printf();

I want to be able to use printf() with Serial.write() and Serial.read(). See discussion in both issues mentioned above.

**Validation**

See the example code in #494

---

I'm not sure how to:

- Get IRQn_Type values from a UART_TypeDef
- Determine which UARTs are defined for a given device 
- Whether I need to set the priority when enabling the interrupt

Issue #467, opened 2019-03-07
Hardware serial ports issue with full-duplex usage


ppescher:
The way serial ports are implemented in the Arduino core can lead to a serial port permanently stop receiving under particular conditions.

If the serial port is transmitting some data, the last transmit callback does not re-enable the TX interrupt if the ring buffer is empty. This is done at the end of the HardwareSerial::write() function, outside the IRQ context.

If a receive interrupt fires at this time, inside the Transmit_IT() function, when the receive callback is finally called it will fail to re-enable the RX interrupt, because Receive_IT() finds the handle locked.

You can see my solution in this commit: https://github.com/RoboTech-srl/opentracker-stm32-core/commit/21cbddd26809e8a4c4dba2bd05e93fc1a1481b28

Another issue that happened while I was trying to debug this nightmare was that the USART kept receiving bytes after hitting a breakpoint and eventually signaled an overrun error. I'm not entirely sure how the error callback should be handled, but the current implementation inside the uart driver does not seem correct: the error flags are not cleared correctly for an STM32L4 (reading the data register is not enough) and probably you should re-enable the RX interrupt because the normal receive callback does not get called.

You can see my solution in this other commit: https://github.com/RoboTech-srl/opentracker-stm32-core/commit/6653e31fa28eb6f0289dfb648a7c1d082ddd6115

I'm not sure if the flags are cleared correctly for all STM32 families, as was only debugging on an STM32L4.

My variant is based on an old STM32 core version, I think it was 1.2.0, but I verified the affected code has not changed significantly in the latest official release 1.5.0.

The first bug depends on how the handle locking mechanism is implemented in the HAL, I don't know if this has changed lately but my guess is it has not. If the lock is already acquired, an asynchrounous function (called from an IRQ for example) simply fails.

And when the Receive_IT() function fails there was no special handling and no way to know something went bad. It took me many hours of debugging to catch this and finally spot the __HAL_LOCK(handle) macro was leaving the function with an error code.

In my solution I just disabled the UART IRQ when the HAL Receive_IT/Transmit_IT are called, implementing a true critical section, but only affecting a single IRQ and not global interrupts.

ShenRen:
This serial problem happens in my 3dprinter frameware too .
It drives me crazy.
I debug and find that RX interrupt will be disable in any time.
I have no ability to solve this problem. Hope a master can solve it.
Help !!!!

ShenRen:
[...]

PR #502, opened 2019-04-17
[PR 496 rework] Disable USART IRQ in uart_write and uart_debug_write 

fpistm:
This PR is a rework of #496  and fixes issues #494 and #467 

It hardens and enhances the original PR.
I've tested several config and it works with Arduino Serial monitor.
Note that if the same U(S)ART is used for `Serial` and `printf()` some bytes could be lost due to IRQ disabling.

@ppescher  and @benwaffle could you review and test it on your side

### DIFF
--- a/cores/arduino/stm32/uart.c
+++ b/cores/arduino/stm32/uart.c
@@ -607,14 +607,24 @@ size_t uart_debug_write(uint8_t *data, uint32_t size)
       if (serial_debug.index >= UART_NUM) {
         return 0;
       }
+    } else {
+      serial_t *obj = rx_callback_obj[serial_debug.index];
+      if (obj) {
+        serial_debug.irq = obj->irq;
+      }
     }
   }
 
+  HAL_NVIC_DisableIRQ(serial_debug.irq);
+
   while (HAL_UART_Transmit(uart_handlers[serial_debug.index], data, size, TX_TIMEOUT) != HAL_OK) {
     if ((HAL_GetTick() - tickstart) >=  TX_TIMEOUT) {
-      return 0;
+      size = 0;
+      break;
     }
   }
+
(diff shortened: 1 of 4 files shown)


cores/arduino/stm32/uart.c
```diff
@@ -581,38 +581,45 @@ void uart_debug_init(void)
...
     if ((HAL_GetTick() - tickstart) >=  TX_TIMEOUT) {
-      return 0;
+      size = 0;
+      break;
     }
   }
 
+  serial_t *obj = rx_callback_obj[serial_debug.index];
+  if (obj && !serial_rx_active(obj)) {
+    HAL_UART_Receive_IT(uart_handlers[obj->index], &(obj->recv), 1);
```
benwaffle: Why are you calling HAL_UART_Receive_IT in uart_debug_write?

benwaffle:
> Note that if the same U(S)ART is used for Serial and printf() some bytes could be lost due to IRQ disabling.
Can you explain under what conditions this would happen?

cores/arduino/stm32/uart.c
fpistm: In fact, I've added this to test. It is removed.
I've fixed an issue with the IRQ which was not correct if Serial already init.

fpistm:
> 
> 
> > Note that if the same U(S)ART is used for Serial and printf() some bytes could be lost due to IRQ disabling.
> 
> Can you explain under what conditions this would happen?

In fact by default, the `printf` will send on the same `U(S)ART` used for the generic instance.
In this case, when the first `Serial.available` is true, then `printf` is called and disabled the `U(S)ART` interrupt then in this window the data is simply lost until the interrupt was set.

fpistm:
I've added some code enhancement which allow to save space and avoid loop to search the right `serial_t obj`.

PR #503, opened 2019-04-19
Re-evaluate interrupt priority values

### LINKED ISSUE #472: Re-evaluate interrupt priority values
I am unable to use uart (via `printf`) in interrupt handlers such as the i2c handler (via `Wire.onRequest`).

I am not sure about all the supported devices, but on the STM32F103, the subpriority is ignored because NVIC_PRIORITYGROUP_4 is used. Currently, the I2C interrupts are set to preemption priority 0. The USART1 preemption priority is also 0. This means that using UART in an interrupt handler will never exit, because HAL_UART_Transmit requires the USART interrupt.

On the other hand, this may be a good idea because using printf() inside an interrupt handler is slow, but interrupt handlers should be really fast.

Currently I can manually do this:
```c
Wire.begin(0x50);
HAL_NVIC_SetPriority(I2C1_ER_IRQn, 1, 0);
HAL_NVIC_SetPriority(I2C1_EV_IRQn, 1, 0);
HAL_NVIC_SetPriority(USART1_IRQn, 0, 0); // usart now has higher priority than i2c
```

Perhaps we should take a look at all of the interrupt priorities in the code base and decide on a global ordering.

fpistm:
Can be redefined thanks `variant.h` or `build_opt.h`

Fix #472 

| IRQ | Old | New |
| --- | --- | --- |
| Systick | 0 | 0 |
| U(S)ART | 0 | 1 |
| USB | 5 | 1 |
| I2C | 0 | 2 |
| EXTI | 6 | 3/6 |
| TIM | 14 | 3/14 |
| RTC | 2 | 2 |

_With **x/y** = x for Cortex-M0 else y_

### DIFF
--- a/cores/arduino/stm32/interrupt.cpp
+++ b/cores/arduino/stm32/interrupt.cpp
@@ -35,7 +35,6 @@
   *
   ******************************************************************************
   */
-#include "stm32_def.h"
 #include "interrupt.h"
 
 /* Private Types */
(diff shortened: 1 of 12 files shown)


fpistm:
@benwaffle ,
let me know your feedback 😉 
thanks in advance

benwaffle:
Thanks, this works for my use case

PR #504, opened 2019-04-22
twi: fix copy-paste error

benwaffle:


cores/arduino/stm32/twi.c
```diff
@@ -395,7 +395,7 @@ i2c_status_e i2c_IsDeviceReady(i2c_t *obj, uint8_t devAddr, uint32_t trials)
 
   switch (HAL_I2C_IsDeviceReady(&(obj->handle), devAddr, trials, I2C_TIMEOUT_TICK)) {
     case HAL_OK:
-      ret = HAL_OK;
+      ret = I2C_OK;
```
fpistm: To fully fix this typo issue, this should also be updated:
https://github.com/stm32duino/Arduino_Core_STM32/blob/64fbe821c536c87cf18a82415a4b0f0289efe867/cores/arduino/stm32/twi.c#L394

benwaffle: fixed

uzi18:
@benwaffle what is value of HAL_OK ?

fpistm:
> @benwaffle what is value of HAL_OK ?

The same as I2C_OK for instance.

uzi18:
@fpistm so this is only typo, not a bug ;)

fpistm:
Right but this could be a bug if HAL_OK value is changed ;)

uzi18:
correct :)

PR #511, opened 2019-04-27
Add support for using -g3 compiler option

dwrobel:
Currently when one need to debug code with full
debugging symbols (including access to macros -g3) then
the linker fails because adding -g3 causes that preprocessed
{preprocld.filepath} file contains a series of #define like:

 #define __STDC__ 1
 #define __STDC_VERSION__ 201112L
 #define __STDC_UTF_16__ 1
 #define __STDC_UTF_32__ 1
 #define __STDC_HOSTED__ 1
 #define __GNUC__ 7
 #define __GNUC_MINOR__ 4
 #define __GNUC_PATCHLEVEL__ 0
 #define __VERSION__ "7.4.0"
 #define __ATOMIC_RELAXED 0
 #define __ATOMIC_SEQ_CST 5
 #define __ATOMIC_ACQUIRE 2
 #define __ATOMIC_RELEASE 3
 ...

which later on causes the linker to not accept such a file:

/usr/lib/gcc/arm-none-eabi/7.4.0/../../../../arm-none-eabi/bin/ld:/home/sw/Arduino/DS18B20Test/.build/preproc.ld:1: ignoring invalid character `#' in expression
/usr/lib/gcc/arm-none-eabi/7.4.0/../../../../arm-none-eabi/bin/ld:/home/sw/Arduino/DS18B20Test/.build/preproc.ld:1: syntax error

The fix removes entirely preprocessing of linker file operation step
and passes an appropriate value of the symbol directly to the linker
using: -Wl,--defsym= option.

Log excerpt:

Linking everything together...
"/usr/bin/arm-none-eabi-gcc" -mcpu=cortex-m3 -mthumb -Os --specs=nano.specs -Wl,--defsym=LD_MAX_SIZE=131072 -Wl,--defsym=LD_MAX_DATA_SIZE=20480

Signed-off-by: Damian Wrobel <dwrobel@ertelnet.rybnik.pl>

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -37,7 +37,7 @@ compiler.cpp.flags={compiler.extra_flags} -c {build.flags.optimize} {compiler.wa
 
 compiler.ar.flags=rcs
 
-compiler.c.elf.flags=-mcpu={build.mcu} -mthumb {build.flags.optimize} {build.flags.ldspecs} -Wl,--cref -Wl,--check-sections -Wl,--gc-sections -Wl,--entry=Reset_Handler -Wl,--unresolved-symbols=report-all -Wl,--warn-common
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @dwrobel 
I assume you set `-g3` in `build_opt.h`?

dwrobel:
> Hi @dwrobel
> I assume you set `-g3` in `build_opt.h`?

I'm using `arduino-builder` so I'm specifying extra flags using: `-prefs=compiler.c.extra_flags=` and `-prefs=compiler.cpp.extra_flags=`

PR #541, opened 2019-06-14
[I2C]  Add general call mode settings

fpistm:
This PR add possibility to enable the I2C general call mode feature.
It replaces #470 .

Adding `true` to the 3 `Wire::begin()` methods will enable the general call mode else `false` per default.

/cc @benwaffle


uzi18:
@LMESTM it looks like if you need to enable general call mode feature, just add parameter "true" to begin, like:
`Wire.begin(true); or Wire.begin(0x70,true);`

fpistm:
I've planned to add in the Wiki after merge:
https://github.com/stm32duino/wiki/wiki/API#i2c

PR #562, opened 2019-07-13
Discussion - Add Basic Dma Handler support

xC0000005:
**Summary**

While working on Neopixel library support, I discovered that sketches had to implement DMA ISR handlers in order to receive DMA updates. This doesn't work in a library, since an ADC library might need to also implement them, and like timers, that should be part of the core.

It takes the same basic form as timer support - an array of DMA handle pointers is maintained and DMA ISR handlers are added. If a DMA handler exists for a given IRQ, it is dispatched.

The usage is simple - call prepare_dma() with the DMA handle to add it, and dma_end() to remove it. In theory, prepare_dma should do more, but I don't understand enough of the DMA use cases to be able to confidently say what should or should not happen for any given call sequence.

The result is that without ISR handlers defined, sketches (and libraries) can use DMA.

There are limitations to this PR - There are DMA "stream" handlers I see, but I don't need these, and haven't added them because I don't know of a good way to test it.

Are there other types of handlers that should be added?

### DIFF
--- a/cores/arduino/stm32/dma.c
+++ b/cores/arduino/stm32/dma.c
@@ -38,6 +38,11 @@
 #include <stddef.h>
 #include "dma.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+#if defined(HAL_DMA_MODULE_ENABLED)
+
 #define NC -1
 
 typedef enum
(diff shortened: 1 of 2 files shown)


xC0000005:
I see travis checks failed - I'll figure out what to do about them shortly.

xC0000005:
I'm going to withdraw this while I fix the stream handlers and work out a few other kinks in the code.

PR #581, opened 2019-07-28
Support LAN8720 ETH BOARD

nopnop2002:
Support LAN8720 ETH BOARD 

Please see here
https://github.com/stm32duino/STM32Ethernet/issues/23


fpistm:
Hi @nopnop2002 
thanks for this PR.
Anyway, I think this is not the right way to support this PHY (mainly due to use of the same name).
I think an update of the library will be better and will allow to support more PHY layer as requested here:
https://github.com/stm32duino/STM32Ethernet/issues/15

I think the PHY configuration will have to be moved from the core to the library.

Issue #583, opened 2019-07-29
STM32 I2C driver "repeated start" issue


jmchiappa:
Hello,

I'm trying to connect to the MMA8451 accelerometer. This chip requires an I2C driver with repeated-start feature enabled to discuss with.
I've  a NUCLEO-F411RE, lib 1.6.1 and no repeated start is generated due to some low level limitations.
Let me explain: 
XferOptions variable is always set to `I2C_NO_OPTION_FRAME` in the `HAL_I2C_Transmit_IT`. So, once the Xfer is completed, it fires `I2C_MasterTransmit_TXE`. As the following condition is always false (stm32f4xx_hal_i2c.c):
`if ((CurrentXferOptions != I2C_FIRST_AND_LAST_FRAME) && (CurrentXferOptions != I2C_LAST_FRAME) && (CurrentXferOptions != I2C_NO_OPTION_FRAME))`
a STOP will be generated (line 5025):
```
      /* Disable EVT, BUF and ERR interrupt */
      __HAL_I2C_DISABLE_IT(hi2c, I2C_IT_EVT | I2C_IT_BUF | I2C_IT_ERR);

      /* Generate Stop */
      SET_BIT(hi2c->Instance->CR1, I2C_CR1_STOP);

```
Does someone can confirm ?
How could we override XferOptions by setting to `I2C_LAST_FRAME_NO_STOP` ?
Ok, I understand that it's not directly related to the framework by itself, but I observed that I2C driver is really different regarding the variant.
In other word, by using `TwoWire::endTransmission(false)`, it should not send a STOP condition (instead of UNUSED(endStop) ).

Thank you,
jean-marc

fpistm:
Hi Jean-Marc,
thanks for the report. Will try to investigate when got a time slot. (will be on vacation soon)
Maybe @LMESTM will have a look when he came back from vacation.

jmchiappa:
Hi Frédéric, 
Thank you for your reply.
After deeper investigation, it seems that `HAL_I2C_Master_Seq_Transmit_IT` is more relevant to generate a 'repeated-start' as XferOptions is passed in argument list.
Curiously, this function is called `HAL_I2C_Master_Sequential_Transmit_IT` in F0,F1,F2,F3 series but it's correctly bind in legacy file. 
I'll have a try as soon as I can (in holidays soon too).

Enjoy your holidays !

fpistm:
Thanks, enjoy too.

[...]

PR #590, opened 2019-08-03
Added the repeated-start feature in I2C driver

jmchiappa:
This solves the repeated-start issue #583 . Tested on F411RE using a MMA8451 accelerometer and OLED SSD1306 I2C peripherals. This seems to be ok but it's a little bit tricky.
It must hugely tested by the community with several I2C peripherals and several platforms.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -218,7 +218,7 @@ uint8_t TwoWire::endTransmission(uint8_t sendStop)
 {
   int8_t ret = 4;
   // check transfer options and store it in the I2C handle
-  if (sendStop == false) {
+  if (sendStop == 0) {
     _i2c.handle.XferOptions = I2C_FIRST_FRAME;
   } else {
     _i2c.handle.XferOptions = I2C_FIRST_AND_LAST_FRAME;
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -216,7 +216,6 @@ void TwoWire::beginTransmission(int address)
 //
 uint8_t TwoWire::endTransmission(uint8_t sendStop)
 {
-  //UNUSED(sendStop);
   int8_t ret = 4;
   // check transfer options and store it in the I2C handle
   if (sendStop == false) {
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -146,9 +146,9 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 
     // perform blocking read into buffer
     uint8_t read = 0;
-    if(sendStop!=false) {
+    if (sendStop != false) {
       _i2c.handle.XferOptions = I2C_LAST_FRAME;
-    }else{
+    } else {
       _i2c.handle.XferOptions = I2C_NO_OPTION_FRAME;
     }
     if (I2C_OK == i2c_master_read(&_i2c, address << 1, rxBuffer, quantity)) {
@@ -219,9 +219,9 @@ uint8_t TwoWire::endTransmission(uint8_t sendStop)
   //UNUSED(sendStop);
   int8_t ret = 4;
   // check transfer options and store it in the I2C handle
-  if(sendStop==false) {
+  if (sendStop == false) {
     _i2c.handle.XferOptions = I2C_FIRST_FRAME;
-  }else{
+  } else {
     _i2c.handle.XferOptions = I2C_FIRST_AND_LAST_FRAME;
   }
 
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -146,7 +146,7 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 
     // perform blocking read into buffer
     uint8_t read = 0;
-    if (sendStop != false) {
+    if (sendStop != 0) {
       _i2c.handle.XferOptions = I2C_LAST_FRAME;
     } else {
       _i2c.handle.XferOptions = I2C_NO_OPTION_FRAME;
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -27,6 +27,7 @@ extern "C" {
 
 #include "Wire.h"
 
+#define I2C_NO_OPTION_FRAME       0xFFFF0000U /*!< XferOptions default value */
 // Initialize Class Variables //////////////////////////////////////////////////
 uint8_t *TwoWire::rxBuffer = nullptr;
 uint8_t TwoWire::rxBufferAllocated = 0;
@@ -116,7 +117,6 @@ void TwoWire::setClock(uint32_t frequency)
 
 uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddress, uint8_t isize, uint8_t sendStop)
 {
-  UNUSED(sendStop);
   if (_i2c.isMaster == 1) {
     allocateRxBuffer(quantity);
     // error if no memory block available to allocate the buffer
@@ -146,6 +146,11 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 
     // perform blocking read into buffer
     uint8_t read = 0;
+    if(sendStop!=false) {
+      _i2c.handle.XferOptions = I2C_LAST_FRAME;
+    }else{
+      _i2c.handle.XferOptions = I2C_NO_OPTION_FRAME;
+    }
     if (I2C_OK == i2c_master_read(&_i2c, address << 1, rxBuffer, quantity)) {
       read = quantity;
     }
@@ -211,8 +216,14 @@ void TwoWire::beginTransmission(int address)
 //
 uint8_t TwoWire::endTransmission(uint8_t sendStop)
 {
-  UNUSED(sendStop);
+  //UNUSED(sendStop);
   int8_t ret = 4;
+  // check transfer options and store it in the I2C handle
+  if(sendStop==false) {
+    _i2c.handle.XferOptions = I2C_FIRST_FRAME;
+  }else{
+    _i2c.handle.XferOptions = I2C_FIRST_AND_LAST_FRAME;
+  }
 
   if (_i2c.isMaster == 1) {
     // transmit buffer (blocking)
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ User can add a STM32 based board following this [wiki](https://github.com/stm32d
 | :---: | --- | :---: | :--- |
 | :green_heart: | BluePill F103C(6-8-B) | *1.2.0* | USB CDC support since *1.5.0*, Maple bootloaders support since *1.6.0* |
 | :green_heart: | BlackPill F103C(8-B) | *1.5.0* |  |
-| :yellow_heart: | BlackPill F401CC | **1.7.0* |  |
+| :yellow_heart: | BlackPill F401CC | **1.7.0** |  |
 | :green_heart: | MapleMini F103CB | *1.2.0* | USB CDC support since *1.5.0*, Maple bootloaders support since *1.6.0* |
 | :green_heart: | HY-TinySTM103T | *1.5.0* |  |
 
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -717,7 +717,7 @@ i2c_status_e i2c_master_write(i2c_t *obj, uint8_t dev_address,
   }
 
   do {
-    if (HAL_I2C_Master_Seq_Transmit_IT(&(obj->handle), dev_address, data, size,obj->handle.XferOptions) == HAL_OK) {
+    if (HAL_I2C_Master_Seq_Transmit_IT(&(obj->handle), dev_address, data, size, obj->handle.XferOptions) == HAL_OK) {
       ret = I2C_OK;
       // wait for transfer completion
       while ((HAL_I2C_GetState(&(obj->handle)) != HAL_I2C_STATE_READY)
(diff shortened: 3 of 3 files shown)


jmchiappa:
Hi @uzi18, yes the second commit concerns style issue (aka whitespaces) only. it fixes some errors during buld on Travis.
But major changes are in my first commit (adding HAL_... seq_....).
Thank you for reviewing it.

uzi18:
looks ok

libraries/Wire/src/Wire.cpp
```diff
@@ -27,6 +27,7 @@ extern "C" {
 
 #include "Wire.h"
 
+#define I2C_NO_OPTION_FRAME       0xFFFF0000U /*!< XferOptions default value */
```
LMESTM: What is this value and how would that be interpreted by ? 
Isn't there a #define amongst I2C_FIRST_AND_LAST_FRAME, I2C_LAST_FRAME,  that can be used instead ?

fpistm: https://github.com/stm32duino/Arduino_Core_STM32/blob/8134a78a09de5b4f6f041c967e2cfaa2b373c210/system/Drivers/STM32F4xx_HAL_Driver/Src/stm32f4xx_hal_i2c.c#L85-L121

I guess I2C_LAST_FRAME_NO_STOP would be better in this case?

libraries/Wire/src/Wire.cpp
```diff
@@ -146,6 +146,11 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 
     // perform blocking read into buffer
     uint8_t read = 0;
+    if (sendStop != false) {
```
fpistm: `sendStop` is an `uint8_t ` so use 0 instead of bool type.
```suggestion
    if (sendStop != 0) {
```

libraries/Wire/src/Wire.cpp
```diff
@@ -211,8 +216,14 @@ void TwoWire::beginTransmission(int address)
 //
 uint8_t TwoWire::endTransmission(uint8_t sendStop)
 {
-  UNUSED(sendStop);
+  //UNUSED(sendStop);
```
fpistm: Please remove comment
```suggestion
```

libraries/Wire/src/Wire.cpp
```diff
@@ -211,8 +216,14 @@ void TwoWire::beginTransmission(int address)
...
   int8_t ret = 4;
+  // check transfer options and store it in the I2C handle
+  if (sendStop == false) {
```
fpistm:
```suggestion
  if (sendStop == 0) {
```

libraries/Wire/src/Wire.cpp
```diff
@@ -146,6 +146,11 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddres
 
     // perform blocking read into buffer
     uint8_t read = 0;
+    if (sendStop != false) {
+      _i2c.handle.XferOptions = I2C_LAST_FRAME;
+    } else {
+      _i2c.handle.XferOptions = I2C_NO_OPTION_FRAME;
```
fpistm: Is `I2C_LAST_FRAME_NO_STOP `would be better in this case?
@LMESTM ?

fpistm:
@jmchiappa any update on this? (I do not know if you came back from vacation 😉 )

libraries/Wire/src/Wire.cpp
jmchiappa: Hi @fpistm, @LMESTM, I'm back !
Well, `I2C_NO_OPTION_FRAME` comes from the stm32f4xx_hal_i2c.c, line 327.
I agree that it's pretty annoying that there's a lot of #define everywhere...
The goal is to not change the behavior of i2c driver when the repeated start is not used. Therefore, the previous routine `HAL_I2C_Master_Transmit_IT` initializes `hi2c->XferOptions = I2C_NO_OPTION_FRAME` (stm32f4xx_hal_i2c.c, line1694).

So, as this value is not defined in the header file stm32f4xx_hal_i2c.h, I need to re-define it in Wire.cpp.

Regarding your suggestions, it may works maybe. But, honestly, I have made a lot of tries before finding a working configuration. The critical test is in stm32f4xx_hal_i2c.c line 5023:
`if ((CurrentXferOptions != I2C_FIRST_AND_LAST_FRAME) && (CurrentXferOptions != I2C_LAST_FRAME) && (CurrentXferOptions != I2C_NO_OPTION_FRAME))`    
So, I agree that this value would work. But why does I2C_NO_OPTION_FRAME has been defined by MCD application team ? Why this value is outside the header instead of being a part of transfer frame option list ? It's a bit messy.

Unfortunately, I don't have the MMA accelerometer anymore.
I hope that someone who gets a device requiring a repeated-start feature will bring its help.
In parallel, I will do my best to get it again.

libraries/Wire/src/Wire.cpp
jmchiappa: Most of Arduino libraries uses the requestFrom routines by passing a bool argument in SendStop position.
I agree that the sendStop type is uint8_t, but in the routine `twi_readFrom` from the Arduino framework says that : `sendStop: Boolean indicating whether to send a stop at the end`.
How to be consistent with the legacy framework ?

fpistm: Even if library use true/false value this will not change behavior.
And this is consistent with legacy framework as prototype is the same

libraries/Wire/src/Wire.cpp
fpistm: Hi @jmchiappa, @LMESTM,
@ABOSTM has discussed with our I2C expert and `I2C_NO_OPTION_FRAME` should not be used.
@ABOSTM will provide more info soon.

libraries/Wire/src/Wire.cpp
fpistm: Wainting update from @ABOSTM.

ABOSTM:
Hi @jmchiappa,
Thanks for your PR.
I had a look at the source code and it seems that using I2C_NO_OPTION_FRAME may cause issue:
1) Call to assert() function in STM32 HAL (if assert is activated) will fail because I2C_NO_OPTION_FRAME is not consider a valid option for external API.
2) Also, on F3 family for example, in HAL_I2C_Master_Seq_Receive_IT()  the XferOptions can be written in xfermode which is then written in hardware register (I2C_CR2) through call to I2C_TransferConfig()
      {
        hi2c->XferSize = hi2c->XferCount;
        xfermode = hi2c->XferOptions;
      }
     . . .
      /* Send Slave Address and set NBYTES to read */
      I2C_TransferConfig(hi2c, DevAddress, hi2c->XferSize, xfermode, xferrequest);

But 0xFFFF0000 is not a valid value to write to register.

So I discussed with our I2C expert, and we finally reach a new proposal:
for both RX and TX:
  if (sendStop == 0) {
    _i2c.handle.XferOptions = I2C_OTHER_FRAME;
  } else {
    _i2c.handle.XferOptions = I2C_OTHER_AND_LAST_FRAME;
  }

Unfortunately F0 and F2 do not yet have those defines but it should come in the future,
also F1/F3 will have it soon (next release).
So we propose to update your proposal with the solution from our expert, only when I2C_OTHER_FRAME define is available. 
And keep former solution for other families (i.e. stop condition always present, and uncompatible for now with device like your accelerometer).
I made a Pull Request on top of yours, in your fork:
https://github.com/jmchiappa/Arduino_Core_STM32/pull/1

I tested this with I2C EEPROM 24LC256 and with Thermal sensor LM75 on NUCLEO_L476RG to be sure there is no regression. Test are passed.

Would it be possible for you to test with your MMA8451 accelerometer ... if you get it ?

fpistm:
@jmchiappa 
is it ok for you ?

jmchiappa:
> 
> 
> @jmchiappa
> is it ok for you ?

Hello Frédéric,
Yes, it is. I should receive the MMA for this weekend.

fpistm:
OK.
Then we will wait your feedback.
I'm currently preparing release 1.7.0 and this should be fine to get it. 😉

jmchiappa:
> 
> 
> OK.
> Then we will wait your feedback.
> I'm currently preparing release 1.7.0 and this should be fine to get it. wink

Hi,
So after a lot of tests, here are some results which are mitigated:
Number of tested frameworks: 3
-1.6.1 - official
-1.6.1 - JM changes
-1.6.1 - ABO proposal

Number of test board : 2
- NUCLEO F411RE
- NUCLEO L476RG

Number of i2c devices : 2
- MMA 8451 (repeated start feature required)
- monochrome OLED 128x64 based on SSD1306 controller (standard I2C driver)

I2C libraries : 3
-Adafruit_MMA8451_Library (MMA library)
-Adafruit_SSD1306 (OLED library)
-U8g2 (OLED library)

Here is the list of tested configurations with their results : 
1.6.1 + MMA + F411RE -> KO
1.6.1 + OLED Adafruit + L476RG-> OK, high number of frame per second (FPS)
1.6.1 + OLED u8g2 +L476RG -> OK, high number of frame per second (FPS)

1.6.1-JM + MMA + F411RE -> OK, no dead time between each i2c frame
1.6.1-JM + Adafruit + L476RG -> OK, high number of frame per second (FPS)
1.6.1-JM + U8g2 + L476RG -> OK, BUT very very low FPS (110 ms of dead time between each i2c frame)
1.6.1-JM + U8g2 + F411RE -> OK, high number of frame per second (FPS)

1.6.1-ABO + MMA + F411RE -> OK, no dead time between each i2c frame
1.6.1-ABO + Adafruit + L476RG -> OK, high number of frame per second (FPS)
1.6.1-ABO + U8g2 + L476RG -> OK, BUT very very low FPS (110 ms of dead time between each i2c frame)
1.6.1-ABO + U8g2 + F411RE-> OK, high number of frame per second (FPS)

So, I think that i2c driver behavior is different for each board.
I don't go further for now, but I think it is not mature enough to be a part of 1.7.0.
your thoughts ?

fpistm:
@jmchiappa 
right, I will not add it to 1.7.0.
Thanks for your quick feedback.
We need to perform further tests.
Anyway, my understanding is you get the same result btw your implementation and @ABOSTM one ? Right?

jmchiappa:
@fpistm, right.
Behavior is the same for both implementations.
Please note that all i2c drivers are not consistent between boards. BTW, could you please clarify the origin of I2C_NO_OPTION_FRAME ? Up to now, this value is present in most HAL i2c drivers from ST.

Thank you for your support and I hope this feature will be available soon.
Let me know if you need more tests.

ABOSTM:
Hi @jmchiappa,
We will try to get a U8g2 compatible device by end of the week, so we could test.
About I2C_NO_OPTION_FRAME, it is used internally (internally to STM32 cube HAL driver) to handle non sequential transfert, in case of "non sequential transfer" (HAL_I2C_Master_Receive_xx() or HAL_I2C_Master_Transmit_xx()) . And it is not aimed to be used in external API and specially not to use "sequential API" for a non sequential transfert.

ABOSTM:
Hi @jmchiappa,
I tested with same OLED kind of display than yours. And face the same difference in measurement.
The Problem comes from U8g2 library that calls for each transfert TwoWire::setClock().
The difference in measurement comes from different I2C hardware versions: 
* one is able to compute by hardware the I2C timing with simply the I2C frequency required (like STM32F4)
* the other need complex software computation... which is of course slower (like STM32L4).
Normally it should be done only once for an application, but in case of U8gé, it is done for each transfert.
I made a PR #646 to workaround this issue: computation is done only once, timings are saved, and reused later (in condition input clock is the same).

I now have high FPS on STM32L4.

There is still one point I would like to clarify with you:
The I2C timing computation which cause issue, is not inegrated in 1.6.1. (timing were hardcoded at that time).
So can you confirm that in fact you made the tests not exactly on 1.6.1, but rather in between 1.6.1 and head of the master ?

jmchiappa:
Hi @ABOSTM, thank you for your quick fix, it sounds good.
My changes have been made on the "pure" 1.6.1 checkout.
I don't have any changes in twi.c file in my repo.
For now, TwoWire::setclock  is calling i2c_setTiming(...)
I do agree that in my stm32duino lib, timings are hardcoded:

```
void i2c_setTiming(i2c_t *obj, uint32_t frequency)
{
  uint32_t f = I2C_10KHz;
  __HAL_I2C_DISABLE(&(obj->handle));

  if (frequency <= 10000) {
    f = I2C_10KHz;
  } else if (frequency <= 50000) {
    f = I2C_50KHz;
  } else if (frequency <= 100000) {
    f = I2C_100KHz;
  } else if (frequency <= 200000) {
    f = I2C_200KHz;
  } else if (frequency <= 400000) {
    f = I2C_400KHz;
  }

#if defined (STM32F0xx) || defined (STM32F3xx) || defined (STM32F7xx) ||\
    defined (STM32G0xx) || defined (STM32H7xx) || defined (STM32L0xx) ||\
     defined (STM32L4xx) || defined (STM32WBxx)
  obj->handle.Init.Timing = f;
#else
  obj->handle.Init.ClockSpeed = f;
  /* Standard mode (sm) is up to 100kHz, then it's Fast mode (fm)     */
  /* In fast mode duty cyble bit must be set in CCR register          */
  if (frequency > 100000) {
    obj->handle.Init.DutyCycle       = I2C_DUTYCYCLE_16_9;
  } else {
    obj->handle.Init.DutyCycle       = I2C_DUTYCYCLE_2;
  }
#endif
  /*
    else if(frequency <= 600000)
      g_i2c_init_info[i2c_id].i2c_handle.Init.ClockSpeed = I2C_600KHz;
    else if(frequency <= 800000)
      g_i2c_init_info[i2c_id].i2c_handle.Init.ClockSpeed = I2C_800KHz;
    else
      g_i2c_init_info[i2c_id].i2c_handle.Init.ClockSpeed = I2C_1000KHz;
  */
  HAL_I2C_Init(&(obj->handle));
  __HAL_I2C_ENABLE(&(obj->handle));
}

```

fpistm:
Fixes #583

LieBtrau:
Hello, will this also fix the same issue on the STM32F103?  
I have issues using the VEML7700 ambient light sensor (as sold by DFRobot & Adafruit).  That sensor requires a repeated start.  Currently I worked around it by using bitbanged I²C.

fpistm:
Probably.

PR #627, opened 2019-09-02
Add SoftwareSerial

### LINKED ISSUE #267: SoftwareSerial needed
Hello

Thanks for your great job  !

A working "SoftwareSerial" instruction for STM32 would be greatly appreciated for compatibility with many existing Arduino sketches.

Best regards

Yannick

armint:
**Summary**

Add SoftwareSerial for STM32 boards.

This PR adds SoftwareSerial support for STM32 boards. The code was based on the SoftwareSerial implementation from the lpc176x arduino port and adapted for STM32. It uses the recently added HardwareTimer code for timing. The timer to be used can be defined in variant.h by defining TIMER_SERIAL. When no TIMER_SERIAL is defined it will use the simplest timer available. 
Tested and working on STM32F446ZE and STM32F401RE. Works with existing SoftwareSerial example code. Tested at 9600 and 115200 baud by communicating with hardware serial interfaces and by logic analyser. 

Fixes  #267


ABOSTM:
Hi @armint,
Thank you for this pull request.
It looks really interresting, and I will make a deep review into your code.

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+}
+
+inline void SoftwareSerial::send()
+{
+  if (--tx_tick_cnt <= 0) {
+    if (tx_bit_cnt++ < 10) {
+      // send data (including start and stop bits)
+      digitalWriteFast(digitalPinToPinName(_transmitPin), (tx_buffer & 1) ? ~_inverse_logic : _inverse_logic);
```
ABOSTM: In order to optimise/minimise Interrupt work:
1) Use STM32 Cube LL to set reset output pin (see example below)
2) save Port and pin number (LL format) in variable of the class thus convertion is done only once (in constructor or in setTX ...) . And in interrupt function recv() and send() you can directly call LL API without overhead.
3) Can we manage _inverse_logic  in write function ? so no work in interrupt fonctions, and also we can make the inversion can be done once for the whole byte (if (_inverse_logic)  tx_buffer = ~tx-buffer;) instead of once every bit.

Example LL:

_receivePinPort = digitalPinToPort(PinName);
_receivePinNumber = STM_LL_GPIO_PIN(digitalPinToPinName(PinName));
LL_GPIO_ReadInputPort(_receivePinPort);

_transmitPinPort = digitalPinToPort(PinName);
_transmitPinNumber = STM_LL_GPIO_PIN(digitalPinToPinName(PinName));
LL_GPIO_SetOutputPin(_transmitPinPort, _transmitPinNumber);  // Set High
LL_GPIO_ResetOutputPin(_transmitPinPort, _transmitPinNumber); // Set Low

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+bool SoftwareSerial::listen()
+{
+  if (_receivePin >= 0) {
+    // wait for any transmit to complete as we may change speed
+    while (active_out)
+      ;
+    if (active_listener) {
+      active_listener->stopListening();
+    }
+    rx_tick_cnt = 1;
```
ABOSTM: Can you add some comment to explain why rx_tick_cnt is set to 1 ?
(I guess it is to sample on 1/3 of a bit with OVERSAMPLE = 3).
As OVERSAMPLE is configurable with a define, maybe you could also use define instead of hard code value '1'

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+}
+
+inline void SoftwareSerial::setRXTX(bool input)
+{
+  if (_half_duplex) {
+    if (input) {
+      if (active_in != this) {
+        setRX();
+        rx_bit_cnt = -1;
+        rx_tick_cnt = 2;
```
ABOSTM: Why rx_tick_cnt value is 2 here whereas it was 1 in listen ?
Maybe you could use define here too

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+    } else {
+      tx_tick_cnt = 1;
+      if (_output_pending) {
+        active_out = nullptr;
+      } else if (tx_bit_cnt > 10 + OVERSAMPLE * 5) {
+        if (_half_duplex && active_listener == this) {
+          // setRXTX(true);
+          pinMode(_receivePin, _inverse_logic ? INPUT_PULLDOWN : INPUT_PULLUP); // pullup for normal logic!
+          rx_bit_cnt = -1;
+          rx_tick_cnt = 2;
```
ABOSTM: dito

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+{
+  if (--rx_tick_cnt <= 0) {
+    //    uint8_t inbit = gpio_get(_receivePin);
+    uint8_t inbit = digitalReadFast(digitalPinToPinName(_receivePin));
+    if (rx_bit_cnt == -1) {
+      // waiting for start bit
+      if (!inbit) {
+        // got start bit
+        rx_bit_cnt = 0;
+        rx_tick_cnt = OVERSAMPLE + 1;
```
ABOSTM: I guess OVERSAMPLE +1 is to sample next bit in the middle of the bit (more precisely inafter 1/3 of the bit due to OVERSAMPLE =3) ?
A comment is welcome.

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+        rx_buffer = 0;
+      } else {
+        rx_tick_cnt = 1;
```
ABOSTM: dito

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+    if (tx_bit_cnt++ < 10) {
+      // send data (including start and stop bits)
+      digitalWriteFast(digitalPinToPinName(_transmitPin), (tx_buffer & 1) ? ~_inverse_logic : _inverse_logic);
+      tx_buffer >>= 1;
+      tx_tick_cnt = OVERSAMPLE;
+    } else {
+      tx_tick_cnt = 1;
+      if (_output_pending) {
+        active_out = nullptr;
+      } else if (tx_bit_cnt > 10 + OVERSAMPLE * 5) {
```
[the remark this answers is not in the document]
armint: This is coming from the original implementation for lpc176x. It seems a pause is required when in half-duplex mode before switching from TX to RX mode. However, the pause here is OVERSAMPLE * 5 bits, which cannot be right. I could be that 5 bits is desired here, but I must admit I have not tested the half-duplex mode. I can remove the half-duplex code completely, or change this to 5 bits delay (and change the condition so the delay is really only there in half-duplex mode)

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+  if (--tx_tick_cnt <= 0) {
+    if (tx_bit_cnt++ < 10) {
+      // send data (including start and stop bits)
+      digitalWriteFast(digitalPinToPinName(_transmitPin), (tx_buffer & 1) ? ~_inverse_logic : _inverse_logic);
+      tx_buffer >>= 1;
+      tx_tick_cnt = OVERSAMPLE;
+    } else {
+      tx_tick_cnt = 1;
+      if (_output_pending) {
+        active_out = nullptr;
```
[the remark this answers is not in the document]
armint: Actually, active_out is not set to nullptr in the else, but in the condition inside the else

cores/arduino/SoftwareSerial.cpp
```diff
@@ -0,0 +1,420 @@
...
+}
+
+//
+// The receive routine called by the interrupt handler
+//
+inline void SoftwareSerial::recv()
+{
+  if (--rx_tick_cnt <= 0) {
+    //    uint8_t inbit = gpio_get(_receivePin);
+    uint8_t inbit = digitalReadFast(digitalPinToPinName(_receivePin));
```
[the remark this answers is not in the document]
armint: True, I will change this

cores/arduino/SoftwareSerial.cpp
armint: Ok, I will adapt the code.

cores/arduino/SoftwareSerial.cpp
armint: Actually, code in recv() is :
`if (--rx_tick_cnt <= 0) {`
  
So when setting rx_tick_cnt to 1, we actually ensure that the first time the recv functions is called the code will start looking for the start bit. However, I do agree this is quite confusing, and I will modify the code so it will be more readable

cores/arduino/SoftwareSerial.cpp
armint: This was meant to discard the next sample when transitioning from TX to RX in half duplex mode.

cores/arduino/SoftwareSerial.cpp
armint: Same here

cores/arduino/SoftwareSerial.cpp
armint: The rewrite will remove this artifact

cores/arduino/SoftwareSerial.cpp
armint: same here

armint:
@ABOSTM Thank you for the review! I have modified the code to address the issues

cores/arduino/SoftwareSerial.cpp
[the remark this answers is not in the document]
ABOSTM: If 5 bit is desired then OVERSAMPLE * 5 bits is good (OVERSAMPLE correspond to 1 bits)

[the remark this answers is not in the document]
armint: I don't think so. We're comparing tx_bit_cnt here, not tx_tick_cnt.

cores/arduino/SoftwareSerial.cpp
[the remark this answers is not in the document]
ABOSTM: Right active_out  is set in  'if (_output_pending) '  and in 'else if (tx_bit_cnt > 10 + OVERSAMPLE * 5)' but not in 'else' which is omited

libraries/SoftwareSerial/src/SoftwareSerial.cpp
```diff
@@ -0,0 +1,432 @@
...
+      }
+      tx_buffer >>= 1;
+      tx_tick_cnt = OVERSAMPLE - 1; // first sample sent, OVERSAMPLE - 1 more samples to go
+    } else {
+      tx_tick_cnt = 0;
+      if (_output_pending || !(_half_duplex && active_listener == this)) {
+        active_out = nullptr;
+        // When in half-duplex mode, we wait for HALFDUPLEX_SWITCH_DELAY bit-periods after the byte has
+        // been transmitted before allowing the switch to RX mode
+      } else if (tx_bit_cnt > 10 + HALFDUPLEX_SWITCH_DELAY) {
```
ABOSTM: I think that if 5 bits are required you need (
* HALFDUPLEX_SWITCH_DELAY * OVERSAMPLE     (old style when send() check if (--tx_tick_cnt <= 0))
*  (HALFDUPLEX_SWITCH_DELAY * OVERSAMPLE) - 1  (new style with if (--tx_tick_cnt < 0) )

libraries/SoftwareSerial/src/SoftwareSerial.cpp
```diff
@@ -0,0 +1,432 @@
...
+    if (tx_bit_cnt++ < 10) {
+      // send data (including start and stop bits)
+      if (tx_buffer & 1) {
+        LL_GPIO_SetOutputPin(_transmitPinPort, _transmitPinNumber);
+      } else {
+        LL_GPIO_ResetOutputPin(_transmitPinPort, _transmitPinNumber);
+
+      }
+      tx_buffer >>= 1;
+      tx_tick_cnt = OVERSAMPLE - 1; // first sample sent, OVERSAMPLE - 1 more samples to go
```
ABOSTM: I disagree with your comment: with tx_tick_cnt = OVERSAMPLE - 1, we still need to wait OVERSAMPLE samples: 
OVESAMPLE - 1 =2, so next interrupt , tx_tick_cnt is decremented (1) before testing 
next interrupt  tx_tick_cnt is decremented again (0) before testing 
next interrupt  tx_tick_cnt is decremented again (-1) before testing and macth condition <0
So I count 3 interrupt (so OVERSAMPLE samples).
I think it was not bad to have former check    if (--tx_tick_cnt <= 0), 
but just add some comments to explain how it works and why some times value are set to 1 or 2.

libraries/SoftwareSerial/src/SoftwareSerial.cpp
```diff
@@ -0,0 +1,432 @@
...
+inline void SoftwareSerial::recv()
+{
+  if (--rx_tick_cnt < 0) {
+    bool inbit = LL_GPIO_IsInputPinSet(_receivePinPort, _receivePinNumber) ^ _inverse_logic;
+    if (rx_bit_cnt == -1) {
+      // waiting for start bit
+      if (!inbit) {
+        // got start bit
+        rx_bit_cnt = 0;
+        rx_tick_cnt = OVERSAMPLE - 1; // first sample received, we need OVERSAMPLE - 1 more samples
```
ABOSTM: The former value was OVERSAMPLE+1, so here (with the new check <0) I think you need OVERSAMPLE.
My understanding is that after the start detection, we want to wait 1 bit +1/3 of bit so that sample are made not to close to the edge, but more in the middle of the bit (to be more tolerant to clock inaccuracy)

cores/arduino/SoftwareSerial.cpp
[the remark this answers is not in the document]
ABOSTM: Yes, but after the 10 bit are send, tx_tick_cnt is no more set to OVERSAMPLE. And so , once tx_tick_cnt reach negative value, it will remain negative. *
So  'if (--tx_tick_cnt <= 0)' will match at each interrupt. Then tx_bit_cnt will be incremented at each interrupt.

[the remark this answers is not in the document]
armint: Right, I didn't catch that one. I will change it.

libraries/SoftwareSerial/src/SoftwareSerial.cpp
armint: I disagree with you analysis and I think the OVERSAMPLE+1 in the original code was an error. Let's take an example where we have a degraded signal, caused by an open-drain pull-up setup with a pull-up (a quite common setup). Now, if the pull-up and line capacitance are relatively high, we see a slowly rising signal, and a high bit coming after a low bit will be only two samples wide instead of three. When we also add the extra delay of one sample period like you suggest, we will mis-interpret the high bit. This would not be the case without the extra delay. 
At the same time, I don't see how clock inaccuracies can be remedied by a delay here, so I don't see the advantage of it.

armint: Also realise this: the first sample is not necessarily on the edge, but can be anywhere between the edge and 1 sample period after the edge

ABOSTM: I see it differently:
I agree that 1st sample is not necessarily on the edge,
it could be between the edge and the 1/3 of the bit (with OVERSAMPLE = 3).
Lets take the case that next bit after start is 1.
If the frequency of the sender is not accurate, the rising edge could arrive later than expected and thus if we continue sampling between theoretical edge of bit and 1/3 of the bit there is a risk that we sample the bit before the rising edge. 

Now If I take your example of a slowly rising edge due to high capacitance, then bit will be only 2 sample wide. But it also imply that lower state is longer than expected (otherwise baurate is far from accurate).
So if we continue sampling between theoretical edge of bit and 1/3 of the bit there is a risk that we sample the bit before the rising edge.

In both case, if we wait for a while to avoid sampling close to the theoretical edge, but more on the middle then we can tolerate drift on the frequency or slowrising edge.

Untitled

fpistm: I've made a test with a NEO6-GPS some bytes are missed with `-1`
```
GPGLL,,,,,092159.00,V,N*4C$GPRMC,092200.0,V,,,,,,,060919,,,N*73
$GPVG,,,,,,,,,N*30
$GPGGA,09200.00,,,,,0,00,99.99,,,,,,*6F$GPGSA,A,1,,,,,,,,,,,,,99.99,99.9,99.99*30
$GPGSV,1,1,04,1,,,22,21,,,22,23,,,21,26,,,2*77
```
 while not if removed.

stas2z:
Tested it with Nucleo F103
57600 cause errors
```
15:58:03.187 -> _S recv: 1111111111111111111111111 : 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 0D 0A 
15:58:03.319 -> MS recv: 111111111111111111111111" : 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 22 1A 14 
15:58:03.319 -> S2 recv: 1111111111111111111111111 : 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 0D 0A 

```
(where S is what was recieved from user console, MS - Software serial recieved from hardware serial, S2 - hardware serial received from software serial, serials are hardwired on the board (rx->tx))
at 115200 board doesn't boot at all

libraries/SoftwareSerial/src/SoftwareSerial.cpp
armint: You're right. I mistakenly assumed that we were actually oversampling here, but what is really being done is just single sampling with a higher resolution clock. And then indeed it makes sense to wait one clock period to make sure the sample is in the middle of the bit.

armint: So... this got me thinking. Why not do some actual oversampling? The extra overhead is minimal and it should bring us more accuracy in case of degraded signals. This will be my last bit of stubbornness on this subject, if you guys think the current code is better I will revert it :)

fpistm:
Closed by #645

PR #631, opened 2019-09-03
Update README.md

dds90:
DIYMROE STM32F407VGT 	to
DIYMORE STM32F407VGT


fpistm:
Hi @dds90 
in fact this was made intentionally by the author of the variant
If you check the silkscreen of this picture:
image

Anyway, I think it is a mistake on the silkscreen.

PR #641, opened 2019-09-08
Fix callbacks array init. Move nptr check in the start of CCcb

ghent360:
**Summary**

HardwareTimer does not 0 all callbacks in the array, resulting in invoking random code


fpistm:
Thanks @ghent360 
Nice catch ;)

PR #643, opened 2019-09-09
Add half-duplex support to HardwareSerial.

ghent360:
**Summary**
TMC2209 and 2208 drivers use UART communication with a single wire. Enabling half-duplex support in the HardwareSerial class would enable using these drivers without wasting another pin for the RX signal.

Proposed API is to enable half-duplex mode if the RX pin on the serial port is NC. Also added a method to set the half-duplex mode explicitly.

**Validation**

Tested using:
https://gist.github.com/ghent360/dc2a7226919af522e4a5914da1ad5252

### DIFF
--- a/cores/arduino/stm32/uart.c
+++ b/cores/arduino/stm32/uart.c
@@ -131,7 +131,7 @@ void uart_init(serial_t *obj, uint32_t baudrate, uint32_t databits, uint32_t par
   USART_TypeDef *uart_rx = pinmap_peripheral(obj->pin_rx, PinMap_UART_RX);
 
   /* Pins Rx/Tx must not be NP */
-  if (uart_rx == NP || uart_tx == NP) {
+  if ((obj->pin_rx != NC && uart_rx == NP) || uart_tx == NP) {
     core_debug("ERROR: at least one UART pin has no peripheral\n");
     return;
   }
@@ -288,7 +288,9 @@ void uart_init(serial_t *obj, uint32_t baudrate, uint32_t databits, uint32_t par
 
   /* Configure UART GPIO pins */
   pinmap_pinout(obj->pin_tx, PinMap_UART_TX);
-  pinmap_pinout(obj->pin_rx, PinMap_UART_RX);
+  if (uart_rx != NP) {
+    pinmap_pinout(obj->pin_rx, PinMap_UART_RX);
+  }
 
   /* Configure uart */
   uart_handlers[obj->index] = huart;
@@ -359,7 +361,11 @@ void uart_init(serial_t *obj, uint32_t baudrate, uint32_t databits, uint32_t par
   }
 #endif
 
-  if (HAL_UART_Init(huart) != HAL_OK) {
+  if (uart_rx == NP) {
+    if (HAL_HalfDuplex_Init(huart) != HAL_OK) {
+      return;
+    }
+  } else if (HAL_UART_Init(huart) != HAL_OK) {
     return;
   }
 }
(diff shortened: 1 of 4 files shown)


cores/arduino/stm32/uart.c
```diff
@@ -128,10 +128,14 @@ void uart_init(serial_t *obj, uint32_t baudrate, uint32_t databits, uint32_t par
 
   /* Determine the U(S)ART peripheral to use (USART1, USART2, ...) */
   USART_TypeDef *uart_tx = pinmap_peripheral(obj->pin_tx, PinMap_UART_TX);
-  USART_TypeDef *uart_rx = pinmap_peripheral(obj->pin_rx, PinMap_UART_RX);
+  USART_TypeDef *uart_rx = NP;
```
fpistm: If `obj->pin_rx` is `NC ` then `uart_rx` will be `NP`.
So I fdon't think this change is required.

cores/arduino/stm32/uart.c
```diff
@@ -732,6 +742,34 @@ void uart_attach_tx_callback(serial_t *obj, int (*callback)(serial_t *))
...
 
+/**
+ * Enable transmitter for half-duplex mode. NOOP in full-fuplex mode
+ * 
+ * @param obj : pointer to serial_t structure
+ * @retval none
+ */
+void uart_enable_tx(serial_t *obj)
+{
+  if (obj == NULL || obj->pin_rx != NC) {
```
fpistm:
```suggestion
  if ((obj != NULL) && (obj->pin_rx == NC)) {
```

invert test and avoid using `return` (I know this is done in other function but I try to remove them when I modify the code :wink: ).

cores/arduino/stm32/uart.c
```diff
@@ -732,6 +742,34 @@ void uart_attach_tx_callback(serial_t *obj, int (*callback)(serial_t *))
...
+
+/**
+ * Enable receiver for half-duplex mode. NOOP in full-fuplex mode
+ * 
+ * @param obj : pointer to serial_t structure
+ * @retval none
+ */
+void uart_enable_rx(serial_t *obj)
+{
+  if (obj == NULL || obj->pin_rx != NC) {
```
fpistm: dito
```suggestion
  if ((obj != NULL) &&  (obj->pin_rx == NC)) {
```

AnHardt:
> Proposed API is to enable half-duplex mode if the RX pin on the serial port is NC. Also added a method to set the half-duplex mode explicitly.

Could also be nice if TX_pin == RX_pin would set up half-duplex.

fpistm:
> > Proposed API is to enable half-duplex mode if the RX pin on the serial port is NC. Also added a method to set the half-duplex mode explicitly.
> 
> Could also be nice if TX_pin == RX_pin would set up half-duplex.

Wit a SoftwareSerial this could be possible but with HardwareSerial there no U(S)ART  RX and TX on the same pin, as far as I know.

cores/arduino/stm32/uart.c
ghent360: done

fpistm:
Thanks @ghent360 , I have to test it. Do you have a sketch example to test this? Else I will write one.

fpistm:
> > > Proposed API is to enable half-duplex mode if the RX pin on the serial port is NC. Also added a method to set the half-duplex mode explicitly.
> > 
> > 
> > Could also be nice if TX_pin == RX_pin would set up half-duplex.
> 
> With a SoftwareSerial this could be possible but with HardwareSerial there no U(S)ART RX and TX on the same pin, as far as I know.

My bad, after some read some documentations this would be possible.
I'm currently reviewing this PR and it seems there is an issue on how the Tx pin is configured.
Documentation recommend this:

> The USART can be configured to follow a single-wire half-duplex protocol where the Tx and Rx lines are internally connected. In this communication mode, only the Tx pin is used forboth transmission and reception.
> The Tx pin is always released when no data is transmitted, thus, it acts as a standard I/O in idle or
> reception modes.
> This means that the I/O must be configured so that the Tx pin is configured as an alternate function open-drain with an external pull-up.

Currently this is not the case as by default it is set in PushPull

PR #701, opened 2019-10-13
New board definition - Adafruit Feather STM32F405

ladyada:
hiya this definition adds support for a new board
https://www.adafruit.com/product/4382

pins are tested as GPIO - having some issue with USB Serial port but will open a separate issue about it


fpistm:
Hi @ladyada 
Nice to see you here 😉 
Thanks for this variant. I will review it carefully.

fpistm:
Hi @ladyada 
if I'm not wrong this is an STM32F405RG ?

ladyada:
yes thats' right
good news - last night i tested all featherwings with the feather. only a few oddities:


- usually when using USB serial, the hardware serial port is called Serial1, but instead stm32duino names them after the USART so the hardware serial is Serial3 - not a big deal but like i said, unusual
- RFM69/RFM9X library wouldn't work, not sure why
- variant doesnt allow creation of multiple SPI / I2C ports like modern arduino does. you can do it by hand, so it isn't blocking - but would be nice if it were possible.
- STM32F4 often have a USB dfu bootloader built in, which is great, but you have to kick into the bootloader each time. it would really help workflow a lot if it was automatic. for example, opening USBSerial port at 1200 baud would trigger a jump to the USB DFU bootloader.

The third item would be by far the most useful for workflow support because right now workflow is optimized for boards which have built in debugger (expensive) or STLink dongle (unecessary!)

ladyada:
Hiya here's an examples of how to create multiple I2C/SPI peripherals. 

image

they are auto-generated here:

https://github.com/adafruit/ArduinoCore-samd/blob/master/libraries/SPI/SPI.cpp#L469

fpistm:
Thanks, will check that.
Why did you close this PR?

ladyada:
i thought this was superceded by the other PR? ok ill reopen!

ladyada:
oh i see this is the same one. sorry about that! late nite hacking :D

ladyada:
ok for workflow question - the great thing about the arduino is that before it, you had slow workflow - to program a new chip you had to manually launch the bootloader. that means pressing a button or something the dev board. with the stm32f405 feather now, because we don't have a built in stlink or openocd chip, you have to manually enter the bootloader my holding BOOT0 high and then pressing reset. If this was done automatically when you click UPLOAD it would allow for very fast iteration. that's the only thing missing :)

fpistm:
Ok understood. 
As said this is planned 😉

ladyada:
ok thanx - how can i help get it implemented

fpistm:
Firstly, could you open an issue for:
- [SPI|I2C]_INTERFACES_COUNT 
- Automatic restart in DFU Bootloader mode

This will allow to track them.

variants/FEATHER_F405/variant.h
```diff
@@ -0,0 +1,167 @@
...
+#define PIN_SPI_MOSI            PB15
+#define PIN_SPI_MISO            PB14
+#define PIN_SPI_SCK             PB13
+#define PIN_SPI_SS              PA15
+
+// SPI FLASH
+#define PIN_SPI3_MOSI           PB5
+#define PIN_SPI3_MISO           PB4
+#define PIN_SPI3_SCK            PB3
+#define PIN_SPI3_SS             PA15
```
fpistm: Just wondering if `PIN_SPI_SS` should be the same than `PIN_SPI3_SS`

ladyada: we dont have an 'assigned hardware SS pin we just need to set it to something

fpistm:
> ok thanx - how can i help get it implemented

Hi @ladyada,
You can try the automatic restart, see #710 .
I did not had time to update script under Windows anyway you can test it manually or update the windows scripts based on what I've made for the linux one.
Let me k ow if you meet any issue.
Thanks

ladyada:
ok great, i use windows 10 :)

ladyada:
could you merge this PR, and rebase the #710 branch - so i only have to follow one branch :)

fpistm:
done

ladyada:
💙 thank u 💙

fpistm:
Welcome and welcome as contributor 😉

ladyada:
hiya i'm about to release the hardware for this, do you know if there's an eta for when the BSP is going to get the next minor update? i will have folks `git clone` the BSP for now

fpistm:
Well I don't think before december. I try to make a release every 3 months.

ladyada:
oki

PR #718, opened 2019-10-22
Add dfu-util as a standard upload method

jgillick:
**Summary**

This PR implements the following features:

* [ ] Adds `dfu-util` as a standard menu upload option.

This is a follow-up to (and dependant on) PR https://github.com/stm32duino/Arduino_Tools/pull/46. 

**Motivation**
The `dfu-util` program is already bundled with the stm32duino packages, so setting it as a primary supported upload tool reduces the need for outside software dependencies and gets the end-user setup and running quicker. Additionally, the STM32CubeProgrammer installer [is known](https://community.st.com/s/question/0D50X00009cdVnZSAU/stm32cubeprogrammer-fails-to-install-in-macos) to have issues on the mac -- there are workarounds, but it's one more hurdle for the end-user to jump through.

**Validation**

* Upload a sketch to a board using the dfu-util upload option.


boards.txt
```diff
@@ -992,6 +992,12 @@ GenF4.menu.pnum.FEATHER_F405.build.product_line=STM32F405xx
 GenF4.menu.pnum.FEATHER_F405.build.variant=FEATHER_F405
 
 # Upload menu
+GenF4.menu.upload_method.dfuUtil=USB (dfu-util)
+GenF4.menu.upload_method.dfuUtil.upload.tool=dfu_util
+GenF4.menu.upload_method.dfuUtil.upload.protocol=
+GenF4.menu.upload_method.dfuUtil.upload.usbID=0483:df11
+GenF4.menu.upload_method.dfuUtil.upload.address=0x08000000:leave
```
fpistm: I guess this should not be hardcoded if we want flash at another address (ex: user have a bootloader)

jgillick: How would you suggest we make this value dynamic? If the user had a custom bootloader or wanted to flash to a different address, I'd assume they'd be running the command manually from the terminal. (wouldn't it also require the linker script to be updated?) To make it dynamic, are you suggesting a new menu option to choose addresses from?

jgillick:
The automatic reset in BL mode PR looks cool. Let me know if there's anything I can do to help with that, too.

boards.txt
fpistm: Linker script can already manage this kind of feature as it is preprocessed.
This is already done for BP F103 for example.
flash_offset is currently hardcoded but can be override by a platform.local.txt
Well, this need some investigation :wink:

jgillick: If I understand the example `flash_offset` example, the solution is to rename `upload.address` to `build.flash_offset` to be consistent with the other build flags (although, I feel like I'm missing something).

jgillick: If I understand the `flash_offset` example, the advice would be to change the variable from `upload.address` to `build.address` and likely move the `:leave` flag to the command in platform.txt. I feel like I must be missing something, though.

Originally I was going to rename this to `build. flash_offset` to match the examples, but that would offset where in the program flash it was writing to (which would reduce the overall size of the program).

fpistm: If a custom bootloader (or any other custom data) is present  the max size of the program is reduced.
It assumes that flash_offset  is from the start of the flash.
I wonder if by default the `--dfuse-address {upload.address}` is required maybe use it only if offset is not null.
This is what is done in maple_upload IIRW.

jgillick: Unfortunately, removing `--dfuse-address` doesn't seem to work, with or without the offset being set. dfu-util returns the error:
```
dfu-util: Only DfuSe file version 1.1a is supported
```

From [what I can tell](https://github.com/particle-iot/device-os/issues/706#issuecomment-149600540) this is because of the version of the bootloader is not setting the flash address correctly from the `alt` setting passed to dfu-util. ([error in code](https://sourceforge.net/p/dfu-util/dfu-util/ci/v0.9/tree/src/dfuse.c#l641))

It looks like this might be fixed with dfu-util 1.0, which isn't released yet.

matthijskooijman: > It looks like this might be fixed with dfu-util 1.0, which isn't released yet.

I could not find anything in the changelog you linked about this. Are you perhaps referring to [this commit](https://sourceforge.net/p/dfu-util/dfu-util/ci/64c9f83c60e2fb3f3c6c7c6e6c967b8013b9c6d4/)? If so, I do not think that fixes this, since it only prevents an address of 0 to be treated the same as "no address specified", but when you specify 0 it will actually try to download to address 0, not the start of the area.

I guess dfu-util could be made smarter to default to the start of the first segment (it already parses the DFU altsetting name (e.g. `@Internal Flash  /0x08000000/04*016Kg,01*064Kg,03*128Kg`) to discover the memory layout (needed for STM32-specific erase additions to the DFU protocol). It does not look the current code already does this (assume the first address), though.

jgillick: Originally I was referencing this line:
> Improve alternate interface index matching (Aleks Chakin)

But, after some research, it doesn't look this would help...

I think I figured out why the `alt` flag works for Maple. It seems that maple boards use the 1.1 version of the DFU bootloader protocol ([source](https://github.com/leaflabs/maple-bootloader/blob/master/usb_descriptor.c#L74), [discussion](http://forums.leaflabs.com/forums.leaflabs.com/topic7309.html?id=1410)) and STM32 chips, by default, use 1.1a. The difference is that 1.1 gets the memory addresses from the chip, while the 1.1a protocol requires the host to specify the address explicitly. ([more info](http://dfu-util.sourceforge.net/dfuse.html))

I wish the `dfu-util` would provide an option to override this -- since it already has these addresses from the USB descriptor -- but alas.

The solution preferred by `dfu-util` is to provide a dfu file, instead of a bin, which includes all this information embedded in the file ([source code](https://sourceforge.net/p/dfu-util/dfu-util/ci/master/tree/src/dfu_file.c#l246)). STM32 provides a [bin -> dfu file converter](https://docs.ghielectronics.com/hardware/loaders/stm32-bootloader.html), however, that program also requires the starting flash address to work.

**Conclusion**
So, ultimately, it seems we're left with the requirement to send the flash address with the command (unless I'm missing something). It would be _possible_ to write a script which uses `dfu-util list` to parse the flash addresses first, but that sounds like overkill and error-prone.

matthijskooijman: > I think I figured out why the alt flag works for Maple. It seems that maple boards use the 1.1 version of the DFU bootloader protocol (source , discussion) and STM32 chips, by default, use 1.1a. The difference is that 1.1 gets the memory addresses from the chip, while the 1.1a protocol requires the host to specify the address explicitly. (more info)

I actually think it is a little bit different. In the official 1.1 protocol, there is no concept of addresses. You just pass a blob of binary to the bootloader, and the bootloader will decide where to place it. The only way to choose between different placements is using different alts (but again, there is no standard way to map these onto addresses).

The "1.1a" version of the protocol (which, AFAICS is not an official standard but something STM came up with and labeled with a non-standard version number which can now no longer be used for future standard versions) does things differently: You must explicitly tell the bootloader where the binary should be uploaded (in addition, sectors must be explicitly erased before writing, which the DFU standard states should be handled by the bootloader automatically). For this, the programmer needs info about addresses and sector layout, which is encoded in the "name" of the alt (e.g. `@Internal Flash /0x08000000/04*016Kg,01*064Kg,03*128Kg` encodes the name, address and sector layout in an STM-specific format).

> I wish the dfu-util would provide an option to override this -- since it already has these addresses from the USB descriptor -- but alas.

Yes, I think all info should be present and such a change to dfu-util is feasible (probably not even very complicated). I would want to have a stab at this, but it's enough of a priority for me or my customer to actually free up time for this, I'm afraid.

> So, ultimately, it seems we're left with the requirement to send the flash address with the command (unless I'm missing something). It would be possible to write a script which uses dfu-util list to parse the flash addresses first, but that sounds like overkill and error-prone.

Agreed.

matthijskooijman: If you want to have stab at implementing the auto address selection in dfu-util, let me know and I'll point you to some things I found while reading the source yesterday :-)

fpistm: @matthijskooijman this could be interesting to know 😃

alextrical:
Any updates on this, I also believe that dfu-util will be a great feature to have as a default. 

It looks like the dependencies have been merged, what is outstanding to get this pull merged?

fpistm:
> Any updates on this, I also believe that dfu-util will be a great feature to have as a default.

Same answer as before:

> Anyway, I will not merge this until the automatic reset in BL mode is not finished.

Unfortunately, I had no time to work on this soon but it will come.

alextrical:
Is there anything I can do to help? My free time is a bit sporadic, and im not the most experienced in this field, but i want to help with the project in general. In the mean time I will have a look through the list of requested boards and see if i can get some of them added as PR's

jerkey:
> > Any updates on this, I also believe that dfu-util will be a great feature to have as a default.
> 
> Same answer as before:
> 
> > Anyway, I will not merge this until the automatic reset in BL mode is not finished.
> 
> Unfortunately, I had no time to work on this soon but it will come.

Unfortunately i just spent all night figuring out how to do exactly what alextrical had already done because i didn't know about this PR, and I was about to make the same PR.  The automatic reset works on my board with :leave (which was removed) so I don't know what the holdup is.  Please merge the work!

fpistm:
@jerkey Which board was removed ?

jerkey:
> @jerkey Which board was removed ?

i was referring to [this commit](https://github.com/stm32duino/Arduino_Core_STM32/pull/718/commits/b9335cf27ca7836cdf98af4803efab787bbd57d6) where ":leave" was removed from the dfu command.
["The device automatically leaves DFU mode after flashing is complete and runs the program. No rearranging jumpers required! To flash the firmware again, just unplug / replug the USB cable to get the board back into DFU mode."](https://rohfle.gitlab.io/posts/2019-10-20-leaving-dfu-mode-without-flashing/)

I don't understand why that commit was made, and you said "Anyway, I will not merge this until the automatic reset in BL mode is not finished" and since :leave makes the board start the user program after flashing, it seemed like the right thing to do but maybe this is accomplished somehow in the present PR.  FYI the hardware will enter dfu mode if boot0 is pulled high during power/reset (via a button for example) and the :leave leaves dfu mode and runs the program.  I don't see how you could improve on this, we don't have a DTR toggle or anything like that - the device doesn't even enable USB unless the user program chooses to do that.  The user has to hold their dfu button and press reset to enter dfu mode.

fpistm:
I close this PR as it is outdated due to several change in the boards.txt.
Add dfu-utils as based tools seems to be a good ideas anyway it is not compatible with all STM32 series.
I'm currently study a way to provide the STM32CubeProgrammer cli as a packages.

PR #748, opened 2019-10-31
I2C Fix

### LINKED ISSUE #583: STM32 I2C driver "repeated start" issue

fpistm:
* Added the repeated-start feature in I2C driver (prviously done thanks #590).    
   Fixes #583
* Fix error management in Read/Write
   When NACK is received by master, instead of immediate re-transmission
   by twi driver, error is reported. Up to application to manage re-transmission.
* Review errors management to be aligned with Arduino [Wire.endTransmission()](https://www.arduino.cc/en/Reference/WireEndTransmission) API Reference:
  - 0:success
  - 1:data too long to fit in transmit buffer
  - 2:received NACK on transmit of address
  - 3:received NACK on transmit of data
  - 4:other error

### DIFF
--- a/cores/arduino/stm32/twi.c
+++ b/cores/arduino/stm32/twi.c
@@ -731,9 +731,10 @@ i2c_status_e i2c_master_write(i2c_t *obj, uint8_t dev_address,
                               uint8_t *data, uint16_t size)
 
 {
-  i2c_status_e ret = I2C_ERROR;
+  i2c_status_e ret = I2C_OK;
   uint32_t tickstart = HAL_GetTick();
   uint32_t delta = 0;
+  uint32_t err = 0;
 
   /* When size is 0, this is usually an I2C scan / ping to check if device is there and ready */
   if (size == 0) {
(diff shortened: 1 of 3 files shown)


cores/arduino/stm32/twi.c
```diff
@@ -807,38 +806,36 @@ i2c_status_e i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint16_t size)
...
+        break;
       }
     }
-    /* When Acknowledge failure occurs (Slave don't acknowledge it's address)
-       Master restarts communication */
-  } while (((HAL_I2C_GetError(&(obj->handle)) & HAL_I2C_ERROR_AF) == HAL_I2C_ERROR_AF)
-           && (delta < I2C_TIMEOUT_TICK));
+
+    err = HAL_I2C_GetError(&(obj->handle));
+    if ((delta > I2C_TIMEOUT_TICK)
```
matthijskooijman: This introduces a new problem: When this timeout is triggered, I2C_OK is triggered. This happens because in the loop above, it checks for `(delta < I2C_TIMEOUT_TICK)`, so it breaks out of the loop as soon as `delta == I2C_TIMEOUT_TICK`. Then, this if checks for `(delta > I2C_TIMEOUT_TICK)`, which is not yet false. Since the hardware reports no errors either, `I2C_OK` is returned.

We noticed this when something confused the I2C hardware and prevented it from continuing somehow. We haven't figured out what the underlying cause is, but we did notice that we did not get an I2C error (just invalid data), which I think can be traced back to this problem.

The same applies to `i2c_write_master` above, I think.

Fixing this is, I think, a matter of replacing `>` with `>=` here.

matthijskooijman: One more thing: This `>` and `<` combination was already present in the old code, but due to the old structure there it would, I think, only be problematic when an ACK failure occured *exactly* a the timeout, which is of course extremely likely. These refactorings have caused this to happen on every timeout instead.

> Fixing this is, I think, a matter of replacing > with >= here.

We just tried this, and it seems to work. @hendrykaak is considering to prepare a PR for this.

fpistm: Right

PR #749, opened 2019-11-02
Fix single-pin half-duplex in SoftwareSerial::send

sjasonsmith:
**Summary**

When SoftwareSerial was adapted from the LPC1768 version, SoftwareSerial::send was altered in a way that prevents single-pin half-duplex communication from working properly.

This change reverts to the behavior seen in other repositories using variations of this code.

* [ ] Fixes one-wire half-duplex communication
* [ ] No intentional breaking changes

This issue was discovered while testing a [Marlin pull request](https://github.com/MarlinFirmware/Marlin/pull/15655) which is adapting Marlin to work with STM32 5.7. During testing I found that one-pin communication with TMC stepper drivers did not work properly, although it did when using similar SoftwareSerial code from other sources.

The image below shows the same read command being sent to a TMC2208 stepper driver, both before and after the change in this PR. (Doesn't scale well, click to zoom)
SWSerial_Send_Compare

**Validation**

I have validated that I can now communicate properly with TMC stepper drivers using one-wire connections inside Marlin.

### DIFF
--- a/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -227,15 +227,15 @@ inline void SoftwareSerial::send()
       tx_tick_cnt = OVERSAMPLE; // Wait OVERSAMPLE tick to send next bit
     } else { // Transmission finished
       tx_tick_cnt = 1;
-      if (_output_pending || !(_half_duplex && active_listener == this)) {
+      if (_output_pending) {
         active_out = nullptr;
-        rx_bit_cnt = -1; // rx_bit_cnt = -1 :  waiting for start bit
-        rx_tick_cnt = 2; // 2 : next interrupt will be discarded. 2 interrupts required to consider RX pin level
-        active_in = this;
-        // When in half-duplex mode, we wait for HALFDUPLEX_SWITCH_DELAY bit-periods after the byte has
+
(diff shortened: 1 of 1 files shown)


sjasonsmith:
Tagging @ABOSTM. Perhaps they have insight into why these changes were originally made when porting SofrtwareSerial to STM32.

sjasonsmith:
I will modify my board this morning so that I can test when using two serial pins, to make sure I didn't break that.

sjasonsmith:
> I will modify my board this morning so that I can test when using two serial pins, to make sure I didn't break that.

I just verified that I can also communicate with my TMC2208 stepper driver using two pins from Marlin.

sjasonsmith:
I ran into another issue during testing involving half-duplex.

The listen() called at the end of begin() causes SoftwareSerial to switch to RX as soon as a write finishes. This is true even if there is no read intended.

This problem can be sporadic, and also only impacts the SoftwareSerial instance which most recently ran begin().

This can cause communications issues for some hardware which is counting on the TX pin keeping the wire at the proper high state between transfers.

An example hardware combination is a BigTreeTech SKR Pro with their BTT 2208 modules.
This combination can read incorrect values, which is resolved by this change.

Before eliminating listen() call for half-duplex:
image

After:
image

fpistm:
Thanks @ABOSTM

PR #753, opened 2019-11-04
Add mechanism to set SoftwareSerial interrupt priority

### LINKED ISSUE #751: Allow override for SoftwareSerial timer ISR priority

The recently added SoftwareSerial library uses the default timer priority for software bit timing. For many processors, this defaults to preemption level 14, which will cause bit timing issues when integrating with systems using other interrupts on the system.

This became apparent when integrating SoftwareSerial with Marlin when testing the following Marlin PR:
https://github.com/MarlinFirmware/Marlin/pull/15655

I suggest the following changes to the SoftwareSerial library. I am willing to contribute these changes, but wanted input on whether these changes were likely to be accepted before creating a PR.

1. Replace TIMER_SERIAL constant with SWSERIAL_TIMER_NUM.
      - Prefix clearly communicates that this constant impacts SoftwareSerial.
      - Providing timer number instead of name simplifies construction of interrupt name at build time.
        - This change could be skipped if there is a good way to determine an interrupt number from a timer base address at runtime (Since TIMx is actually a macro to the base address).
      - \#error can be used to communicate change if TIMER_SERIAL is defined

1. Add SWSERIAL_TIMER_INT_PRIO to override interrupt priority
    - Only change the priority if the constant is defined, else use default.
    - Alternatively, default to highest interrupt priority for best bit-timing performance, and allow users to downgrade priority with constant.

1. Expose additional macros with SWSERIAL prefix, allowing users to customize behavior if needed:
    - SWSERIAL_FORCE_BAUD_RATE (Currently override-able as FORCE_BAUD_RATE)
    - SWSERIAL_OVERSAMPLE
    - SWSERIAL_HALFDUPLEX_DELAY

Working around this in Marlin was quite cumbersome. We were able to get this to communicate by arranging other timer interrupts around the default timer interrupt level, but that solution is not ideal. It likely causes other timing issues due to changes to prioritization among hardware ISRs, etc.

I attempted to set the timer ISR priority from within Marlin, but the change did not persist. I believe that when the timer is reset during a baud rate switch (setSpeed) the timer is being restored to its default priority level.

sjasonsmith:
**Summary**

Add a mechanism to set timer interrupt priority for the timer used By SoftwareSerial.

* [ ] #751 Allow override for SoftwareSerial timer ISR priority

**Validation**

I have reworked an in-development version of Marlin to use this to configure the SoftwareSerial interrupt priority during HAL initialization. Using digital outputs added to the interrupt handlers, I was able to confirm the nesting of multiple preempted timer interrupts, as desired.

In the image below it can be seen that the serial interrupt is the highest of the three priorities, allowing it to preempt both of the other timer interrupts.

Triple_Preemption

Fixes #751

### DIFF
--- a/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -35,6 +35,7 @@
 // Includes
 //
 #include "SoftwareSerial.h"
+#include <timer.h>
 
 #define OVERSAMPLE 3 // in RX, Timer will generate interruption OVERSAMPLE time during a bit. Thus OVERSAMPLE ticks in a bit. (interrupt not synchonized with edge).
 
@@ -96,6 +97,8 @@
 // Statics
 //
 HardwareTimer SoftwareSerial::timer(TIMER_SERIAL);
+const IRQn_Type SoftwareSerial::timer_interrupt_number = static_cast<IRQn_Type>(getTimerUpIrq(TIMER_SERIAL));
+uint32_t SoftwareSerial::timer_interrupt_priority = NVIC_EncodePriority(NVIC_GetPriorityGrouping(), TIM_IRQ_PRIO, TIM_IRQ_SUBPRIO);
 SoftwareSerial *SoftwareSerial::active_listener = nullptr;
 SoftwareSerial *volatile SoftwareSerial::active_out = nullptr;
 SoftwareSerial *volatile SoftwareSerial::active_in = nullptr;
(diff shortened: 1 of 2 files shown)


fpistm:
Hi @sjasonsmith 
Thanks for this PR, during the review, I've though this will be more generic to add a mechanism to set timer interrupt priority for the timer used By HardwareTimer then it is possible to do it for SoftwareSerial.
Could you check  #755 and give us your feedback.

fpistm:
Replaced by #755

PR #755, opened 2019-11-04
[PR 753 rework][SoftwareSerial] Add function to set interrupt priority

### LINKED ISSUE #751: Allow override for SoftwareSerial timer ISR priority

fpistm:
Fixes #751

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -33,7 +33,7 @@
 #define MAX_RELOAD ((1 << 16) - 1) // Currently even 32b timers are used as 16b to have generic behavior
 
 /* Private Variables */
-HardwareTimerObj_t *HardwareTimer_Handle[TIMER_NUM] = {NULL};
+timerObj_t *HardwareTimer_Handle[TIMER_NUM] = {NULL};
 
 IRQn_Type getTimerUpIrq(TIM_TypeDef *tim);
 IRQn_Type getTimerCCIrq(TIM_TypeDef *tim);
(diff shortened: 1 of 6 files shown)


sjasonsmith:
I like making this available for all hardware timers, instead of my change for only SoftwareSerial.

I think this will also solve another issue, which is the need to re-set the priority every time the timer is resumed.

I would like to test this on hardware, but will not have time to do so until tonight. You should have my results in the morning!

fpistm:
@sjasonsmith 
Thanks

About:


> I think this will also solve another issue, which is the need to re-set the priority every time the timer is resumed.

Well this is the case even with this PR because it is done by the `HAL_TIM_Base_MspDeInit` which is called by `HAL_TIM_Base_Init` each time resume is called.
Anyway, compared to the first implementation you've made, this is done only once instead of twice in `setSpeed` (one by the `resume` then call to `NVIC_SetPriority`)

sjasonsmith:
@fpistm 

I reworked this into my Merlin branch, and confirmed that it is working.

I am able to rearrange my three timers (2 HardwareTimer + 1 SoftwareSerial) in any order and verify proper preempting behavior.

I also verified that I can now set the priority once when initializing the HardwareTimer, rather than after resume as was previously required.

fpistm:
@sjasonsmith 
Thanks for tests.
In that case I will merge this one instead of #753.
I will wait @ABOSTM feedback as he is the author of the HardwareTimer library.

LinoBarreca:
@fpistm I highly approve this PR (although it breaks interface compatibility with Arduino's HardwareTimer) it's kinda necessary on STM32 platform.

PR #764, opened 2019-11-08
Add methods to check if there's an IRQ handler

LinoBarreca:
**Summary**

Added methods to check if there's an user-IRQ-handler currently attached.

This PR implements the following **feature**

* [ ] It's easier to understand if you need to attach or detach an IRQ
* [ ] Lower memory usage in complex scenarios

**Why**
..well...first of all...why not?
The HardwareTimer, internally, knows **if** there's a callback. Exposing this to the user makes no harm.
It's a good C++ programming practice to give a way for an object to expose its status. 

Now let's talk about why should it be there.
Imagine a situation in which you want to disable the callback (for example on a channel) leaving the timer running and receiving other interrupts on other channels.

When you have multiple timers and each of them has multiple channels it's difficult to keep track of them all, moreover if they aren't enabled and disabled from the "main" schetch but by some other interrupt handler.

In a situation where you have "multiple sources" of detach/attach this can be cumbersome.
Providing a "zero-cost" (each HardwareTimer knows it already) property which will expose the status of the interrupt handler (present/absent) will make code easier to write and with less bugs.

**Validation**

Should be good.

**Code formatting**

Should be good.

**Closing issues**

Doesn't close any issue but helps programmers to **not** create new ones.

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -715,8 +715,11 @@ bool HardwareTimer::hasInterrupt()
   * @param  channel: Arduino channel [1..4]
   * @retval returns true if a channel compare match interrupt has already been set
   */
-bool HardwareTimer::hasInterrupt(uint32_t channel);
+bool HardwareTimer::hasInterrupt(uint32_t channel)
 {
+  if ((channel == 0) || (channel > (TIMER_CHANNELS + 1))) {
+    Error_Handler();  // only channel 1..4 have an interrupt
+  }
   return callbacks[channel] != NULL;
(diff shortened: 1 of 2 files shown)


LinoBarreca:
NOTE: update to wiki is needed (to inform users that they have the new methods available)

LinoBarreca:
I can do that too..but since it's on a different repository I first wanted to see if this gets merged :)

LinoBarreca:
Ok. I go edit the wiki then, adding the methods as supported from 1.8.0 :)

LinoBarreca:
Done.

Issue #760, opened 2019-11-07
OpenAMP list.h name collision with freertos list.h


adammunich:
When compiling with stm32freertos, list.h in OpenAMP's name over-rides list.h with freertos and compilation fails. Renaming OpenAMP's list.h solves this issue.

fpistm:
Hi @adammunich 
Thanks for reporting this.
Currenlty, I guess I will remove OpenAMP from include list in the platform.txt as it is not used.
/CC @kbumsik

fpistm:
@adammunich 
How you met the issue?
I have no issue. I thought OpenAmp include path was in the platform.txt but this is not.

adammunich:
Some platforms and makefiles I have seen include Arduino_Core_STM32/system/Middlewares/ST recursively, which causes the issue.

kbumsik:
@fpistm Please reconsider deleting. The virtual serial using OpenAMP is completed and it's code cleaning phase. I'm going to make a PR right after #717 is merged.

@adammunich Could I ask what platforms are? I'm personally interested in projects using stm32freertos.
> Arduino_Core_STM32/system/Middlewares/ recursively

This sounds like their responsibility to update the project build though.

fpistm:
> This sounds like their responsibility to update the project build configurations though.

I agree with @kbumsik.
I see no issue with `list.h` as OpenAmp requires to include it thanks: `#include <metal/list.h>`

PR #717, opened 2019-10-22
[New variant] STM32MP157_DK 

kbumsik:
Depends: ~~#605~~ and [Arduino-Tools#47](https://github.com/stm32duino/Arduino_Tools/pull/47) 

Required by: #766 

This PR is to add STM32MP157 discovery boards (both DK1 and DK2).

Since MP1 boards are quite different from other board I explain things to consider to review this PR.

~~For convenience to test, this PR has commits from #605.  You can review commits from 83bc63f.~~ Rebased.

Great reference to read: [STM32CubeMP1 development guidelines](https://wiki.st.com/stm32mpu/wiki/STM32CubeMP1_development_guidelines)

---
### Engineering mode vs Production mode
The M4 coprocessor of STM32MP1 has two operating modes: Engineering mode and Production mode (or normal mode). Engineering mode is for debugging the M4 while disabling the Linux A7 core. In this mode you can use ST-Link to upload and debug the M4, but the M4 doesn't have a flash memory: the firmware will disappear in the next boot. To avoid the confusion to the users I focus on Production mode only and made a necessary tools for Production mode. The new `run_arduino_gen.sh` tool is explained in [Arduino-Tools#47].

In the Production mode, the core clock configuration is done by the Linux/U-Boot host so we need simply ignore related functions like `SystemClock_Config()`. The new `IS_ENGINEERING_BOOT_MODE()` from HAL is used to detect the Production mode.
https://github.com/kbumsik/Arduino_Core_STM32/blob/aebcddff59375b4ac53134babc61dd655b028414/variants/STM32MP157_DK/variant.cpp#L147-L149

### `lock_resources.h` and `lock_resources.c` 
Because GPIO ports are shared with the Linux host, a locking mechanism is needed to avoid race conditions. I copied the source directly from the Cube library: 28898e3 

### GPIOZ and ANA_1:2
The new GPIOZ and ANA_x pins are added. GPIOZ is not special but I modified port/pin definitions since that name was used to indicate the end of GPIO names. The new ANA_1/ANA_2 pins are something like PADC_TEMP but they are physically exposed pins. We don't need to GPIO control for ANA_1/ANA_2 because they are always directly connected to the ADC channels: 442ca6c 

### pseudo-EEPROM using RETRAM
[...]

PR #766, opened 2019-11-08
[MP1] Add RPMsg virtual serial protocol support (VirtIOSerial)

kbumsik:
Depends: 
- [x] #717
- [x] https://github.com/stm32duino/Arduino_Tools/pull/47

~This PR currently includes commits from #717. Please look at commits with [VirtIO] title.~

As #717 is almost being merged I propose VirtIOSerial as promised.
VirtIOSerial.cpp is a virtual serial based on [RPMsg messaging protocol](https://github.com/OpenAMP/open-amp/wiki/RPMsg-Messaging-Protocol), which is built on top of [VirtIO](https://github.com/OpenAMP/open-amp/wiki/OpenAMP-RPMsg-Virtio-Implementation).
To clarify the terminologies [the OpenAMP project](https://github.com/OpenAMP/open-amp/wiki/OpenAMP-Overview) standardizes RPMsg, VirtIO, and etc.

I named it VirtIOSerial because this is the most intuitive name for the most users. The code is based on [OpenAMP_TTY_echo example from ST](https://github.com/STMicroelectronics/STM32CubeMP1/tree/master/Projects/STM32MP157C-DK2/Applications/OpenAMP/OpenAMP_TTY_echo) but it is heavily modified for Arduino Stream/Serial implementation.

The work is almost done and VirtIOSerial currently works well, but they are some minor issues needed to be addressed:
- [x] OpenAMP list.h name may collision with freertos list.h #760
  - I don't know how to check this properly yet.
- [x] Arduino cannot send messages until Linux sends any data first. https://github.com/OpenAMP/open-amp/issues/182
- [x] Clean some debugging messages (`OPENAMP_log_dbg()`, `log_dbg()`....)
- [x] More documentations
- [x] Maybe cleaning IAR/Keil specific directives (`__ICCARM__`, `__CC_ARM`) to reduce code noise?

[This is the updated version of README.md](https://github.com/kbumsik/Arduino_Core_STM32/tree/3b88ae207b3186a1769b84fff60f116109f759bf/variants/STM32MP157_DK)

-----
I'm currently testing with the simple example code below:
```c++
// the setup function runs once when you press reset or power the board
void setup() {
  Serial.begin(115200);
  // Serial7 is the standard Arduino serial pin (D0 and D1)
  Serial7.begin(115200);
  pinMode(LED_BUILTIN, OUTPUT);
}

// the loop function runs over and over again forever
void loop() {
  if(Serial.available()){
      while(Serial.available()){
        Serial.print(Serial.read());
      }
      Serial.println("");
//      Serial7.println("Gotcha");
  } else {
      Serial.println("No data");
//      Serial7.println("no data");
  }
  // Just blink
  digitalWrite(LED_BUILTIN, HIGH);
  delay(1000);
  digitalWrite(LED_BUILTIN, LOW);
  delay(1000);
}
```
You can compile & upload the script, and run `sh run_arduino.sh start`. After that you can test it by running minicom: `$ TERM=xterm minicom -D /dev/ttyRPMSG0`

Edit: there is a better Arduino Sketch to test:
```c++
int available;
char buffer[1024];

unsigned long time = 0;

void setup() {
  Serial.begin(115200);
  pinMode(LED_BUILTIN, OUTPUT);
}

void loop() {
  available = Serial.available();
  while (available > 0) {
    int size = min(available, Serial.availableForWrite());
    Serial.readBytes(buffer, size);
    Serial.write(buffer, size);
    available -= size;
  }

  // Heartbeat. If Arduino stops the LED won't flash anymore.
  if ((millis() - time) > 1000) {
    time = millis();
    digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN));
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/stm32duino/arduino_core_stm32/766)


fpistm:
Great! 
Thanks @kbumsik !

kbumsik:
A workaround for https://github.com/OpenAMP/open-amp/issues/182 has been added.

Now `run_arduino.sh` has serial related commands:

    sh run_arduino.sh monitor
        Monitor data received from the coprocessor via the virtual serial.

    sh run_arduino.sh send-msg <message...>
        Send a message to the coprocessor via the virtual serial.

    sh run_arduino.sh send-file <filename>
        Send a file content to the coprocessor via the virtual serial.

    sh run_arduino.sh minicom
        Launch minicom interactive serial communication program.

fpistm:
> * [ ]  Maybe cleaning IAR/Keil specific directives (`__ICCARM__`, `__CC_ARM`) to reduce code noise?

I think it should be kept to avoid any issue with OpenAMP update.

fpistm:
After a first quick review, here's my first feedback:
* I wonder if this should not be put as a builtin library?
* The VirtIO should be disabled by default
* I see no issue with `list.h` as OpenAmp requires to include it thanks: `#include <metal/list.h>`

kbumsik:
@fpistm 
> * I wonder if this should not be put as a builtin library?
> 
> * The VirtIO should be disabled by default

Do you mean splitting like other libraries such as [STM32duinoBLE](https://github.com/stm32duino/STM32duinoBLE)?

Once of the main idea of this is to take over `Serial`. If a user wants to use Arduino for STM32MP1, they will likely expect that they can remove UART/USB serial communication between the host (it was PC or Raspberry Pi, now it is the A7 Linux) and Arduino, and that is the main benefit of STM32MP1. Therefore VirtIO might be pretty much default  Serial communication for STM32MP1.

Anyway I'm okay with any decision as long as there is a way to take over `Serial`.

kbumsik:
> If you talk about files copied in the core then yes they could be removed else ones in OpenAmp no to avoid any issue with OpenAMP update

Yes I'm talking especially about [`rsc_table.c`](https://github.com/kbumsik/Arduino_Core_STM32/blob/601e343c92969339239dc5ca74705a7e9c4bd672/cores/arduino/stm32/virtio/rsc_table.c). They are just hard to read 😮

fpistm:
@kbumsik 
> Do you mean splitting like other libraries such as [STM32duinoBLE](https://github.com/stm32duino/STM32duinoBLE)?
No, like a builtin library this is not the exactly the same.
I agree about Serial superseded but I think it should be possible. I will test and get back to you if possible.



> Do you mean the default should be `none` like USB CDC does?

Yes

> Yes I'm talking especially about [`rsc_table.c`](https://github.com/kbumsik/Arduino_Core_STM32/blob/601e343c92969339239dc5ca74705a7e9c4bd672/cores/arduino/stm32/virtio/rsc_table.c). They are just hard to read

So yes this could be cleaned

fpistm:
@kbumsik 
Just to keep you informed, I'm currently working to release the 1.8.0 and I will not have the time to well validate this PR. I will validate it for the next release.

kbumsik:
@fpistm Please take your time, the holiday is coming :)

ABOSTM:
@arnopo FYI

ABOSTM:
Hi @kbumsik,
Thanks a lot for your PR, ... and your patience. I know it is quite a long time since you submitted this PR.
But I finally found time to make a deep review and to propose some update.
This is a huge work and it is well working. Thanks.

As I proposed some improvement, instead of making a review like usually, I propose a PR in your fork:
https://github.com/kbumsik/Arduino_Core_STM32/pull/1
I kept your commits as is (except a rebase), and add new commits on top of yours. Of course, discussion is open about those proposed changes.

In addition, I have few questions/remarks:

1) In variants/STM32MP157_DK/README.md , you can remove limitation :
`* Currently there is no easy way for communication between the Linux host and Arduino coprocessor.`
And on the contrary add it as an implemented feature.

2) Question about `RPMSG_BUFFER_SIZE`:
```
#ifdef RPMSG_BUFFER_SIZE
#error "RPMSG_BUFFER_SIZE is already defined"
#else
```
Why raising an error? why not allowing user to redefine this value ? Is it linked to definition in linux driver ?
In this case, a comment would be welcome.  

3) Is there any rational behind
`#define VIRTIO_BUFFER_SIZE (RPMSG_BUFFER_SIZE * 2) ` ?
Is it made to deal with some kind of Half buffer management? link to below code:
 ```
 if (prev_write_available < RPMSG_BUFFER_SIZE
      && virtio_buffer_write_available(&_VirtIOSerialObj.ring) >= RPMSG_BUFFER_SIZE) {
    MAILBOX_Notify_Rx_Buf_Free();
  }
```
If so, maybe you can add a comment before `VIRTIO_BUFFER_SIZE `definition and replace `RPMSG_BUFFER_SIZE `by `VIRTIO_BUFFER_SIZE/2` in the checks:
```
if (prev_write_available < (VIRTIO_BUFFER_SIZE / 2)
      && virtio_buffer_write_available(&_VirtIOSerialObj.ring) >= (VIRTIO_BUFFER_SIZE / 2))
```
  
4) Add some header to new functions:
Even if not always respected, we try to use a template from HAL, for example:
```
/**
  * @brief  Checks if there's an interrupt callback attached on Capture/Compare event
  * @param  channel: Arduino channel [1..4]
  * @retval returns true if a channel compare match interrupt has already been set
  */
```

**Comment by fpistm** [comment]
Hi @kbumsik,
I know this PR is old and probably you switch on other stuff.
Anyway, could you gave us a feedback or if you really could not give it, we probably do a clean PR with @ABOSTM update/review then close this one?
Let us know what you prefer.
Thanks in advance
BR

**Comment by kbumsik** [comment]
Hi @ABOSTM @fpistm,

I was just focused on other stuff until this week and now I can look into this PR again.

For 1 and 4, this PR currently needs a little bit of cleanup after receiving feedbacks of the initial design, so I will do them after the design is confirmed. Also note that this PR is a redesign of [ST's OpenAMP_TTY example](https://github.com/STMicroelectronics/STM32CubeMP1/tree/master/Projects/STM32MP157C-DK2/Applications/OpenAMP/OpenAMP_TTY_echo). So there are some uncleaned traces from the example code yet.

For 2, 

> Why raising an error? why not allowing user to redefine this value ? Is it linked to definition in linux driver ?

yes. The Linux kernel specified the buffer size as 512. You can see this here: [https://elixir.bootlin.com/linux/v5.5.2/source/drivers/rpmsg/virtio_rpmsg_bus.c#L137](https://elixir.bootlin.com/linux/v5.5.2/source/drivers/rpmsg/virtio_rpmsg_bus.c#L137)

It was also a guard to make sure the definition is not overritten by rpmsg_virtio.h. I was originally tried to use rpmsg_virtio.h for RPMSG_BUFFER_SIZE but including this header causes problems that comes from the included headers of rpmsg_virtio.h. So I created 

virtio_config.h and included a checker like this to make sure virtio_config.h is effective.

I think it is fine to remove this guard to let developers to adjust the size, as long as the dev don't forget to increase the buffer in the Linux kernel as well.

Maybe we can give `#warn` message instead of throwing an `#error`.

For 3,

> Is it made to deal with some kind of Half buffer management?

No, also replacing it by `VIRTIO_BUFFER_SIZE/2`does not make sense. Whatever the size of `VIRTIO_BUFFER_SIZE` is it must call `MAILBOX_Notify_Rx_Buf_Free()` when the buffer has more than `RPMSG_BUFFER_SIZE` available. 

The design of the ring buffer is influenced by [cdc-queue.h](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/stm32/usb/cdc/cdc_queue.h). The ring buffer is introduced to make the original ST's example code mentioned above more general-purpose, and it increases responsiveness of the Linux's synchronous rpmsg channel but not Arduino.

    virtio_rpmsg_bus.c                           virtqueue      vring   rpmsg_virtio.c
     ========   <-- new msg ---=============--------<------             ==========
    ||      || rvq (rx)    || IPCC CHANNEL 1 ||  svq (tx_vq) -> vring0 ||        || 
    ||  A7  ||  ------->-------=============--- buf free-->            ||   M4   ||
    ||      ||                                                         ||        ||
    ||master||  <-- buf free---=============--------<------            || slave  ||
    ||      || svq (tx)    || IPCC CHANNEL 2 ||  rvq (rx_vq) -> vring1 ||(remote)|| 
     ========   ------->-------=============----new msg -->             ==========

Let's look at the diagram here, the virtio / rpmsg / OpenAMP is implemented using IPCC interrupts and shared memory ([MCU SRAM3 region](https://wiki.st.com/stm32mpu/wiki/STM32MP15_RAM_mapping)).

Imagine A7 send a message to /dev/ttyRPMSG0. A7 sends a packet (max 512 bytes - 16 bytes header) to rvq virtqueue and trigger "new msg" interrupt. The message is stored in vring1 buffer which is associated with rvq. When copying vring1 to our Arduino rx ringbuffer is done, M4 need to trigger "buf free" interrupt back to A7 by calling `MAILBOX_Notify_Rx_Buf_Free()`. A7 sends another packet if the message is longer than 512 -16 bytes.

The important thing is that A7's sending the message to /dev/ttyRPMSG0 is fully synchronous operation. The A7 process is blocked until it receives "buf free" interrupt. If M4 don't trigger it in timely manner the A7 process will be blocked indefinitely until it gets kernel timeout error in the worst case. So my idea is to increase the ring buffer to `RPMSG_BUFFER_SIZE * 2` so that it can immediately send back "buf free" events until 2 full packet transmissions even if the Arduino application is not consuming the RX messages right now. We can adjust the size of the ring buffer but the code in question should be kept IMO.

I will look into your PR and will leave more feedbacks :)

**Comment by ABOSTM** [comment]
Hi @kbumsik 
Thanks for your answer.
For 2, I am fine to keep this guard, just add a comment to explain it comes from linux setting.

**Comment by arnopo** [comment]
Hi @kbumsik,
I dobn't review all you patches in details, so don't hesitate if o missed something... 

> ```
> virtio_rpmsg_bus.c                           virtqueue      vring   rpmsg_virtio.c
>  ========   <-- new msg ---=============--------<------             ==========
> ||      || rvq (rx)    || IPCC CHANNEL 1 ||  svq (tx_vq) -> vring0 ||        || 
> ||  A7  ||  ------->-------=============--- buf free-->            ||   M4   ||
> ||      ||                                                         ||        ||
> ||master||  <-- buf free---=============--------<------            || slave  ||
> ||      || svq (tx)    || IPCC CHANNEL 2 ||  rvq (rx_vq) -> vring1 ||(remote)|| 
>  ========   ------->-------=============----new msg -->             ==========
> ```
> 
> Let's look at the diagram here, the virtio / rpmsg / OpenAMP is implemented using IPCC interrupts and shared memory ([MCU SRAM3 region](https://wiki.st.com/stm32mpu/wiki/STM32MP15_RAM_mapping)).
> 
> Imagine A7 send a message to /dev/ttyRPMSG0. A7 sends a packet (max 512 bytes - 16 bytes header) to rvq virtqueue and trigger "new msg" interrupt. The message is stored in vring1 buffer which is associated with rvq. When copying vring1 to our Arduino virtio_buffer is done (thie is what `rxCallback` does), M4 need to trigger "buf free" interrupt back to A7 by calling `MAILBOX_Notify_Rx_Buf_Free()`. A7 sends another packet if the message is longer than 512 -16 bytes.
> 
> The important thing is that A7's sending the message to /dev/ttyRPMSG0 is fully synchronous operation. The A7 process is blocked until it receives "buf free" interrupt. 

Does it means that you have only one RPmsg buffer (VRING_NUM_BUFFS = 1)? i don't see that you update this value in the resource table, could you point it out?
The A7 process client of /dev/rpmsgTTY0 is not blocked even if if there is no buffer free as it uses rpmsg_trysend function. But perhaps you are using a old version of the code ( please check that your linux code is aligned wit this code https://github.com/STMicroelectronics/linux/blob/v4.19-stm32mp/drivers/rpmsg/rpmsg_tty.c#L138).
Anyway it is possible to send message until there is no more buffers free ( number of buffer defined in stm32cube resource table by VRING_NUM_BUFFS). 
another point:  "buff free" IPCC interrupt is not mandatory to unblock. the RPMsg and the mailbox usage are decorrelated. 
- The IPCC interrupt is a doorbell to inform the remote processor on a buffer action. 
- The RPmsg/virtio manage the buffers in memory.
So in theory a firmware can manage RPMsg using polling method without any mailbox/doorbell.
The free of the buffer is automatic on return from  the rpmsg ept callback. 

>If M4 don't trigger it in timely manner the A7 process will be blocked indefinitely until it gets kernel timeout error in the worst case. So my idea is to increase virtio_buffer to `RPMSG_BUFFER_SIZE * 2` so that it can immediately send back "buf free" events until 2 full packet transmissions even if the Arduino application is not consuming the RX messages right now. We can adjust the size of the virtio_buffer but the code in question should be kept IMO.

IMO if you faced this kind of behavior it is more an issue in the project design. 
- could share  a trace related to the timeout you mention?
- if M4 doesn't trig in time that probably means that the M4 is not able to treat message in time. The question is why?
- If you don't want to consume the messages when you receive it, either you make a copy or you have to  adjust the numbers of RPMsg buffers, not the size...
FYI zero copy enhancement should be integrated soon in openAMP: https://github.com/OpenAMP/open-amp/pull/162.

Regards
Arnaud


> 
> I will look into your PR and will leave more feedbacks :)

**Comment by kbumsik** [comment]
@ABOSTM 
> For 2, I am fine to keep this guard, just add a comment to explain it comes from linux setting.

I see, let me have about a couple of days. I need to set up my workbench to test your merged commits.

Hi @arnopo,
It's great we have a kernel dev here :) And thanks for your detailed explanation. It helped my understandings of rpmsg.

> Does it means that you have only one RPmsg buffer (VRING_NUM_BUFFS = 1)? i don't see that you update this value in the resource table, could you point it out?

No, it's `VRING_NUM_BUFFS = 4` as original. As you guessed I haven't touched the resource table because I'm not confident about the consequences of modifying it. My explanation here is based on my observation, not the theory sides. 

> The A7 process client of /dev/rpmsgTTY0 is not blocked even if if there is no buffer free as it uses rpmsg_trysend function. But perhaps you are using a old version of the code ( please check that your linux code is aligned wit this code https://github.com/STMicroelectronics/linux/blob/v4.19-stm32mp/drivers/rpmsg/rpmsg_tty.c#L138).

You are right, I'm testing with an early version (around July of the last year) of openstlinux and the kernel source is different...it uses `rpmsg_send()` rather than your `rpmsg_trysend()` https://github.com/STMicroelectronics/meta-st-stm32mp/blob/b5ef9c5a747edd7d15b7ac260153c9844347fd6f/recipes-kernel/linux/linux-stm32mp/4.19/4.19.9/0013-ARM-stm32mp1-r0-rc1-REMOTEPROC-RPMSG.patch#L2395

So I am using a kernel version earlier than your patch [`rpmsg: tty: use trysend to not block the tty write`](https://github.com/STMicroelectronics/linux/commit/09ffbf10bf1eeaf2a46eac6d37cfb8bee9abf4ea#diff-40c2ca8f25f46c3ba8fc5eabab1a1b79).

> another point: "buff free" IPCC interrupt is not mandatory to unblock. the RPMsg and the mailbox usage are decorrelated.

But I definitely saw the blocking behavior until I triggered the IPCC interrupt. Do you also think this might be related to my old kernel mentioned above?

> could share a trace related to the timeout you mention?
Sure, if I still have the problem after using the newer kernel. Just in case, do you have any preferred tracing tool to work with other than ftrace/uftrace?

---

Based on your response, it feels like I definitely need to try it again with the latest kernel, and there might be possible design improvements using `VRING_NUM_BUFFS` instead of `virtio_buffer` ring buffer. Let me have some time to play around...

**Comment by ABOSTM** [comment]
@arnopo,
FYI Arduino_Core_STM32 is currently on Cube version 1.1
Is it necessary/interesting to update to latest version 1.2 ?

**Comment by arnopo** [comment]
> > Does it means that you have only one RPmsg buffer (VRING_NUM_BUFFS = 1)? i don't see that you update this value in the resource table, could you point it out?
> 
> No, it's `VRING_NUM_BUFFS = 4` as original. As you guessed I haven't touched the resource table because I'm not confident about the consequences of modifying it. My explanation here is based on my observation, not the theory sides.

seems that VRING_NUM_BUFFS = 16 (https://github.com/stm32duino/Arduino_Core_STM32/pull/766/files#diff-52bdd02a93c61dc71972b188b05e133dR160)

> 
> > The A7 process client of /dev/rpmsgTTY0 is not blocked even if if there is no buffer free as it uses rpmsg_trysend function. But perhaps you are using a old version of the code ( please check that your linux code is aligned wit this code https://github.com/STMicroelectronics/linux/blob/v4.19-stm32mp/drivers/rpmsg/rpmsg_tty.c#L138).
> 
> You are right, I'm testing with an early version (around July of the last year) of openstlinux and the kernel source is different...it uses `rpmsg_send()` rather than your `rpmsg_trysend()` https://github.com/STMicroelectronics/meta-st-stm32mp/blob/b5ef9c5a747edd7d15b7ac260153c9844347fd6f/recipes-kernel/linux/linux-stm32mp/4.19/4.19.9/0013-ARM-stm32mp1-r0-rc1-REMOTEPROC-RPMSG.patch#L2395
> 
> So I am using a kernel version earlier than your patch [`rpmsg: tty: use trysend to not block the tty write`](https://github.com/STMicroelectronics/linux/commit/09ffbf10bf1eeaf2a46eac6d37cfb8bee9abf4ea#diff-40c2ca8f25f46c3ba8fc5eabab1a1b79).
> 
> > another point: "buff free" IPCC interrupt is not mandatory to unblock. the RPMsg and the mailbox usage are decorrelated.
> 
> But I definitely saw the blocking behavior until I triggered the IPCC interrupt. Do you also think this might be related to my old kernel mentioned above?

To be able to receive new IPCC interrupt you need first to clear the flag. if you don'y modify the mailbox_ipcc ?c file it it the case(  HAL_IPCC_NotifyCPU is used to clear the IPCC IRQ flag).
Then the rpmsg_send is blocking until new buffer is free(https://elixir.bootlin.com/linux/v5.6-rc1/source/drivers/rpmsg/virtio_rpmsg_bus.c#L596). So yes to unblock the M4 has to release a buffer.
SO IMO the good question in this case is why M4 does not free it in time ( 15 second timeout)

> 
> > could share a trace related to the timeout you mention?
> 
> Sure, if I still have the problem after using the newer kernel. Just in case, do you have any preferred tracing tool to work with other than ftrace/uftrace?
>

you can add dynamic debug traces on Linux side:
echo -n 'file stm32-ipcc.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file remoteproc*.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_rpmsg_bus.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_ring.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file rpmsg_tty.c +p' > /sys/kernel/debug/dynamic_debug/control
 
Regards
Arnaud

**Comment by arnopo** [comment]
> @arnopo,
> FYI Arduino_Core_STM32 is currently on Cube version 1.1
> Is it necessary/interesting to update to latest version 1.2 ?

There are few improvements and fixes around the OpenAMP middleware so could be interesting...
The v1.2 release is on going, But i think that the code is now available on github

**Comment by kbumsik** [comment]
> SO IMO the good question in this case is why M4 does not free it in time ( 15 second timeout)

Because this is an Arduino library that tartgets entry-level programmers, not for a time-critical embedded system. This port is not only just for developers but it may run exisiting Arduino Sketches out-of-the-box (for example, I'm planning to run Firmata after this PR).

So I mention that case because it would be great if it is a little bit tolerant to errors by devs (or existing sketches) errors that do not read the incomming messages on time. Regarding it is Arduino library haning an A7 process seems a little bit harsh to me.

**Comment by ABOSTM** [comment]
Hi @kbumsik,
Thanks for this update, nice job.
Only one remark about logging:
the function `log_buff()` should not be placed in Print.cpp, 
In the same way `uart_debug_write()` is not in Print.cpp.
Instead it could be placed either in VirtIOSerial.h/.c or in new file in OpenAmp directory ...

**Comment by kbumsik** [comment]
Hi @ABOSTM,
Thanks for the quick feedback before I write a comment :) I was actually about to ask about the location of `log_buff()` too. I moved it to virtio_log.c.

I also wrote a preliminary version of the documentation.

At this moment, I found a deadlock condition within `metal_mutex_acquire() ` block when read() and write() simultaneously because I handle Virtring in the IPCC interrupts. I'm fixing my design right now.

I also found some interesting behavior of Linux rpmsg driver. I will post a trace when things get more clear to me...

File: cores/arduino/stm32/virtio/virtio_config.h [hunk 1]
```diff
@@ -0,0 +1,10 @@
+#ifndef __VIRTIO_CONFIG_H
+#define __VIRTIO_CONFIG_H
+
+#ifdef RPMSG_BUFFER_SIZE
+#error "RPMSG_BUFFER_SIZE is already defined"
```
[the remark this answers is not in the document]
kbumsik: Oh, you might be looking at the old changes. I already made it configurable ;)

kbumsik:
@ABOSTM I fixed the errors but I cannot fix some of them in /system/Middlewares/OpenAMP because these should be kept as-is.

Anyway, the work is pretty much done and it's time to do a final review and squash & cleanup commits:smile:. After I have better understandings, I resolved the issues related to deadlock condition and the timeout errors. The issues were mainly caused by my design miss that I tried to handle message in an ISR. Thanks @arnopo for your help!

@arnopo I think I found a regression with the latest OpenSTLinux kernel. Here I attach my trace logs and explanations.

I tested with the two versions of OpenSTLinux
* openstlinux-4.19-thud-mp1-19-02-20 (the old version)
* openstlinux-20-02-19 (the latest version)

The traces I enabled:
```
echo -n 'file stm32-ipcc.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file remoteproc*.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_ring.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_rpmsg_bus.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file rpmsg_tty.c +p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_rpmsg_bus.c format "rpmsg_virtio TX:" -p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file virtio_rpmsg_bus.c format "rpmsg_virtio RX:" -p' > /sys/kernel/debug/dynamic_debug/control
echo -n 'file rpmsg_tty.c format "rpmsg_tty_cb00" -p' > /sys/kernel/debug/dynamic_debug/control
```
Note that I disabled some traces that have big impacts on the real-time which make it harder to reproduce.

Also I tested with an Arduino Sketch that has a best performance, just echo every loop:
```c++
int available;
char buffer[1024];

unsigned long time = 0;

void setup() {
  Serial.begin(115200);
  pinMode(LED_BUILTIN, OUTPUT);
}

void loop() {
  available = Serial.available();
  while (available > 0) {
    int size = min(available, Serial.availableForWrite());
    Serial.readBytes(buffer, size);
    // This will make linux kernel debugging harder to read
//    Serial.write(buffer, size);
    available -= size;
  }

  // Heartbeat. If Arduino stops due to a deadlock
  // the LED won't flash anymore.
  if ((millis() - time) > 1000) {
    time = millis();
    digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN));
  }
}
```

Firstly I loaded the Arduino firmware:
```bash
$ sh run_arduino_Blink.sh start
```
Then I dd /dev/ttyRPMSG0:
```bash
$ dd if=/dev/zero of=/dev/ttyRPMSG0 bs=496 count=16
16+0 records in
16+0 records out
7936 bytes (7.9 kB, 7.8 KiB) copied, 0.00187093 s, 4.2 MB/s
```
This works fine with the both kernels, until I send 16 messages (which is the number of buffer).
```bash
# The old kernel
$ dd if=/dev/zero of=/dev/ttyRPMSG0 bs=496 count=50
50+0 records in
50+0 records out
24800 bytes (25 kB, 24 KiB) copied, 15.6219 s, 1.6 kB/s

# The new kernel
$ dd if=/dev/zero of=/dev/ttyRPMSG0 bs=496 count=40

(...blocked forever...more than 24 hours...)

(...Ctrl + C to stop)
32+0 records in
31+0 records out
15376 bytes (15 kB, 15 KiB) copied, 270.045 s, 0.1 kB/s
```
When dd command sends more than 16 messages, you will see the blocking behavior, which is expected though.
Note that enable dynamic tracing slows down write performance so that does not fill all 16 buffers at a time, thus makes dd command often successful without blocking when enabled.

With the old kernel, dd command is blocked for some seconds when it uses all 16 buffers, and then try it again. See the performance is decreased from 4.2MB/s => 1.6kB/s.

The problem is, the latest kernel blocks forever (more than 24 hours, I tested it) and never retry again. I understand the blocking behavior of the kernel, although not the ideal. However I believe the complete blocking behavior shown with the newer kernel is clearly a regression.

I attach log here, you will be interested in the message after "No more buffers in queue".
[debug-log-old.log](https://github.com/stm32duino/Arduino_Core_STM32/files/4284757/debug-log-old.log)
[debug-log-new.log](https://github.com/stm32duino/Arduino_Core_STM32/files/4284758/debug-log-new.log)

cores/arduino/stm32/virtio/virtio_config.h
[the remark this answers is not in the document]
ABOSTM: Right, it is even an old comment !
Sorry

kbumsik:
In the last commit, I added both virtio options to CI. I added them because this is less used option so I wanted to prevent a silent break in the future. I am not sure if I configured it correctly.

fpistm:
We will see result soon 😉 
As a first look it seems correct.

fpistm:
Hi @kbumsik 
is it OK for final review or do you plan again some small update ? :wink:

kbumsik:
@fpistm 
Yes, you can proceed to a final review. There might be only one thing to be updated but I think it can be added as a single commit later (meaning it should not be squashed) or just another tiny PR. Please see the next comment.

cores/arduino/stm32/OpenAMP/mbox_ipcc.c
```diff
@@ -0,0 +1,203 @@
...
+        core_debug("Running vring1 (ch_2 new msg)\n");
+        /**
+         * This calls rpmsg_virtio_rx_callback(), which calls virt_uart rx callback
+         * RING_NUM_BUFFS times at maximum.
+         */
+        rproc_virtio_notified(vdev, VRING1_ID);
+        msg_received_ch2 = RX_NO_MSG;
+
+        /* The OpenAMP framework does not notify for free buf: do it here */
+        rproc_virtio_notified(NULL, VRING1_ID);
```
kbumsik: I copied this line from the CubeMP1 examples but I suspect it should be removed. I'm waiting for a response from a ST employee to confirm if it is okay to remove or it should be replaced by anything else: https://community.st.com/s/question/0D53W0000009CfFSAU/a-question-regarding-openamp-rpmsg-example

fpistm: OK. I guess @arnopo could answer to that 😉 
@arnopo any update on this, please?
Thanks in advance.

arnopo: Hi,
i confirm yes you can remove it.

fpistm: Thanks @arnopo
@kbumsik could you proceed to this update then we will able to merge.

kbumsik: @fpistm Sure :)
@arnopo Thanks for confirming it. By the way, could I ask you any update regarding my report above?

arnopo:
Minors comments then LGTM

kbumsik:
Fixed 😛  Thanks for the review @arnopo 😉

kbumsik:
@fpistm I made a mistake in the last commit "mbox_ipcc.c: rx_status_t -> mbox_status_t", which is fixed in the following fixup commit. However, the CI somehow passed the build of the faulty commit. I guess the CI ignored board options. It should be fixed but I don't know how to locally test the CI though.

fpistm:
I will check.

fpistm:
I've found the issue. I've introduced a regression with https://github.com/stm32duino/Arduino_Core_STM32/pull/976 😖 
Fixed in https://github.com/stm32duino/Arduino_Core_STM32/pull/982

kbumsik:
@fpistm Thanks for approval. Sure, I kept them unclean to preserve the history but it's time to squash.
But, wow it's 58 commits. Allow me a day 😛

fpistm:
> @fpistm Thanks for approval. Sure, I kept them unclean to preserve the history but it's time to squash.
> But, wow it's 58 commits. Allow me a day 😛

For sure 😉

kbumsik:
```
/github/home/.arduino15/packages/STM32/hardware/stm32/1.8.0/CI/build/examples/BareMinimum/BareMinimum.ino:46:10: error: 'class VirtIOSerial' has no member named 'setRx'
   46 |   Serial.setRx(PIN_SERIAL_RX);
      |          ^~~~~
```
Well, it turns out to be it was a great idea to enable CI for VirtIOSerial.
So, should we just exclude VirtIO in that code block like what [USBD_USE_CDC did](https://github.com/stm32duino/Arduino_Core_STM32/blob/f14d8b92e65a6cb7db45eb0d8dd442ee9e6b5c5c/CI/build/examples/BareMinimum/BareMinimum.ino#L45), or do we need to add mocks for `setRx`/`setTx`?

fpistm:
> Well, it turns out to be it was a great idea to enable CI for VirtIOSerial.

Right, that's why I've updated also the CI sketch else this would not be raised.
That's fine for me. No reason to mock those methods.

Thanks for the update and the clean PR 👍 🥇

Issue #763, opened 2019-11-08
HardwareTimer incomplete?


HendryKaak:
Hey, 
I've been struggling with the implementation of the current `HardwareTimer` class with a custom stm32f401 board and would like to share a few points that caused some problems here and there.

What i've been trying to do is using the HardwareTimer class to setup a PWM output for a buzzer and used a specific timer channel to do so. But the first problem I encountered was that the timer or channel was not disabled when the `Pause()` function is called. Looking at the code I found out that in the HAL `stm32f4xx_hal_tim.h`, the timer would only be disabled when all timer channels are disabled: https://github.com/stm32duino/Arduino_Core_STM32/blob/1e2a5985dabb20a7581a4c284537c3031d968779/system/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_hal_tim.h#L1022-L1031

So disabling the timer never happens because the `HAL_TIM_OC_Stop` macro is never called with the correct timer channel in the `pause()` function. 

https://github.com/stm32duino/Arduino_Core_STM32/blob/1e2a5985dabb20a7581a4c284537c3031d968779/cores/arduino/HardwareTimer.cpp#L101-L109

Only the interrupts are being stopped over there as you can see, but never the channel itself and thus the timer is never stopped.

To resolve this:
* At least give the option to disable a timer channel, but leave the timer enabled when another Input Capture/Output Capture channel is still in use, so that concurrent channel can be used.
* It would probably be a good idea to be able to disable independent channels in place of enabling all of the channels at once (also for concurrent timer channel usage).
* I also wonder about the purpose of `pause()` and `resume()`. Essentially, these functions really stop/disable and start/enable the timer, where pause would suggest it just stops the timer without reconfiguring anything (in particular, I would not expect pause to change the current pin output), Perhaps these functions should be renamed and/or better documented? Maybe real pause/resume methods should be added?

Another thing that I found questionable is the `setMode()` function. AFAIK is that the channels modes are first stored in RAM, but only written to registers when `resume()` is called. This means that the only way to change a channel's configuration is by restarting the timer. It might be feasible to sometimes reconfigure a channel while the timer is running (an obvious usecase is to disable or re-enable a single channel without affecting the others, but also changing other setttings on the fly would make sense). Maybe setMode should just apply the changes directly? As long as the timer is disabled, there should be no harm in applying the channel settings, I believe?

fpistm:
@ABOSTM 
Any thought on that?

[...]

PR #767, opened 2019-11-08
Flexible interrupts handling

LinoBarreca:
**Summary**

This PR modifies (internally) how timers are started so that it will be possible to attach an interrupt to an already running timer.

* [x] Bug: If the timer is initially started without a user-interrupt (it's rather a callback from an interrupt) it's impossible to assign one without stopping and restarting the timer. If the interrupt is assigned before starting the timer, instead the user can attach and detach them freely and have the callback still invoked. This was undocumented before today (I altered the wiki adding this warning but then I realized that a patch which handled it was better than a warning)
* [x] Bug: The user calls `detachInterrupt` and he expects to not have something besides systick that can interrupt his code. His code is still interrupted by HardwareTimer IRQ handler (and this is bad for time critical code, for example when the user, using the callback, wants to give a clock to something) 
* [x] Feature: you can now attachInterrupt and detachInterrupt anytime

**Motivation**
Improvement in usage and predictable behavior if the user didn't read the "warning" in the documentation.

**Validation**
Sorry it's untested at the moment. But, in theory, it should work fine.

**Code formatting**
I changed everything that travis complained about.

**Closing issues**
Closes the issues I mentioned in "bugs".
I didn't open them because I preferred to see if I could use the same time to find a solution instead of creating a problem :)

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -632,7 +632,7 @@ void HardwareTimer::setInterruptPriority(uint32_t preemptPriority, uint32_t subP
   */
 void HardwareTimer::attachInterrupt(void (*callback)(HardwareTimer *))
 {
-  if (callbacks[0] == null) { //if there's no callback the ISR here is not enabled
+  if (callbacks[0] == NULL) { //if there's no callback the ISR here is not enabled
     __HAL_TIM_ENABLE_IT(&(_timerObj.handle), TIM_IT_UPDATE);
   }
 
@@ -665,7 +665,7 @@ void HardwareTimer::attachInterrupt(uint32_t channel, void (*callback)(HardwareT
     Error_Handler();  // only channel 1..4 have an interrupt
   }
 
-  bool mustActivateISR = callbacks[channel] == null; //no user callback means ISR is disabled
+  bool mustActivateISR = (callbacks[channel] == NULL); //no user callback means ISR is disabled
 
   callbacks[channel] = callback;
 
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -179,57 +179,29 @@ void HardwareTimer::resumeChannel(uint32_t channel)
   if (IS_TIM_PWM_MODE(_channelOC[channel - 1].OCMode)) {
     HAL_TIM_PWM_ConfigChannel(&(_timerObj.handle), &_channelOC[channel - 1], timChannel);
 
-    if ((channel < (TIMER_CHANNELS + 1)) && (callbacks[channel] != NULL)) {
-      // Only channel 1..4 can have interruption
 #if defined(TIM_CCER_CC1NE)
-      if (isComplementaryChannel[channel]) {
-        HAL_TIMEx_PWMN_Start_IT(&(_timerObj.handle), timChannel);
-      } else
+    if (isComplementaryChannel[channel]) {
+      HAL_TIMEx_PWMN_Start(&(_timerObj.handle), timChannel);
+    } else
 #endif
-      {
-        HAL_TIM_PWM_Start_IT(&(_timerObj.handle), timChannel);
-      }
-    } else {
-#if defined(TIM_CCER_CC1NE)
-      if (isComplementaryChannel[channel]) {
-        HAL_TIMEx_PWMN_Start(&(_timerObj.handle), timChannel);
-      } else
-#endif
-      {
-        HAL_TIM_PWM_Start(&(_timerObj.handle), timChannel);
-      }
+    {
+      HAL_TIM_PWM_Start(&(_timerObj.handle), timChannel);
     }
   } else if (IS_TIM_OC_MODE(_channelOC[channel - 1].OCMode)) {
     HAL_TIM_OC_ConfigChannel(&(_timerObj.handle), &_channelOC[channel - 1], timChannel);
 
-    if ((channel < (TIMER_CHANNELS + 1)) && (callbacks[channel] != NULL)) {
-      // Only channel 1..4 can have interruption
 #if defined(TIM_CCER_CC1NE)
-      if (isComplementaryChannel[channel]) {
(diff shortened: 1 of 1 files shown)


cores/arduino/HardwareTimer.cpp
```diff
@@ -660,7 +632,11 @@ void HardwareTimer::setInterruptPriority(uint32_t preemptPriority, uint32_t subP
   */
 void HardwareTimer::attachInterrupt(void (*callback)(HardwareTimer *))
 {
-  callbacks[0] = callback;
+  if (callbacks[0] == NULL) { //if there's no callback the ISR here is not enabled
+    __HAL_TIM_ENABLE_IT(&(_timerObj.handle), TIM_IT_UPDATE);
```
ABOSTM: There are some use case were this will not work:
__HAL_TIM_ENABLE_IT() will modify hardware register to enable interrupt, but in the following example case, 
```
  MyTim->setMode(channel, TIMER_OUTPUT_COMPARE_PWM1, pin);
  // MyTim->setPrescaleFactor(8); // Due to setOverflow with MICROSEC_FORMAT, prescaler will be computed automatically based on timer input clock
  MyTim->setOverflow(100000, MICROSEC_FORMAT); // 10000 microseconds = 10 milliseconds
  MyTim->setCaptureCompare(channel, 50, PERCENT_COMPARE_FORMAT); // 50%
  MyTim->attachInterrupt(Update_IT_callback);
  MyTim->attachInterrupt(channel, Compare_IT_callback);
  MyTim->resume();
```
the timer peripheral clock (the clock of register) is enabled only on attachInterrupt() (thanks to HAL_TIM_Base_Init()). Thus it is not possible to write register on __HAL_TIM_ENABLE_IT().
For information, I inherit this behavior, from danieleff implementation (see https://github.com/stm32duino/Arduino_Core_STM32/issues/146#issuecomment-503980512)

LinoBarreca: Sorry @ABOSTM I think I didn't understood....Can't you write a register if the timer is disabled? What prevents it?
The only thing which could make this not work is that HAL_TIM_Base_Init overwrites the DIER register....but from what I saw (before making the changes and the PR) it doesn't... TIM_Base_SetConfig doesn't touch that register...
So actually you set the DIER register in the attachInterrupt and then you start the timer....which should behave accordingly to what the registers say, generating the interrupt.

What am I missing?

ABOSTM: @LinoBarreca , the point is to provide the clock of let say TIM1 peripheral. Without this clock, register are not accessible.  This is done by writing some RCC (Reset and Clock Controller). This is currently done thanks to:
HAL_TIM_Base_MspInit() --> enableTimerClock()  from timer.c
And this is called from HAL_TIM_Base_Init().

fpistm:
Hi @LinoBarreca 
Is there any reason to close it?

LinoBarreca:
@fpistm yes. Let me explain it.
Actually, despite the "logic" for me was good, it seems I made a mistake.
I made the assumption that the timer registers could be set even when the timer were not "active" (before resume).

@ABOSTM told me that, if the timer doesn't have a clock, you can't access its registers. I assume he is right because he is way more expert than me.

what was the logic behind my PR:
- user creates an instance of HardwareTimer()
- user calls `attachInterrupt()` to say "I want to receive the interrupt when the timer will be active"
- attachInterrupt remembers the user callback and sets a registers on the timer (this fails according to Alexandre)
- user calls `resume()`
- resume gives the clock to the timer and starts it. Since the above call failed the register in the timer will not have the information "I must trigger the IRS" and the user callback never gets called.

I thought my logic was working because with the `setOverFlow()` (actually called before `resume()`) we still access a register (ARR) of an "unstarted timer" so I thought that should work the same with the DIER register (called by `attachInterrupt()` ). I didn't really understand why, before resume, we can write to ARR but not DIER but if he says so I believe him (and I closed the PR because it worked in my mind but it will not work in real life)

If I misunderstood something (or you could clarify my doubt), however, let me know :)

ABOSTM:
Hi @LinoBarreca , Your understanding of the issue is correct,
nevertheless, as I said in #763, I am thinking about enabling timer peripheral clock at HardwareTimer object creation.
Thus if we go for this implementation, your pull request makes sense. So we can keep this PR opened

LinoBarreca:
ok. Thanks Alexandre but...can you please explain me why, with the timer clock disabled, we can write to ARR but not to DIER?

ABOSTM:
From hardware point of view it is not possible to write ARR or PSC registres (or any register) with timer clock disabled.
Concerning SetOverflow and setPrescaler, if you look carefully, there is a trick:
that is right thas we try to update registers (in case clock is already present, it allows to update ARR and PSC on the fly while timer is running), but in case clock is not present, we also save the value in structure so that it will be configured later (during the resume):
```
  __HAL_TIM_SET_AUTORELOAD(&_HardwareTimerObj.handle, ARR_RegisterValue);
  _HardwareTimerObj.handle.Init.Period = ARR_RegisterValue;
```

LinoBarreca:
> 
> ```
>   __HAL_TIM_SET_AUTORELOAD(&_HardwareTimerObj.handle, ARR_RegisterValue); <-this fails if the timer doesn't have a clock (=not yet started) but succeeds if the timer is running.
>   _HardwareTimerObj.handle.Init.Period = ARR_RegisterValue; <- this saves the value and sets it again at timer start
> ```

now I understood. You were very kind.

I guess changing the init structure to store also the interrupts requested is not to consider?
So you wouldn't need two methods, per timer mode, to start the timer (with _IT and without _IT) but just one...
and to maintain backwards compatibility we could keep the _IT methods and in the implementation we could set .Init.FireInterrupts = true and then call the "base" start in which we set all the registers *including* DIER...

fpistm:
I add label _OnHold_ as it depends on #763

fpistm:
Hi @LinoBarreca 
@ABOSTM implemented #763 and included your PR during this. 
I you can provide your feedback, this will be fine, then we could close this one as it is now duplicated.
Thanks in advance

fpistm:
I close this one has integrated in #806.

PR #771, opened 2019-11-13
[Timer] Add getHandle function

HendryKaak:
This enables users to config timers using the HAL layers for more advanced timer
settings when needed. This is especially useful when multiple timer channels
require separate configurations under the same timer instance.

This feature is related to issue #763

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -713,6 +713,18 @@ void HardwareTimer::refresh()
 }
 
 /**
+  * @brief  Return the timer object handle object for more advanced setup
+  * @note   Using this function and editing the Timer handle is at own risk! No support will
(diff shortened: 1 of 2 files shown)


fpistm:
Note that users who use this function and modify the Timer handle will use it at their own risk. No support will be provided if Hardware Timer does not work as expected after Timer handle custom change.

HendryKaak:
Wh00ps, OK sorry about that. Is this better @fpistm?

>Note that users who use this function and modify the Timer handle will use it at their own risk. No support will be provided if Hardware Timer does not work as expected after Timer handle custom change.

I added an extra note in the cpp file, hopefully that is enough.

fpistm:
Fine for the comment except the astyle issue linked to it :wink: (Trailing space)
https://travis-ci.com/stm32duino/Arduino_Core_STM32/jobs/256462599

PR #774, opened 2019-11-15
MP1 update

fpistm:
This PR:
* Fix issue with ADC (in both mode: engineering or production)
* Small cleanup
* Update the README according to @FCHSTM 
 https://github.com/stm32duino/Arduino_Core_STM32/pull/717#discussion_r346368969

~I will probably also update the generated script name with the sketch name.
@kbumsik as discussed in https://github.com/stm32duino/Arduino_Tools/pull/47#issuecomment-554214941, do you think it is correct to use sed in the generate part of the script?~ Done

~Maybe I will add the~ MP1 Cube update 1.1.1:
 * HAL and CMIS device not changed
 * OpenAMP updated to v1.0.3


kbumsik:
@fpistm 
> do you think it is correct to use sed in the generate part of the script?

Yes. I tested it.

```diff
diff --git a/run_arduino_gen.sh b/run_arduino_gen.sh
index 3b67ecba..2704625f 100755
--- a/run_arduino_gen.sh
+++ b/run_arduino_gen.sh
@@ -219,10 +219,11 @@ case "$1" in
     TERM=xterm minicom -D $RPMSG_DIR
     ;;
   generate)
-    generate_packaged_script $2 $3
-    echo "$(readlink -f "$3") generated successfully."
+    output=$(echo "$3" | sed 's/\.ino\././g')
+    generate_packaged_script $2 $output
+    echo "$(readlink -f "$output") generated successfully."
     echo "This file should be uploaded manually by SCP, SFTP, Kermit, or etc."
-    echo "Then run \"sh ./$(basename $3) start\" command in the board's console."
+    echo "Then run \"sh ./$(basename $output) start\" command in the board's console."
     echo "For detailed instructions, please visit:"
     echo "  https://github.com/stm32duino/Arduino_Core_STM32/tree/master/variants/STM32MP157_DK/README.md"
     ;;
```
should work, tested in Linux and Windows.

variants/STM32MP157_DK/variant.cpp
```diff
@@ -254,6 +254,19 @@ void SystemClock_Config(void)
...
+  PeriphClkInit.AdcClockSelection = RCC_ADCCLKSOURCE_PER;
+  if (HAL_RCCEx_PeriphCLKConfig(&PeriphClkInit) != HAL_OK) {
+    Error_Handler();
+  }
+
+  /* Configure VREFBUF */
+  __HAL_RCC_VREF_CLK_ENABLE();
+  HAL_SYSCFG_VREFBUF_HighImpedanceConfig(SYSCFG_VREFBUF_HIGH_IMPEDANCE_DISABLE);
+  HAL_SYSCFG_EnableVREFBUF();
+
```
kbumsik: Just in case, this block will be ignored since we do not use Engineering Boot mode.

fpistm: But in case we use Engineering Boot it will not be

variants/STM32MP157_DK/README.md
```diff
@@ -96,7 +96,7 @@ There are additional pins for LEDs and buttons.
 
 ## Uploading
 
-As mentioned above `run_arduino.sh` file should be uploaded manually in order to work. There are many ways to upload the file.
+As mentioned above `run_arduino.sh` file have to be uploaded manually to the Linux file system in order to be executed on the board directly through the Linux console. There are many ways to upload the file.
```
kbumsik: What about using easier words like "From the PC to the board"?
```suggestion
As mentioned above `run_arduino.sh` file have to be uploaded manually from the PC to the board's Linux file system in order to be executed on the board directly through the board's Linux console. There are many ways to upload the file.
```

variants/STM32MP157_DK/variant.cpp
kbumsik: OK, it should be fine then. I just wanted to point out this block is not tested for this moment ;)

FCHSTM: It is good to put that  code under Engineering boot as this stuff is handled by Cortex A processor. that's why , you need in that case to well define ADC ressource in resource manager node in DT, so that OpenSTLinux does it. 
However, not familiar with Arduino distribution but not sure if SYSCFG/VREBUF config is relevant in SySTemclock Config. for sure , these kind of settings are also applicable for STM32 MCU families. In Cube example, this kind of code would be in HAL_PPP_msp_init (with PPP = ADC in our case)

fpistm: This could be in `HAL_ADC_msp_init`

variants/STM32MP157_DK/README.md
```diff
@@ -13,49 +13,49 @@ Because every OS may have different software configurations (especially Device T
...
-    <Arduino build output path>/run_arduino.sh generated successfully.
+    <Arduino build output path>/run_arduino_<sketch name>.sh generated successfully.
     This file should be uploaded manually by SCP, SFTP, Kermit, or etc.
-    Then run "sh ./run_arduino.sh start" command in the board's console.
+    Then run "sh ./run_arduino_<sketch name>.sh start" command in the board's console.
     For detailed instructions, please visit:
         https://github.com/stm32duino/Arduino_Core_STM32/tree/master/variants/STM32MP157_DK/README.md
 
-In this example, the user **must** upload `<Arduino build output path>/run_arduino.sh` file manually. Uploading instruction is described later in the [Uploading](#Uploading) section. 
+In this example, the user **must** upload `<Arduino build output path>/run_arduino_<sketch name>.sh` file manually. Uploading instruction is described later in the [Uploading](#Uploading) section.
```
kbumsik: In my opinion, the new name of `<Arduino build output path>/run_arduino_<your project name>.sh` reduces readability of the document and users may take more time to notice what `<Arduino build output path>/run_arduino_<your project name>.sh` is. So I think adding example samples of `<Arduino build output path>/run_arduino_<your project name>.sh` will help the users. What about adding the example after this paragraph?
```markdown
`<Arduino build output path>/run_arduino_<your project name>.sh` looks like this:

* `C:/Users/kbumsik/AppData/Local/Temp/arduino_build_668148/run_arduino.Blink.ino.sh` (For Windows)
* `/tmp/arduino_build_668148/run_arduino_Blink.ino.sh` (For Linux/macOS)
```

variants/STM32MP157_DK/README.md
```diff
@@ -13,49 +13,49 @@ Because every OS may have different software configurations (especially Device T
 * [STM32 MPU OpenSTLinux Distribution]
 * [Balena OS]
 
-Note that the first two OSes should select `stm32mp157c-dk2-m4-examples-sdcard` boot mode (or device tree configuration) on boot time. `stm32mp157c-dk2-sdcard` boot mode is also known to work but it is not guaranteed. See [ST Wiki page on boot mode] for more detail. If you are using it for your own OS, see the [Linux Device Tree considerations](#Linux-Device-Tree-considerations) section.
+__Note that the first two OSes should select `stm32mp157c-dk2-m4-examples-sdcard` boot mode (or device tree configuration) on boot time.__ `stm32mp157c-dk2-sdcard` boot mode is also known to work but it is not guaranteed. See [ST Wiki page on boot mode] for more detail. If you are using it for your own OS, see the [Linux Device Tree considerations](#Linux-Device-Tree-considerations) section.
```
kbumsik: Because our generate script is quite generic we can add this note?
```suggestion
Other distributions not listed here may run Arduino but they are not tested. __Note that the first two OSes should select `stm32mp157c-dk2-m4-examples-sdcard` boot mode (or device tree configuration) on boot time.__ `stm32mp157c-dk2-sdcard` boot mode is also known to work but it is not guaranteed. See [ST Wiki page on boot mode] for more detail. If you are using it for your own OS, see the [Linux Device Tree considerations](#Linux-Device-Tree-considerations) section.
```

fpistm: Note added

kbumsik: Great, thanks for updating! :)

variants/STM32MP157_DK/README.md
```diff
@@ -13,49 +13,63 @@ Because every OS may have different software configurations (especially Device T
...
-    <Arduino build output path>/run_arduino.sh generated successfully.
+    <Arduino build output path>/run_arduino_<sketch name>.sh generated successfully.
     This file should be uploaded manually by SCP, SFTP, Kermit, or etc.
-    Then run "sh ./run_arduino.sh start" command in the board's console.
+    Then run "sh ./run_arduino_<sketch name>.sh start" command in the board's console.
     For detailed instructions, please visit:
         https://github.com/stm32duino/Arduino_Core_STM32/tree/master/variants/STM32MP157_DK/README.md
 
-In this example, the user **must** upload `<Arduino build output path>/run_arduino.sh` file manually. Uploading instruction is described later in the [Uploading](#Uploading) section. 
+`<Arduino build output path>/run_arduino_<your project name>.sh` looks like this for *Blink* example:
```
kbumsik:
```suggestion
`<Arduino build output path>/run_arduino_<sketch name>.sh` looks like this for *Blink* example:
```

variants/STM32MP157_DK/README.md
```diff
@@ -13,49 +13,63 @@ Because every OS may have different software configurations (especially Device T
...
-    sh run_arduino.sh stop
+    sh run_arduino_<sketch name>.sh stop
         Stop the coprocessor.
 
-    sh run_arduino.sh restart
+    sh run_arduino_<sketch name>.sh restart
         Restart the coprocessor.
 
-See the source code for the full help page and the more details about [run_arduino.sh].
+See the source code for the full help page and the more details about [run_arduino_<sketch name>.sh].
```
kbumsik: It looks like `<>` in link does not work, rendering like `run_arduino_`.
```suggestion
See the source code for the full help page and the more details about [run_arduino_gen.sh].
```

variants/STM32MP157_DK/README.md
```diff
@@ -152,7 +166,7 @@ And then the Device Tree should enable TIM1 for the coprocessor, although this d
 [Balena OS]: https://github.com/kbumsik/balena-st-stm32mp
 [ST Wiki page on boot mode]: https://wiki.st.com/stm32mpu/wiki/STM32CubeMP1_Package#Getting_started_with_STM32CubeMP1_Package
 
-[run_arduino.sh]: https://github.com/stm32duino/Arduino_Tools/blob/master/linux/run_arduino_gen.sh
+[run_arduino_<sketch name>.sh]: https://github.com/stm32duino/Arduino_Tools/blob/master/linux/run_arduino_gen.sh
```
kbumsik:
```suggestion
[run_arduino_gen.sh]: https://github.com/stm32duino/Arduino_Tools/blob/master/run_arduino_gen.sh
```

variants/STM32MP157_DK/README.md
fpistm: Welcome 😉 
Thanks too for your review. 👍

fpistm:
Well,
I guess this PR could be merged when CI build will be ended.

variants/STM32MP157_DK/README.md
fpistm: I've updated and add a sentence about the gen script to be more precise.

variants/STM32MP157_DK/README.md
fpistm:
```
See the source code run_arduino_gen.sh for the full help page and the more details about the run_arduino_<sketch name>.sh generated.

run_arduino_gen.sh is the shell script that produces a copy of the script called run_arduino_<sketch name>.sh but with the sketch binary self-contained.
```

variants/STM32MP157_DK/README.md
fpistm: fixed

PR #780, opened 2019-11-18
Add printf to print class

MCUdude:
This PR uses [Adafruits printf implementation](https://github.com/adafruit/ArduinoCore-samd/blob/361481d34d0e6a9d5b2cd074ee0a990d78df3396/cores/arduino/Print.cpp#L189-L197). 
For those who didn't know, printf is super useful when you want to format your output. When printf is a part of the Print class, all libraries that inherits the print class (such as Serial, LiquidCrystal, SoftwareSerial + other LCD libraries).

This PR makes this possible:
```cpp
Serial.printf("Milliseconds since start: %d\n", millis());
```

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -196,12 +196,22 @@ size_t Print::println(const Printable &x)
   return n;
 }
 
-void Print::printf(const char format[], ...)
+void Print::printf(const char *format, ...)
 {
-  char buf[PRINTF_BUF];
+  char buf[PRINTF_BUFFER];
   va_list ap;
   va_start(ap, format);
   vsnprintf(buf, sizeof(buf), format, ap);
+  write(buf);
+  va_end(ap);
+}
+
+void Print::printf(const __FlashStringHelper *format, ...)
+{
+  char buf[PRINTF_BUFFER];
+  va_list ap;
+  va_start(ap, format);
+  vsnprintf(buf, sizeof(buf), (const char *)format, ap);
   write(buf);
   va_end(ap);
 }
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -196,6 +196,16 @@ size_t Print::println(const Printable &x)
   return n;
 }
 
+void Print::printf(const char format[], ...)
+{
+  char buf[PRINTF_BUF];
+  va_list ap;
+  va_start(ap, format);
+  vsnprintf(buf, sizeof(buf), format, ap);
+  write(buf);
+  va_end(ap);
+}
+
 // Private Methods /////////////////////////////////////////////////////////////
 
 size_t Print::printNumber(unsigned long n, uint8_t base)
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -211,7 +211,7 @@ void Print::printf(const __FlashStringHelper *format, ...)
   char buf[PRINTF_BUFFER];
   va_list ap;
   va_start(ap, format);
-  vsnprintf(buf, sizeof(buf), (const char *)format, ap);
+  vsnprintf_P(buf, sizeof(buf), (const char *)format, ap);
   write(buf);
   va_end(ap);
 }
--- a/cores/arduino/Print.h
+++ b/cores/arduino/Print.h
@@ -32,7 +32,6 @@
 #define OCT 8
 #define BIN 2
 
-class __FlashStringHelper;
 
 #ifndef PRINTF_BUFFER
 #define PRINTF_BUFFER 80
--- a/cores/arduino/Print.h
+++ b/cores/arduino/Print.h
@@ -23,7 +23,6 @@
 #include <inttypes.h>
 #include <stdio.h> // for size_t
 #include <stdarg.h> // for printf
-#define PRINTF_BUF 80
 
 #include "WString.h"
 #include "Printable.h"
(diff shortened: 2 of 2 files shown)


MCUdude:
Any updates on this? I forgot to mention that this does not add any extra flash usage when not used in the user program.

fpistm:
@MCUdude 
I will review that when got a free time slot.
I have to check Arduino compatibility and what is involved and potential regression if any.

uzi18:
does it support F() macro?

MCUdude:
The current PR does not. However, I can add it. I will probably use the code from the [Teensy core files](https://github.com/PaulStoffregen/cores/blob/73ea157600a7082686d9cc48786a73caa7567da9/teensy3/Print.cpp#L94-L114), since this supports the F() macro as well. 

If you and @fpistm want, I can test this code and update the PR with these changes afterward.

uzi18:
nice to have possibility to redefine buf size

MCUdude:
Actually, the printf code from Teensy didn't work very well when printing simultaneously on all serial ports. Adafruit's code work well, so I'll stick with this.

It doesn't seems like the F() macro doesn't do much really. It calls the PSTR() macro, which is just a dummy. But sure, There's no problem supporting F macros inside printf, but it won't affect the memory usage.

cores/arduino/Print.cpp
```diff
@@ -196,6 +196,26 @@ size_t Print::println(const Printable &x)
...
+  write(buf);
+  va_end(ap);
+}
+
+void Print::printf(const __FlashStringHelper *format, ...)
+{
+  char buf[PRINTF_BUFFER];
+  va_list ap;
+  va_start(ap, format);
+  vsnprintf(buf, sizeof(buf), (const char *)format, ap);
```
fpistm:
```suggestion
  vsnprintf_P(buf, sizeof(buf), (const char *)format, ap);
```

cores/arduino/Print.h
```diff
@@ -31,6 +32,12 @@
 #define OCT 8
 #define BIN 2
 
+class __FlashStringHelper;
```
fpistm:
```suggestion

```

cores/arduino/Print.h
```diff
@@ -31,6 +32,12 @@
...
+
```
fpistm:
```suggestion
```

fpistm:
This addition requires to be documented in the Wiki:
 * Using `printf` will add additional code space. See example hereafter.
 * Have to not be used in third party library else this library would not be compatible across arch.
 * Use an hidden temporary buffer which can be redefined 😕 

Code example:
```C++
#include <Streaming.h>

const char * lorem = "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.";

void setup() {
  Serial.begin(9600);
}

void loop() {
  static uint32_t i = 0;
  // 1
  Serial.print("Iter ");
  Serial.print(i++);
  Serial.print(" --> ");
  Serial.println(lorem);
  // 2
  Serial.printf("Iter %i --> %s\n", i++, lorem);
  // 3
  Serial << "Iter " << i++ << " --> " << lorem << endl;
}
```

### Size:
 * Current usage (1):
```
Sketch uses 9192 bytes (1%) of program storage space. Maximum is 524288 bytes.
Global variables use 320 bytes (0%) of dynamic memory, leaving 130752 bytes for local variables. Maximum is 131072 bytes.
```

 * `Print::printf` usage (2)
```
Sketch uses 11600 bytes (2%) of program storage space. Maximum is 524288 bytes.
Global variables use 448 bytes (0%) of dynamic memory, leaving 130624 bytes for local variables. Maximum is 131072 bytes.
```
 * Streaming usage (3)
```
Sketch uses 9176 bytes (1%) of program storage space. Maximum is 524288 bytes.
Global variables use 320 bytes (0%) of dynamic memory, leaving 130752 bytes for local variables. Maximum is 131072 bytes. 
```

As you can see Streaming usage is the best way.

MCUdude:
I tried the online Resolve and commit functionality for the first time. Hopefully it is OK now

MCUdude:
Cool! Will this be available in release 1.8.0?

fpistm:
> Cool! Will this be available in release 1.8.0?

Well thinking about that, I really not like the buffer.
I'm currently testing the usage of `vdprintf`.

fpistm:
I merge this and will provide a hardening to remove `PRINTF_BUFFER` drawback.

MCUdude:
Sure! I do agree that having a hardcoded buffer isn't optimal, but at least it is a start. I'm not sure how relevant it is, but here's [how it is implemented on 8-bit AVRs](https://github.com/MCUdude/MiniCore/blob/6a3054280e9934919fd2fa1f03a20e8927de5693/avr/cores/MCUdude_corefiles/Print.cpp#L201-L228), without any hardcoded buffer.

gigaj0ule:
Where can I find vdprintf?

```
system32/Arduino_Core_STM32/cores/arduino/Print.cpp:267:16: error: 'vdprintf' was not declared in this scope; did you mean 'vwprintf'?
```

It tried to include <stdio.h> but still no luck.

fpistm:
> 

Could you be more precise? What is your setup? Board selected? Core version? Arduino IDE version? Sketch?...

Because I never seen this issue. vdprintf is provided by newlib.

PR #787, opened 2019-11-21
Add generic F401Rx variant

MCUdude:
This is my take on a new generic variant for the STM32F401Rx chip family. This pinout can easily be adapted to other variants in the future, and I'm planning to do this if this PR goes through.


fpistm:
As a first look, I really dislike move all pin_arduino.* files.
As I expected the number of duplicated files/code are impressive and will growth with the new boards addition.
Maintains this would become hard.

Why not, for example add some guard:

```
#ifndef PIN_A0 
...
#endif
```
and define `Pin_A0` in the `variant.h`.

or add a custom define :
```
#ifndef CUSTOM_PIN_ARDUINO
...
#endif
```

Anyway, as I said, it's a first feeling and I have to go further. :wink:

MCUdude:
I'm willing to do whatever it takes for this PR to be merged 🙂  I'm busy tonight, but will continue working on this tomorrow. I'll see what I can do about the global pins_arduino.c/h files.

But apart from this, what do you think? Positive?

fpistm:
I guess removing Analog pins restrictions is really fine but as said moving the pin_arduino files is not the good way, see the diff numbers:
```diff
+27,286
-0
```
while 99% are the same

MCUdude:
I'm absolutely open to other ways of achieving this.

However, I do need full control over :
* all the analog pin definitions
* all the PIN_SPI and PIN_WIRE macros
* the ATEMP, VREF and AVBAT macros
* the analogInputToPinName macro
* Get rid of some static asserts that prevents this from being compiled

fpistm:
> * all the analog pin definitions

As I said could be achieved easily with guard

> * all the PIN_SPI and PIN_WIRE macros

This is already the case

> * the ATEMP, VREF and AVBAT macros

Why this is hardware specific and not linked to a pin and already managed.

> * the analogInputToPinName macro

Same, with a guard or a define.

> * Get rid of some static asserts that prevents this from being compiled

Static assert will be updated if needed and are present to avoid any issue with wrong definition.

MCUdude:
>> the ATEMP, VREF and AVBAT macros

> Why this is hardware-specific and not linked to a pin and already managed.

To prevent compiler warnings I had to change the numbers a bit on these macros. I defined ATEMP as `NUM_DIGITAL_PINS + 2` instead of `NUM_DIGITAL_PINS + 1`, because +1 is already used in analogInputToDigitalPin. I didn't use `NC` in analogInputToDigitalPin because the compiler complained about comparing different types and enums. 

I'll see if I can roll back the commit where all the pins_arduino.c/h files were added.

fpistm:
> I didn't use `NC` in analogInputToDigitalPin because the compiler complained about comparing different types and enums.

Right and it is fine the compiler raise warning. This force to use cast if really needed.
Note that NC is used to know if the pin is correct or not so you probably introduce regression.

All you mentioned as a requirement is de facto linked how pins are managed. This simply means all have to be aligned with the new implementation. Which is normal. :wink:

MCUdude:
> Right and it is fine the compiler raise warning. This force to use cast if really needed.
Note that NC is used to know if the pin is correct or not so you probably introduce regression.

Sorry, but I'm having difficulties understanding what you mean here. Compiler warnings are there for a reason; the code isn't written the way it should. I always tries my best to get rid of warnings, because it "pollutes" the compiler output. IMO the core files should not generate warnings at all. This way, if the users see some red text in the output window, it is not caused by the core files, but rather the user program or included libraries. What I can do however is to just return (p) if the pin isn't correct. this is [what's originally done anyways](https://github.com/stm32duino/Arduino_Core_STM32/blob/2560d7b8090e491c3f853c116077abcc5f0b5e6b/cores/arduino/pins_arduino.h#L242), and does not generate any warnings. The analogInputToDigitalPin is rarely used by user programs anyways; It's mainly used within the core files.

> All you mentioned as a requirement is de facto linked how pins are managed. This simply means all have to be aligned with the new implementation. Which is normal.

Hmm. I don't understand everything here, but I assume you think the requirements I have for this are OK. Am I right?

fpistm:
> > Right and it is fine the compiler raise warning. This force to use cast if really needed.
> > Note that NC is used to know if the pin is correct or not so you probably introduce regression.
> 
> Sorry, but I'm having difficulties understanding what you mean here. Compiler warnings are there for a reason; the code isn't written the way it should. I always tries my best to get rid of warnings, because it "pollutes" the compiler output. IMO the core files should not generate warnings at all. This way, if the users see some red text in the output window, it is not caused by the core files, but rather the user program or included libraries. What I can do however is to just return (p) if the pin isn't correct. this is [what's originally done anyways](https://github.com/stm32duino/Arduino_Core_STM32/blob/2560d7b8090e491c3f853c116077abcc5f0b5e6b/cores/arduino/pins_arduino.h#L242), and does not generate any warnings. The analogInputToDigitalPin is rarely used by user programs anyways; It's mainly used within the core files.

Well forgot that and focus only on the original requirement: `remove analog Pins definitions limitations`

> Hmm. I don't understand everything here, but I assume you think the requirements I have for this are OK. Am I right?

Yes and no. I just want to say that "your" requirements are all dependent and related to each other.
If you change the way analog are defined so you need to update all related stuff...
So given a full control on `ATEMP`,  for example, is not (from my point of view) a requirement as it is not a "pin" but an internal ADC. It has nothing to do in the variant folder just have to be updated in a way it does not disturbed analog pins management.

MCUdude:
@fpistm how about now? Now only minimal changes to the global pins_arduino.h file are done. It still works as it should, and does not affect any other variants.

fpistm:
@MCUdude 
This is better than the previous version :wink:
Less risky and easier to maintain, it focuses on what you wanted to do.

MCUdude:
So what changes do I have to do in order tofix the Astyle formatting and hopefully get this merged?

fpistm:
See details of the check you will see what goes wrong.
You can use Astyle to format properly the PR.
See https://github.com/stm32duino/wiki/wiki/Astyle

About merge, I will have to review deeply and check all uses case.
And the commits needs to be reworked.
Adding a file  in a commit then delete it in one other is not good practice.
Each commit have to be for one goal and should be build able --> atomic
See the [CONTRIBUTING.md](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/CONTRIBUTING.md)

MCUdude:
> And the commits needs to be reworked.

To be honest, I'm absolutely no git expert, _at all_, sorry about that. Would you like it to be one commit where I add the variant files, one commit where I add the new variant to the boards.txt and one commit where I change the global pins_arduino.h file? In that case I'll have to roll back everything and start clean.

fpistm:
You do not need to be a git expert.
I advise you to read the contributing.md, mainly this part:
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/CONTRIBUTING.md#5-rebasing-pull-requests

Your PR should only contains 2 commits:
 * 002dc26d9fdeec2edea701d2940c763d51981a25 should be the first with a commit message more detailed (why you do this?)
 * all others should be squashed in 1 commit

MCUdude:
Thanks for the heads-up, hopefully, it's OK now.

boards.txt
```diff
@@ -1051,6 +1087,10 @@ GenF4.menu.upload_method.hidMethod.upload.tool=hid_upload
 GenF4.menu.upload_method.hidMethod.build.flash_offset=0x4000
 GenF4.menu.upload_method.hidMethod.build.bootloader_flags=-DBL_HID -DVECT_TAB_OFFSET={build.flash_offset}
 
+GenF4.menu.upload_method.MassStorage=Mass Storage
```
fpistm: This is not a valid upload method for Generic menu.

cores/arduino/pins_arduino.h
```diff
@@ -60,108 +60,161 @@ enum {
...
                "First analog pin value (A0) must be greater than or equal to NUM_ANALOG_INPUTS");
+#endif
 
 // Defined for backward compatibility with Firmata which unfortunately use it
+#ifndef AEND
 #define AEND (NUM_ANALOG_FIRST+NUM_ANALOG_INPUTS)
+#endif
 
 #if NUM_ANALOG_INPUTS > 0
+#ifndef PIN_A0
```
fpistm: This could be add on the previous line and avoid adding an `#endif`

Same for all following `#ifndef PIN_Ax`

cores/arduino/pins_arduino.h
```diff
@@ -239,10 +292,14 @@ uint32_t pinNametoDigitalPin(PinName p);
 // Convert an analog pin number to a digital pin number
 #if defined(NUM_ANALOG_INPUTS) && (NUM_ANALOG_INPUTS>0)
 // Used by analogRead api to have A0 == 0
+#ifndef analogInputToDigitalPin
 #define analogInputToDigitalPin(p)  (((uint32_t)p < NUM_ANALOG_INPUTS) ? (p+A0) : p)
+#endif
 #else
+#ifndef analogInputToDigitalPin
```
fpistm: This have to be removed as this never happen.
If we are here then there is no Analog pin.
```suggestion
```

variants/Generic_F401Rx/ldscript.ld
```diff
@@ -0,0 +1,186 @@
...
+** OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**
+*****************************************************************************
+*/
+
+/* Entry Point */
+ENTRY(Reset_Handler)
+
+/* Highest address of the user mode stack */
+_estack = 0x20018000;    /* end of RAM */
```
fpistm: The linker script have to be updated to use FLASH and RAM size definition.
Example:
https://github.com/stm32duino/Arduino_Core_STM32/blob/2560d7b8090e491c3f853c116077abcc5f0b5e6b/variants/BLACK_F407XX/ldscript.ld#L55-L65

variants/Generic_F401Rx/ldscript.ld
```diff
@@ -0,0 +1,186 @@
...
+/* Generate a link error if heap and stack don't fit into RAM */
+_Min_Heap_Size = 0x200;;      /* required amount of heap  */
+_Min_Stack_Size = 0x400;; /* required amount of stack */
+
+/* Specify the memory areas */
+MEMORY
+{
+FLASH (rx)      : ORIGIN = 0x08000000, LENGTH = 512K
```
fpistm: dito

variants/Generic_F401Rx/variant.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+// ----------------------------------------------------------------------------
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/**
+  * @brief  System Clock Configuration
+  *         The system Clock is configured as follow :
+  *            System Clock source            = PLL (HSE, external clock)
```
fpistm: Remarks:
I advise to set HSI instead of HSE this would be more generic.
Or maybe try to set HSE and if not present fallback to HSI?

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,171 @@
...
+//                 |--------|-----------|----------|------------------------|-----------|
+#define PH0  48 // |        |           |          |                        | OSC_IN    |
+#define PH1  49 // |        |           |          |                        | OSC_OUT   |
+//                 |--------|-----------|----------|------------------------|-----------|
+
+// This must be a literal
+#define NUM_DIGITAL_PINS        50
+#define NUM_ANALOG_INPUTS       16
+#define NUM_ANALOG_FIRST        0
+#define AEND                    36
```
fpistm: I guess this could be removed as AEND is no more used in Firmata.

cores/arduino/pins_arduino.h
```diff
@@ -60,108 +60,161 @@ enum {
...
+
 _Static_assert(NUM_ANALOG_INPUTS <= MAX_ANALOG_INPUTS,
                "Core NUM_ANALOG_INPUTS limited to MAX_ANALOG_INPUTS");
+#ifndef PIN_A0
 _Static_assert(NUM_ANALOG_FIRST >= NUM_ANALOG_INPUTS,
                "First analog pin value (A0) must be greater than or equal to NUM_ANALOG_INPUTS");
+#endif
 
 // Defined for backward compatibility with Firmata which unfortunately use it
+#ifndef AEND
```
fpistm: I guess this could be removed as it is no more used in Firmata since a while.

variants/Generic_F401Rx/variant.cpp
MCUdude: Would definitely try to set HSE and instead fallback to HSI if no oscillator is attached. Do you have the code for this?

variants/Generic_F401Rx/variant.h
MCUdude: OK. I'll remove AEND. But how should all these small changes be committed? One commit for each file? One commit for everything?

variants/Generic_F401Rx/ldscript.ld
MCUdude: I was hoping this pinout was generic enough to be used with F401RB, F401RC, F401RD and F401RE. Is there any way we can make the script "detect" what target is selected in boards.txt? would it be better if I created a generic pinout for each *RB -> *RE variant?

variants/Generic_F401Rx/variant.cpp
fpistm: Well if one of the `HAL_RCC_*Config` failed then fallback to HSI.
Here how mbed do it:
https://github.com/ARMmbed/mbed-os/blob/cdcef4babe74e9d4cf3ea8262b359be3eeed2970/targets/TARGET_STM/TARGET_STM32F1/TARGET_BLUEPILL_F103C8/device/system_clock.c#L131-L231

variants/Generic_F401Rx/variant.h
fpistm: Well for example in this case:
`AEND` has been added to this new file, so simply remove it from the commit which add it.
For the `AEND `in the `pin_arduino.h` then a single commit is a good idea and set a proper commit message to explain that it is removed as Firmata does not use it anymore.

cores/arduino/pins_arduino.h
MCUdude: Do you mean like
```c
#if !defined(PIN_A0) && NUM_ANALOG_INPUTS > 0
```

variants/Generic_F401Rx/variant.h
MCUdude: So one separate commit that removes AEND from pins_arduino.h and variant.h with a proper description that describes why the are being removed?

cores/arduino/pins_arduino.h
fpistm: Yes but I'm wrong. I've just figure out that 
`static const uint8_t A0 = PIN_A0;`
was made here.
So, this is correct, my bad.

boards.txt
MCUdude: Because no other programmers than the on-board STlink on the Nucleo boards have this option? I've been using this option while I've worked on the new pinout.

variants/Generic_F401Rx/variant.h
fpistm: > So one separate commit that removes AEND from pins_arduino.h ~and variant.h~ with a proper description that describes why they are being removed?

For variant.h this have to be in the same commit than 93a67409eed6c9e620e93bfd58b604cabc6e7c90

MCUdude: Small comments all around. I'll sum it up in a separate comment when we have figured this out before I start fixing these small issues. At this point, I'm too afraid of pushing something that might pollute the history...

variants/Generic_F401Rx/ldscript.ld
MCUdude: Any comments on this? This is my biggest worry. It's also important that it will work in other environments such as PlatformIO too.

fpistm: Well I've already answered.
Use those definitions:
```
LD_MAX_DATA_SIZE
LD_FLASH_OFFSET
LD_MAX_SIZE 
```

Then for all will be properly set.
The linker script is pre-processed and then all values will be set to the linked variant selected based on definition in the boards.txt.
https://github.com/stm32duino/Arduino_Core_STM32/blob/2560d7b8090e491c3f853c116077abcc5f0b5e6b/platform.txt#L41

boards.txt
fpistm: Yes. I know you use the Nucleo to validate but this is not a valid options for all other boards.
Keep in mind this is a _Generic_ Core and It have to be as generic as possible. If I add this option and user select the Black F407, this will not be functional... then I will have to explain why it is not... again and again.
If an user wants some customization for his convenience then he can:
 * https://github.com/stm32duino/wiki/wiki/Custom-board-based-on-a-core
 * use platform.local.txt
 * ...

MCUdude:
Some new commits. I've tried to add descriptive comments.
The last commit is also properly tested on actual hardware. I've tried to remove (and short out) the external oscillator, and it switches to its internal oscillator on boot. The code you linked in for this was for an F103C8, but I found the correct code for it a little deeper into the repo.

MCUdude:
Done! Anything else you'd like to be changed?

fpistm:
In fact, I will split this PR in 2 parts:
* Analog pins definition rework
* Add generic F401Rx variant

MCUdude:
> Analog pins definition rework

Can you point out to me what exact commits/commit hashes you'd want me to remove/roll back?

fpistm:
I would not have the variant addition in the same PR than the analog pin definitions rework.

MCUdude:
I would assume you'd mean all changes I've done to pins_arduino.h? Or just the AEND part of it?

The reason why I did it this way is that the generic F401Rx pinout relies on the changes done to pins_arduino.h.

I'm not trying to be rude or anything, but it has been pretty time consuming for me with everything that I've put into this PR. It would be great if you were more specific about what you want and what you don't want to be included when I first opened this PR. Anyways, I'm slowly learning from my mistakes, and hopefully, next time can just provide a clean PR without having to re-do commits again and again. Should be better for us both 😉

MCUdude:
@fpistm would be great if you could reply to my previous post. Hopefully, I can finish this up today 👍

fpistm:
@MCUdude 
Sorry, I have a huge workload.
Well to answer you,  it's also _pretty time consuming_ for me. I've spent several time also to answer you and give advice. 
If we get back 5 days ago your PR made a lot of useless changes 
```diff
+27,286
-0
```
It is now better formed. :wink:

All what I advised are described in the Contributing.md (best practice, commits rules, ...).
As an example, it still a commit for removing the MassStorage upload method. Well, why this commit? This PR add and remove it, so this should not be there. Again simply squash this commit to the one which add it then this will clean the PR instead of having annoying codes changes.
Maybe, I'm not enough clear and concise and I'm sorry about that. I do my best to answer and support all request and ensure a clean git history. 😞 :disappointed:

Currently, my concerns is that the PR should be **ONLY** to add the _generic F401Rx variant_.
I will open a new PR based on your work around the analog pins definition (which still need several tests to ensure not add regression) and then I will be able to merge this PR to add the generic board support.

So give me some time to do the job. 😉

MCUdude:
This PR now only contains the generic variant, and not anything else. Do you want me to open a new PR where changes to pins_arduino.h are done or do you want to do this yourself to get it exactly how you want it to be?

fpistm:
Thanks  @MCUdude 
I will do the PR as I will probably add some more related commits.

MCUdude:
> Thanks @MCUdude
I will do the PR as I will probably add some more related commits.

Great! I'm pretty sure Travis CI fails to compile this PR because it's missing the PR you're about to do that changed the behavior of pins_arduino.h. But apart from this, are you satisfied with the PR now? I assume you won't (and can't) merge this before pins_arduino.h is modified.

fpistm:
At a first look, it seems OK but I will deeply review and test it later using the new PR I will submit.
And yes this will not be merged before the new one.

fpistm:
Required #800

fpistm:
@MCUdude 
unfortunately, as I expected redefining `analogInputToDigitalPin` is not the good way.
In this variant it works but for other this could not because there is no way to have the same value for
AnalogRead(A0) and analogRead(0).

Ex of pin definitions:
```c
#define PC13 0
#define PC14 1
#define PC15 2
#define PF0  3
#define PF1  4
#define PC0  5  // A0
#define PC1  6  // A1
#define PC2  7  // A2
#define PC3  8  // A3
#define PA0  9  // A4/USER_BTN
#define PA1  10 // A5
#define PA2  11 // A6
#define PA3  12 // A7
#define PF4  13
#define PF5  14
#define PA4  15 // A8
#define PA5  16 // A9
#define PA6  17 // A10
#define PA7  18 // A11
#define PC4  19 // A12
#define PC5  20 // A13
#define PB0  21 // A14
#define PB1  22 // A15
```

MCUdude:
Well, the pinout has to be mapped in a specific way to support both analogRead(0) _and_ analogRead(A0). That's just the nature of it. TBH a function like analogReadChannel(0) would be much more suitable, because it may be difficult to combine the Ax constant and the channel number

What pinout is the code posted above from?

I suggest that all analog pins are only available on pinouts that allow channel number and pin number to be combined, like in this pinout

fpistm:
> What pinout is the code posted above from?

Disco_F030R8

> I suggest that all analog pins are only available on pinouts that allow channel number and pin number to be combined, like in this pinout

Unfortunately, this will not be maintainable and hard to review. This have to be as generic as possible.

MCUdude:
It seems like for the Disco F030R8 the NUM_ANALOG_FIRST is defined as 50. This makes for a nice offset to prevent channel numbers and Ax constants to overlap. However, you can't expect analogRead(5) (or analogRead(PC0)) to work, because it will choose analog channel 5, not the channel attached to digital pin 5.

The question really is, How can we make all analog pins available on as many targets as possible. To me, it's not as important if the what number/constant/macro the analogRead function takes, but it has to be consistent. What do you suggest? How can we achieve 16 analog inputs on the generic F401Rx pinout without having to re-map the digital pins to fit/match the analog ones?

fpistm:
In your proposal, `NUM_ANALOG_FIRST` is not used so this could not work.
We agreed this need to be consistent that's why I had split your PR, to have a dedicated one focused on analog pin definition.
I have an idea but I need some time to work on it.

MCUdude:
Looking forward to seeing what you're able to come up with. What's important for me is that we can keep this exact pin mapping and still be able to access all analog pins in some way. This should be a goal for this repository as well IMO 🙂

fpistm:
> Looking forward to seeing what you're able to come up with. What's important for me is that we can keep this exact pin mapping and still be able to access all analog pins in some way. This should be a goal for this repository as well IMO 🙂

Unfortunately, I will not be able to find a way to support this.
This is always the same drawback having:
`analogRead(Ax)` == `analogRead(x)` == `analogRead(PYn)`

If `PYn` is less than `NUM_ANALOG_INPUTS`, there is no way to differentiate if `x` or `PYn`.

So if you want the generic variant for F401Rx, you will have to reorder the pin number.
```c
//                 | ANALOG | USART     | TWI      | SPI                    | SPECIAL   |
//                 |--------|-----------|----------|------------------------|-----------|
#define PA0  21 // | A0     |           |          |                        |           |
#define PA1  22 // | A1     |           |          |                        |           |
#define PA2  23 // | A2     | USART2_TX |          |                        |           |
#define PA3  24 // | A3     | USART2_RX |          |                        |           |
#define PA4  25 // | A4     |           |          | SPI1_SS, (SPI3_SS)     |           |
#define PA5  26 // | A5     |           |          | SPI1_SCK               |           |
#define PA6  27 // | A6     |           |          | SPI1_MISO              |           |
#define PA7  28 // | A7     |           |          | SPI1_MOSI              |           |
#define PA8  0  // |        |           | TWI3_SCL |                        |           |
#define PA9  1  // |        | USART1_TX |          |                        |           |
#define PA10 2  // |        | USART1_RX |          |                        |           |
#define PA11 3  // |        | USART6_TX |          |                        |           |
#define PA12 4  // |        | USART6_RX |          |                        |           |
#define PA13 5  // |        |           |          |                        | SWD_SWDIO |
#define PA14 6  // |        |           |          |                        | SWD_SWCLK |
#define PA15 7  // |        |           |          | SPI1_SS, (SPI3_SS)     |           |
//                 |--------|-----------|----------|------------------------|-----------|
#define PB0  29 // | A8     |           |          |                        |           |
#define PB1  30 // | A9     |           |          |                        |           |
#define PB2  8 // |        |           |          |                        | BOOT1     |
#define PB3  9 // |        |           | TWI2_SDA | SPI1_SCK,  (SPI3_SCK)  |           |
#define PB4  10 // |        |           | TWI3_SDA | SPI1_MISO, (SPI3_MISO) |           |
#define PB5  11 // |        |           |          | SPI1_MOSI, (SPI3_MOSI) |           |
#define PB6  12 // |        | USART1_TX | TWI1_SCL |                        |           |
#define PB7  13 // |        | USART1_RX | TWI1_SDA |                        |           |
#define PB8  14 // |        |           | TWI1_SCL |                        |           |
#define PB9  15 // |        |           | TWI1_SDA | SPI2_SS                |           |
#define PB10 16 // |        |           | TWI2_SCL | SPI2_SCK               |           |
#define PB12 17 // |        |           |          | SPI2_SS                |           |
#define PB13 18 // |        |           |          | SPI2_SCK               |           |
#define PB14 19 // |        |           |          | SPI2_MISO              |           |
#define PB15 20 // |        |           |          | SPI2_MOSI              |           |
//                 |--------|-----------|----------|------------------------|-----------|
#define PC0  31 // | A10    |           |          |                        |           |
#define PC1  32 // | A11    |           |          |                        |           |
#define PC2  33 // | A12    |           |          | SPI2_MISO              |           |
#define PC3  34 // | A13    |           |          | SPI2_MOSI              |           |
#define PC4  35 // | A14    |           |          |                        |           |
#define PC5  36 // | A15    |           |          |                        |           |
#define PC6  37 // |        | USART6_TX |          |                        |           |
#define PC7  38 // |        | USART6_RX |          |                        |           |
#define PC8  39 // |        |           |          |                        |           |
#define PC9  40 // |        |           | TWI3_SDA |                        |           |
#define PC10 41 // |        |           |          | SPI3_SCK               |           |
#define PC11 42 // |        |           |          | SPI3_MISO              |           |
#define PC12 43 // |        |           |          | SPI3_MOSI              |           |
#define PC13 44 // |        |           |          |                        |           |
#define PC14 45 // |        |           |          |                        | OSC32_IN  |
#define PC15 46 // |        |           |          |                        | OSC32_OUT |
//                 |--------|-----------|----------|------------------------|-----------|
#define PD2  47 // |        |           |          |                        |           |
//                 |--------|-----------|----------|------------------------|-----------|
#define PH0  48 // |        |           |          |                        | OSC_IN    |
#define PH1  49 // |        |           |          |                        | OSC_OUT   |
//                 |--------|-----------|----------|------------------------|-----------|
```
As for generic variant, user mainly use the `PYn` notation this will not be an issue.
Of course, the `digitalPin[]` array have to be reordered and define` NUM_ANALOG_FIRST` to 21.

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,170 @@
...
+#define _VARIANT_ARDUINO_STM32_
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+
+//                 | ANALOG | USART     | TWI      | SPI                    | SPECIAL   |
+//                 |--------|-----------|----------|------------------------|-----------|
+#define PA0  0  // | A0     |           |          |                        |           |
```
fpistm: Please review pin numbering as mentioned here:
https://github.com/stm32duino/Arduino_Core_STM32/pull/787#issuecomment-559700664

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,170 @@
...
+#define PC15 46 // |        |           |          |                        | OSC32_OUT |
+//                 |--------|-----------|----------|------------------------|-----------|
+#define PD2  47 // |        |           |          |                        |           |
+//                 |--------|-----------|----------|------------------------|-----------|
+#define PH0  48 // |        |           |          |                        | OSC_IN    |
+#define PH1  49 // |        |           |          |                        | OSC_OUT   |
+//                 |--------|-----------|----------|------------------------|-----------|
+
+// This must be a literal
+#define NUM_DIGITAL_PINS        50
```
fpistm: To be updated according to number of PinName in digitalPin[] array.

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,170 @@
...
+#define NUM_ANALOG_INPUTS       16
+#define NUM_ANALOG_FIRST        0
```
fpistm: To be updated after Analog pin number update.

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,170 @@
...
+
+// On-board LED pin number
+#define LED_BUILTIN             PA5
+#define LED_GREEN               LED_BUILTIN
+
+// On-board user button
+#define USER_BTN                PC13
+
+// Analog pin definitions
```
fpistm: all `PIN_Ax` have to be removed.

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,170 @@
...
+#define PIN_A9       PB1
+#define PIN_A10      PC0
+#define PIN_A11      PC1
+#define PIN_A12      PC2
+#define PIN_A13      PC3
+#define PIN_A14      PC4
+#define PIN_A15      PC5
+
+// Convert analog channel, analog pin or digital pin to digital pin
+#define analogInputToDigitalPin(p) ((p <= 7) ? (p) : (p <= 9) ? (p + 8) : (p <= 15) ? (p + 21) : (p))
```
fpistm: `analogInputToDigitalPin` have to be removed

MCUdude:
> Unfortunately, I will not be able to find a way to support this.
This is always the same drawback having:
analogRead(Ax) == analogRead(x) == analogRead(PYn)
If PYn is less than NUM_ANALOG_INPUTS, there is no way to differentiate if x or PYn.

How is this solved on the Disco_F030R8 today? I bet analogRead(Ax) == analogRead(x) == analogRead(PYn) wont work. Any my pinout has nothing to do with this.

Why would it be a problem to modify the `analogInputToDigitalPin ` just for this pinout? It allows for flexible analogRead, and it does not affect any other pinouts. All existing ones can stay the same.

fpistm:
> How is this solved on the Disco_F030R8 today? I bet analogRead(Ax) == analogRead(x) == analogRead(PYn) wont work. Any my pinout has nothing to do with this.

Currently, this works. 

> Why would it be a problem to modify the `analogInputToDigitalPin ` just for this pinout? It allows for flexible analogRead, and it does not affect any other pinouts. All existing ones can stay the same.

Because, it could not be applied to all possible pins mapping.

Anyway, I've just found a way to have a new generic way to define analog pins and seems removed all drawback.

MCUdude:
>> Why would it be a problem to modify the analogInputToDigitalPin just for this pinout? It allows for flexible analogRead, and it does not affect any other pinouts. All existing ones can stay the same.

> Because, it could not be applied to all possible pins mapping.

I don't really see why `analogInputToDigitalPin ` has to be the same on all pinouts. After all, this is a helper macro that is used to get the "Arduino things" right.

> Anyway, I've just found a way to have a new generic way to define analog pins and seems removed all drawback.

Tell me about it!

Bottom line: what I eventually want to achieve with this pinout (and maybe more in the future) is that a user can Start Arduino IDE or VSCode/PlatformIO and just pick the STM32 that's needed for the job. No board, NO *only the top of the line* chip available, but _all_ memory options. A little bit like what I've done with my various AVR cores. This is also super useful if you find a commercial product with an STM32 on that you's like to write your own FW to.

fpistm:
> Bottom line: what I eventually want to achieve with this pinout (and maybe more in the future) is that a user can Start Arduino IDE or VSCode/PlatformIO and just pick the STM32 that's needed for the job. No board, NO _only the top of the line_ chip available, but _all_ memory options. A little bit like what I've done with my various AVR cores. This is also super useful if you find a commercial product with an STM32 on that you's like to write your own FW to.

I think you don't understand you can already do that... As I said for generic board user will use the PYn format and are not aware of which value it is in the background. So, I really do not understand your point about that as the pin mapping could easily be done like I do in a previous comment. While doing an analogread(PYn) works...

fpistm:
@MCUdude 

> Tell me about it!

this is now available in #800

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,169 @@
...
+
+// On-board LED pin number
+#define LED_BUILTIN             PA5
+#define LED_GREEN               LED_BUILTIN
+
+// On-board user button
+#define USER_BTN                PC13
+
+// Analog pin definitions
+#define PIN_A0       PA0
```
fpistm: Remove all `#define PY_n` this is done in `pin_arduino.h`

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,169 @@
...
+#define PIN_A9       PB1
+#define PIN_A10      PC0
+#define PIN_A11      PC1
+#define PIN_A12      PC2
+#define PIN_A13      PC3
+#define PIN_A14      PC4
+#define PIN_A15      PC5
+
+// Convert analog channel, analog pin or digital pin to digital pin
+#define analogInputToDigitalPin(p) ((p <= 7) ? (p) : (p <= 9) ? (p + 8) : (p <= 15) ? (p + 21) : (p))
```
fpistm: `analogInputToDigitalPin` is generic so no need to redefine it.

variants/Generic_F401Rx/variant.h
```diff
@@ -0,0 +1,148 @@
...
+#define PC15 46  // | 46      |        |           |          |                        | OSC32_OUT |
+//                  |---------|--------|-----------|----------|------------------------|-----------|
+#define PD2  47  // | 47      |        |           |          |                        |           |
+//                  |---------|--------|-----------|----------|------------------------|-----------|
+#define PH0  48  // | 48      |        |           |          |                        | OSC_IN    |
+#define PH1  49  // | 49      |        |           |          |                        | OSC_OUT   |
+//                  |---------|--------|-----------|----------|------------------------|-----------|
+
+// This must be a literal
+#define NUM_DIGITAL_PINS        34
```
fpistm:
```suggestion
#define NUM_DIGITAL_PINS        50
```

MCUdude: But you did this with the other pinouts?

fpistm: The other I modified, I removed the duplicated pins that's all. All pins are digital so in your variants it is 50.

MCUdude:
OK. Fixed now

variants/Generic_F401Rx/variant.cpp
```diff
@@ -22,16 +22,9 @@
...
 const PinName digitalPin[] = {
-  PA_0,  // Digital pin 0
-  PA_1,  // Digital pin 1
-  PA_2,  // Digital pin 2
-  PA_3,  // Digital pin 3
-  PA_4,  // Digital pin 4
-  PA_5,  // Digital pin 5
-  PA_6,  // Digital pin 6
-  PA_7,  // Digital pin 7
+  // PA_0..PA_7 are analog pins
```
fpistm: I didn't saw this before but you have to define those pins here else nothing will works.
This array references all digital pins.

variants/Generic_F401Rx/variant.cpp
```diff
@@ -41,8 +34,7 @@ const PinName digitalPin[] = {
   PA_14, // Digital pin 14
   PA_15, // Digital pin 15
 
-  PB_0,  // Digital pin 16
-  PB_1,  // Digital pin 17
+  // PB_0..PB_1 are analog pins
```
fpistm: dito

variants/Generic_F401Rx/variant.cpp
```diff
@@ -57,12 +49,7 @@ const PinName digitalPin[] = {
   PB_14, // Digital pin 29
   PB_15, // Digital pin 30
 
-  PC_0,  // Digital pin 31
-  PC_1,  // Digital pin 32
-  PC_2,  // Digital pin 33
-  PC_3,  // Digital pin 34
-  PC_4,  // Digital pin 35
-  PC_5,  // Digital pin 36
+  // PC_0..PC_5 are analog pins
```
fpistm: dito

MCUdude:
How about now?

uzi18:
@MCUdude is it possible for you to create also generic variant for F407VET/VGT?

MCUdude:
> @MCUdude is it possible for you to create also generic variant for F407VET/VGT?

It's certainly is possible, but I'm not planning to use an F407 in a design at the moment. I also don't have any F407 hardware to test with. However, I _do_ plan to add F401Cx, F411Rx, and F411Cx.

PR #793, opened 2019-11-26
Correct H743ZI2 ADC pin inputs

jimbo00000:
## Fix analog pin definitions for NUCLEO H743ZI2

**Summary**
Some analog pins on the H743ZI2 were returning duplicate values. This change updates the entries for Arduino pins A4, A6 and A8 to the correct inputs so the below sketch gives correct results.
Thank you to DIY DSP for retrieving the relevant info from the datasheet.

**Validation**
```
const int8_t analogPins[] = {
  // A6=PF_4 A7=PF_5 A8=PF_6(CN10)  A9=PF11(CN12)
  A0, A1, A2, A3, A4, A5, A6, A7, A8, A9,
};
const uint8_t numPins = sizeof(analogPins) / sizeof(analogPins[0]);

void setup() {
  Serial.begin(9600);
}

void loop() {
  for (uint8_t i = 0; i < numPins; ++i) {
    int pin = analogPins[i];
    int val = analogRead(pin);
    Serial.print("A"); Serial.print(pin-A0); Serial.print("=");
    Serial.print(val); Serial.print("\t");
  }
  Serial.println();
  delay(10); // delay in between reads for stability
}
```


fpistm:
Hi @jimbo00000 
You're right there is an issue on those pins anyway this is not the correct fix.
The channel is OK according to the datasheet. The channel you set are for another packages.
I've do the fix in https://github.com/stm32duino/Arduino_Core_STM32/commit/4be877dde0f95b1f9d4e7928da616f3befe29bee

PR #794, opened 2019-11-26
SPI: Fix prescaler at low baudrates

LinoBarreca:
**Summary**
The current prescaler, if the user requests a slow clock, is set to a value which makes no sense.
We should give a frequency closer to what the user asks.

* [x] BUG: Clock is too high when the user requests a slow clock.

**motivation** for making this change:
If a SPI device can't handle high speed clock and the user requests a slow one we should give him the slowest clock we can. The code, before the PR, had unexpected behavior which makes difficult to track down where the problem is.

**Validation**
Should work.

### DIFF
--- a/cores/arduino/stm32/spi_com.c
+++ b/cores/arduino/stm32/spi_com.c
@@ -215,8 +215,8 @@ void spi_init(spi_t *obj, uint32_t speed, spi_mode_e mode, uint8_t msb)
     handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_128;
   } else if (speed >= (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ)) {
     handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256;
-  } else {
-    handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_16;
+  } else { //speed is lower than (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ). we can't go below this.
+    handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256; // Set prescaler at max value so we have the lowest frequency possible
   }
 
   handle->Init.Direction         = SPI_DIRECTION_2LINES;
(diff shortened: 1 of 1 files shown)


cores/arduino/stm32/spi_com.c
```diff
@@ -215,8 +215,8 @@ void spi_init(spi_t *obj, uint32_t speed, spi_mode_e mode, uint8_t msb)
     handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_128;
   } else if (speed >= (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ)) {
     handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256;
-  } else {
-    handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_16;
+  } else { //speed is lower than (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ). we can't go below this.
```
fpistm: This make sense. Nice catch.
We could  probably remove the test:
`(speed >= (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ))`
as it is the default value and adding a comment that in any case this would be the best choice.
Something like that.

LinoBarreca: Yes, we could. I just left it for clarity.
Do you want me to remove it?
(you can do it yourself anyway, since I left the "Allow edits from maintainers" option checked)

..and thanks for the "nice catch" :smiley:
I found it while trying to figure out why all the users of the SKR-Pro can't access a TransFlash connected to SPI1 but they can while it's connected to SPI2.
Since I don't have the board (it's still in transit) I had to analyze the code "on the paper" without debugging it....and that could be a reason: Since SPI1 and SPI2 use different clocks, in one case (SPI2) we might fall under the "(speed >= (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ))" if, while in the other (SPI1) we just try to access it with a clock definitely not compatible with TF cards range (which, according to [here](http://elm-chan.org/docs/mmc/mmc_e.html) should be 100-400kHz)

I don't know the actual clocks so I might be wrong but this should have been fixed anyway.

fpistm: Well if you can do it this will be much easier.
For clarity, as I said, simply add a comment like that:
```c
  } else {
    /*
     * As it is not possible to go below (spi_freq / SPI_SPEED_CLOCK_DIV256_MHZ).
     * Set prescaler at max value so get the lowest frequency possible.
     */
    handle->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256;
  }
```

LinoBarreca: I thought you had this menu too:
immagine
That's why I said you could do that easily :)
Anyway.... Done

LinoBarreca:
> Thanks @LinoBarreca

no, thank you :)

cores/arduino/stm32/spi_com.c
fpistm: Ah Right.
I thought I need to fetch your fork to do this.
We always learn ;) Thanks

PR #795, opened 2019-11-26
Harden Print::printf() method

fpistm:
Based on @PaulStoffregen works for Teensyduino:
https://github.com/PaulStoffregen/cores

* Remove `PRINTF_BUFFER` usage.
* Aligned declaration with `printf()` one by returning the number of characters printed (excluding the null byte used to end output to strings).

### Tested:
 * core debug
 * standard `printf` usage on the `DEBUG_UART`
 * Serial over USB

Harden #780 from @MCUdude.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -21,8 +21,8 @@
  */
 
 #include <stdlib.h>
-#include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 #include <math.h>
 #include "Arduino.h"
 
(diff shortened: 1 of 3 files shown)


MCUdude:
Can you verify that this can be used on two or more serial ports at the same time? I tested out the version from the Teensy core, but `Serial.printf()` and `Serial1.printf()` all printed to the same serial port. Adafruits implementation (found in #780) worked fine.

fpistm:
Serial is just an aliases to a SerialX instance. That why it print on the same serial port.

MCUdude:
It's still important that we can use SerialX.printf on more than one serial port at the time. Will this be possible?

fpistm:
This is possible and tested.

fpistm:
Example for Nucleo-F411
`Serial1` uses PA10 and PA9
`Serial` is mapped on `Serial2` connected to STLink
```C++
HardwareSerial Serial1(USART1);

const char * lorem = "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.";

void setup() {
  Serial.begin(9600);
  Serial1.begin(9600);
}

void loop() {
  static uint32_t i = 0;
  Serial.printf("Iter %i --> %s\n", i++, lorem);
  Serial1.printf("Iter %i --> %s\n", i++, lorem);
}
```

MCUdude:
I did test the Teensy implementation on a F401Rx board. I manually defined Serial1 at the beginning at the sketch and looked at the output in the serial monitor (Serial) and with a logic analyzer (Serial1). Even though it shouldn't, I got the same output in the serial monitor and in the logic analyzer capture window. However, I can test this particular implementation on by Nucle F401RE board tomorrow.

MCUdude:
I just tested it on my Nucleo F401RE. Works perfectly! I do agree that this implementation is better than #780. And getting rid of the buffer is definitely a win!

PR #796, opened 2019-11-27
Swap digitalPin names of {PD_10, PG_4, PG_5} with {PF_12, PF_13, PF_14}

jimbo00000:
Moving the positions of these pins allows for 3 more analog inputs. Tested with the following sketch:

```
const int8_t analogPins[] = {
  A0, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12,
};
const uint8_t numPins = sizeof(analogPins) / sizeof(analogPins[0]);

void setup() {
  Serial.begin(9600);
}

void loop() {
  for (uint8_t i = 0; i < numPins; ++i) {
    int pin = analogPins[i];
    int val = analogRead(pin);
    Serial.print("A"); Serial.print(pin-A0); Serial.print("=");
    Serial.print(val); Serial.print("\t");
  }
  Serial.println();
  delay(10); // delay in between reads for stability
}
```


variants/NUCLEO_H743ZI/PeripheralPins.c
```diff
@@ -106,6 +106,9 @@ WEAK const PinMap PinMap_ADC[] = {
   //  {PF_13, ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // ADC2_INP2
   //  {PF_14, ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // ADC2_INN2
   //  {PF_14, ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC2_INP6
+  {PF_12, ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 6, 0)}, // ADC1_INP6 - A10
```
fpistm: Those lines are already present.
It only requires to uncomment them.

fpistm:
@jimbo00000 
I will not merge this.
We currently working to remove analog pins definition and avoid them to be consecutive.
All variants will be updated to add more ADC.

PR #806, opened 2019-11-29
HardwareTimer: rework internal mapping of arduino API to HAL/LL API

### LINKED ISSUE #763: HardwareTimer incomplete?

ABOSTM:
**Summary**

Rework HardwareTimer internal working. And add new API to pause/resume individual channel.
Main rework:
* `HAL_TIM_Base_Init()` is now called only once at object creation
* `HAL_TIM_xxx_ConfigChannel` is now done in `setMode()`
* `HAL_TIM_xxx_Start` is done in `resumeChannel()`
* use LL when possible
* Configuration are directly made through hardware register access (when possible),
  then remove useless attribute
* Add new API to pause only one channel (fixes #763):
```c
     pauseChannel(uint32_t channel);
     resumeChannel(uint32_t channel);
```
* integration of PR #767 Flexible interrupts handling

- [x] Update wiki when merged

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -64,38 +64,37 @@ HardwareTimer::HardwareTimer(TIM_TypeDef *instance)
   _timerObj.handle.Lock = HAL_UNLOCKED;
   _timerObj.handle.State = HAL_TIM_STATE_RESET;
 
-  _timerObj.handle.Instance = instance;
   _timerObj.__this = (void *)this;
   _timerObj.preemptPriority = TIM_IRQ_PRIO;
   _timerObj.subPriority = TIM_IRQ_SUBPRIO;
 
-  // Enable Timer clock
+  /* Enable timer clock. Even if it is also done in HAL_TIM_Base_MspInit(),
+     it is done there so that it is possible to write registers right now */
   enableTimerClock(&(_timerObj.handle));
 
-  // Configure HAL structure for all channels
-  for (int i = 0; i < TIMER_CHANNELS; i++) {
-    _channelOC[i].OCMode = TIMER_NOT_USED;
-    _channelOC[i].OCPolarity = TIM_OCPOLARITY_HIGH;
-    _channelOC[i].OCFastMode = TIM_OCFAST_DISABLE;
-
-#if defined(TIM_CR2_OIS1)
-    _channelOC[i].OCIdleState = TIM_OCIDLESTATE_RESET;
-#endif
-#if defined(TIM_CCER_CC1NE)
-    _channelOC[i].OCNPolarity = TIM_OCNPOLARITY_HIGH;
-#if defined(TIM_CR2_OIS1)
-    _channelOC[i].OCNIdleState = TIM_OCNIDLESTATE_RESET;
-#endif
-#endif
-    _channelIC[i].ICPolarity = TIMER_NOT_USED;
-    _channelIC[i].ICSelection = TIM_ICSELECTION_DIRECTTI;
-    _channelIC[i].ICPrescaler = TIM_ICPSC_DIV1;
-    _channelIC[i].ICFilter = 0;
-
-    for (int i = 0; i < TIMER_CHANNELS + 1 ; i++) {
-      callbacks[i] = NULL;
-    }
+  // Initialize NULL callbacks
+  for (int i = 0; i < TIMER_CHANNELS + 1 ; i++) {
+    callbacks[i] = NULL;
   }
+
+  // Initialize channel mode and complementary
+  for (int i = 0; i < TIMER_CHANNELS; i++) {
+#if defined(TIM_CCER_CC1NE)
+    isComplementaryChannel[i] = false;
+#endif
+    _ChannelMode[i] = TIMER_DISABLED;
+  }
+
+  /* Configure timer with some default values */
+  _timerObj.handle.Init.Prescaler = 0;
+  _timerObj.handle.Init.Period = 0xFFFF; // 16bit max value
+  _timerObj.handle.Init.CounterMode = TIM_COUNTERMODE_UP;
+  _timerObj.handle.Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
+#if defined(TIM_RCR_REP)
+  _timerObj.handle.Init.RepetitionCounter = 0;
+#endif
+  _timerObj.handle.Init.AutoReloadPreload = TIM_AUTORELOAD_PRELOAD_DISABLE;
+  HAL_TIM_Base_Init(&(_timerObj.handle));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


cores/arduino/HardwareTimer.cpp
```diff
@@ -660,6 +810,12 @@ void HardwareTimer::setInterruptPriority(uint32_t preemptPriority, uint32_t subP
   */
 void HardwareTimer::attachInterrupt(void (*callback)(HardwareTimer *))
 {
+  if (callback != NULL) {
+    // Clear flag before enabling IT
+    __HAL_TIM_CLEAR_FLAG(&(_timerObj.handle), TIM_FLAG_UPDATE);
+    // Enable update interrupt only is callback is valid
```
fpistm:
```suggestion
    // Enable update interrupt only if callback is valid
```

cores/arduino/HardwareTimer.h
```diff
@@ -131,13 +134,13 @@ class HardwareTimer {
     // The following function(s) are available for more advanced timer options
     TIM_HandleTypeDef *getHandle();  // return the handle address for HAL related configuration
   private:
-    TIM_OC_InitTypeDef _channelOC[TIMER_CHANNELS];
-    TIM_IC_InitTypeDef _channelIC[TIMER_CHANNELS];
+    uint32_t _ChannelMode[TIMER_CHANNELS];
```
fpistm: Should be `TimerModes_t`?
```suggestion
    TimerModes_t _ChannelMode[TIMER_CHANNELS];
```

cores/arduino/HardwareTimer.cpp
```diff
@@ -64,38 +64,35 @@ HardwareTimer::HardwareTimer(TIM_TypeDef *instance)
...
+  for (int i = 0; i < TIMER_CHANNELS + 1 ; i++) {
+    callbacks[i] = NULL;
+  }
+
+  // Initialize channel mode and complementary
   for (int i = 0; i < TIMER_CHANNELS; i++) {
-    _channelOC[i].OCMode = TIMER_NOT_USED;
-    _channelOC[i].OCPolarity = TIM_OCPOLARITY_HIGH;
-    _channelOC[i].OCFastMode = TIM_OCFAST_DISABLE;
+    isComplementaryChannel[i] = false;
```
fpistm:
```suggestion
#if defined(TIM_CCER_CC1NE)
    isComplementaryChannel[i] = false;
#endif
```

fpistm:
```suggestion
#if defined(TIM_CCER_CC1NE)	
    isComplementaryChannel[i] = false;
 #endif
```

cores/arduino/HardwareTimer.cpp
```diff
@@ -164,6 +192,112 @@ int HardwareTimer::getChannel(uint32_t channel)
...
 
+/**
+  * @brief  Convert arduino channel into LL channel
+  * @param  Arduino channel [1..4]
+  * @retval LL channel. return -1 if arduino channel is invalid
+  */
+int HardwareTimer::getLLChannel(uint32_t channel)
+{
+  uint32_t return_value;
+  if (isComplementaryChannel[channel - 1]) {
```
fpistm:
```suggestion
#if defined(TIM_CCER_CC1NE)
  if (isComplementaryChannel[channel - 1]) {
```

cores/arduino/HardwareTimer.cpp
```diff
@@ -164,6 +192,112 @@ int HardwareTimer::getChannel(uint32_t channel)
...
+        break;
+#if defined(LL_TIM_CHANNEL_CH4N)
+      case 4:
+        return_value = LL_TIM_CHANNEL_CH4N;
+        break;
+#endif
+      default:
+        return_value = -1;
+    }
+  } else {
```
fpistm:
```suggestion
  } else 
  #endif
  {
```

cores/arduino/HardwareTimer.cpp
ABOSTM: Done

cores/arduino/HardwareTimer.h
ABOSTM: Done

cores/arduino/HardwareTimer.cpp
ABOSTM: Done

ABOSTM: Done

cores/arduino/HardwareTimer.cpp
ABOSTM: Done

cores/arduino/HardwareTimer.cpp
ABOSTM: Done

ABOSTM:
Hi @matthijskooijman,
Thank for your comments.
I disagree with your proposal to enable interrupt only in resume()/resumeChannel() and only when timer or channel is running:
the following basic sketch would not work anymore, interrupts will never be enabled:
```
  MyTim->setMode(channel, TIMER_OUTPUT_COMPARE_PWM1, pin);
  // MyTim->setPrescaleFactor(8); 
  MyTim->setOverflow(100000, MICROSEC_FORMAT); // 100000 microseconds = 100 milliseconds
  MyTim->setCaptureCompare(channel, 50, PERCENT_COMPARE_FORMAT); // 50%
  MyTim->attachInterrupt(Update_IT_callback);
  MyTim->attachInterrupt(channel, Compare_IT_callback);
  MyTim->resume();
```

Also, I think about taking only part of your proposal: in attachInterrupt(), enable IT only if channel is running. 
But there is currently no mean to know a channel is enabled (mode is not enough as 0 is a valid mode) and we can have tha case where we attach a callback to compare event without output on IO.
I think globally it is easier to understand:
* attachInterrupt() will attach the call back and enable IT. 
* resume()/resumeChannel will resume/enable IT 
* dettach()/pause()/pauseChannel() will do the opposite.
And thus no need to have complex crosscase documentation like "what happends if ..."

That is right that if you call attachInterrupt() while channel is no running, it will enable the IT, and it can fire the callback. Up to application to configure channel 1rst.

cores/arduino/HardwareTimer.cpp
```diff
@@ -680,10 +843,21 @@ void HardwareTimer::detachInterrupt()
...
+    Error_Handler();
+  }
+
   if ((channel == 0) || (channel > (TIMER_CHANNELS + 1))) {
     Error_Handler();  // only channel 1..4 have an interrupt
   }
 
+  if (callback != NULL) {
+    // Clear flag before enabling IT
+    __HAL_TIM_CLEAR_FLAG(&(_timerObj.handle), interrupt);
```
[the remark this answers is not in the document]
matthijskooijman: I am not entirely sure about the timing details on STM32, but I can imagine that the interrupt arrives at the same time as the clear, and thus is cleared before it can run.

Another case is when code is called when interrupts are temporarily disabled (e.g. using the global interrupt mask, or when running from an ISR with higher priority). In that case, the flag might be set before the clear, but the interrupt cannot run yet, so the clear loses one interrupt.

> And Clear is necessary for example in the following use case:
The timer is already running with interrupts disabled. After a while, Flags Update or Compare will be set. Then if we enable interrupt we must clear the flag to discard past interrupts and take into account only new ones.

Yeah, I agree on that. Hence my suggestion to only clear if the interrupt is not enabled yet, which I think should cover the case you describe? If the interrupt is already enabled, then the clear is not needed, assuming that any interrupts that are still pending should be handled by the new callback rather than be lost.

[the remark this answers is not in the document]
ABOSTM: Right, I agree with your proposal. I implement it in https://github.com/stm32duino/Arduino_Core_STM32/pull/852

PR #818, opened 2019-12-04
Added experimental support for BlackPill F411CE

Jozo132:
Added a new board to dropdown menu based on existing variant.
This is no breaking change.
Bugs are probably hiding somewhere because we are referencing a slightly different variant with the same layout to support this board.
Tests need to be done to see if everything is compatible.

I own two BlackPills F401CC and F411CE, both on the same PCB.
AliExpress link: https://www.aliexpress.com/item/4000069263843.html
The only difference seems to be 256K Flash + 64K RAM versus 512K Flash + 128K RAM.
Compiles, uploads and works for me over DFU with serial out.


fpistm:
Hi @Jozo132 
Thanks for this PR

> Tests need to be done to see if everything is compatible.

Let me know if all is OK.

fpistm:
Hi @Jozo132 
I will not merge this PR.
The variant for this board will use the Generic variant definition.

PR #822, opened 2019-12-08
bugfix for marlin

MangaValk:
**Summary**

This PR fixes an issue i had on Marlin when using multiple servos.
Issue: https://github.com/MarlinFirmware/Marlin/issues/16138

### DIFF
--- a/libraries/Servo/src/Servo.h
+++ b/libraries/Servo/src/Servo.h
@@ -110,8 +110,9 @@ class Servo {
     int read();                        // returns current pulse width as an angle between 0 and 180 degrees
     int readMicroseconds();            // returns current pulse width in microseconds for this servo (was read_us() in first release)
     bool attached();                   // return true if this servo is attached, otherwise false
-  private:
+  protected:
     uint8_t servoIndex;               // index into the channel data for this servo
+  private:
     int8_t min;                       // minimum is this value times 4 added to MIN_PULSE_WIDTH
     int8_t max;                       // maximum is this value times 4 added to MAX_PULSE_WIDTH
 };
--- a/libraries/Servo/src/stm32/Servo.cpp
+++ b/libraries/Servo/src/stm32/Servo.cpp
@@ -29,7 +29,7 @@ static volatile int8_t timerChannel[_Nbr_16timers] = {-1}; // counter for the se
 
 static HardwareTimer TimerServo(TIMER_SERVO);
 
-uint8_t ServoCount = 0;                                    // the total number of attached servos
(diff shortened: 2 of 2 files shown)


fpistm:
Hi @MangaValk 
could you be more precise about why you need to do those updates?
And add this explanation in the commit message.
Thanks

libraries/Servo/src/Servo.h
```diff
@@ -110,8 +110,9 @@ class Servo {
     int read();                        // returns current pulse width as an angle between 0 and 180 degrees
     int readMicroseconds();            // returns current pulse width in microseconds for this servo (was read_us() in first release)
     bool attached();                   // return true if this servo is attached, otherwise false
-  private:
+  protected:
```
ABOSTM: I don't see a reason to change this attribute to protected instead of private.
This variable should not be modified outside of the class.
I can understand that you need to read this variable, but in this case we can rather implement an accessor.

libraries/Servo/src/stm32/Servo.cpp
```diff
@@ -29,7 +29,7 @@ static volatile int8_t timerChannel[_Nbr_16timers] = {-1}; // counter for the se
 
 static HardwareTimer TimerServo(TIMER_SERVO);
 
-uint8_t ServoCount = 0;                                    // the total number of attached servos
+static uint8_t ServoCount = 0;                                    // the total number of attached servos
```
ABOSTM: From my point of view there is no restriction to define this variable static.
Nevertheless, if this cause you issue in Marlin, this probably means you have a variable with the same name ? Can you confirm ? maybe you also need to take some action on marlin to avoid conflict too?

libraries/Servo/src/Servo.h
MangaValk: A getter function would be fine too. I'm not a professional c++ programmer, so I thought a protected var would be fine.

libraries/Servo/src/stm32/Servo.cpp
MangaValk: Here is link to the marlin change I made https://github.com/MarlinFirmware/Marlin/pull/16151#pullrequestreview-328631174

Somehow this made it work, if I turn it back to normal it stops working.

fpistm:
@MangaValk 
what is strange is for other arch supported by Marlin and using the same Servo Class definition (Official Arduino), this is not requested.
My guess is there is an issue how it is implemented for the HAL_STM32 in Marlin.
@ABOSTM already tested the library with multiple Servo without any issue.

MangaValk:
I'm closing this PR since someone found a fix that works without these changes.

PR #848, opened 2019-12-30
Add New Variant VCCGND_F103ZET6

dds90:
Schematics of the board:
[https://ydsd1.oss-cn-shenzhen.aliyuncs.com/STM32F1/103ZEmini.rar](https://ydsd1.oss-cn-shenzhen.aliyuncs.com/STM32F1/103ZEmini.rar)

Problems:
- Seems to work only one Serial (needed 3)
- SD Example of SD32SD library no compile


fpistm:
Hi @dds90 
HAL SD has been modified and I will have to update the STM32SD library.

About Serial, it seems normal like you define 3 times the same define and the last is the 3.
Anyway, I have made a PR to support all the STM32F103Zx in a generic way #851 and add as a derivative the 2 VCC GND board (mini and normal).
So I close this one.

PR #852, opened 2020-01-03
HardwareTimer: Clear IT flag only if callback not previously configured

ABOSTM:
**Summary**

HardwareTimer: Clear IT flag only if callback not previously configured
rework #806 after discussion with @matthijskooijman


matthijskooijman:
Looks good to me, thanks!

PR #863, opened 2020-01-08
Add STM32G0316-DISCO

### LINKED ISSUE #858: [New Variant] STM32G0316-DISCO
I'm interested in using STM32duino for the STM32G0316-DISCO which is based on a STM32G0 in SO8 package.

**Board description:**
 - Name: STM32G0316-DISCO
 - Link to all relevant information:
   * [Documentations](https://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-mpu-eval-tools/stm32-mcu-mpu-eval-tools/stm32-discovery-kits/stm32g0316-disco.html)
   * [Schematics](https://www.st.com/resource/en/schematic_pack/mb1454-g031j6-b02_schematic.pdf)
   * [STM32G031J6 datasheet](https://www.st.com/resource/en/datasheet/stm32g031j6.pdf)

Whilst I think I could add the variant myself according to [the documentation](https://github.com/stm32duino/wiki/wiki/Add-a-new-variant-(board)), the STM32G031J6 has different pads connected to each pin:
> A multi-bonding approach is used on the small packages in order to offer a maximum of alternate functions and analog inputs. This approach results in multiple die pads connected internally to a single package pin.

I am not sure if specific care should be taken or if it's better to let users use pins as if they were on different pins.

Also, how is pin remapping handled by STM32duino? Maybe @fpistm knows about STM32G0 specifics.

DurandA:
This PR adds the STM32G0316-DISCO variant which is based on a STM32G0 in SO8 package.

Fix #858. This is a draft PR as I want to gather some feedback from the maintainers first.


DurandA:
Hi @fpistm, I have a few questions before the PR is ready (and I need to test it properly):
- How is is used `.node` from _boards.txt_? What value should I use for `Disco.menu.pnum.DISCO_G0316.node`?
- As discussed in #858, there is no serial line that goes to the ST-Link in this board. It requires to connect an external serial adapter in order to use _USART1_ (which is the only UART interface in this MCU except _LPUART1_ but no RX pin is available on the STM32G031J6 for _LPUART1_). Is it preferred to completely remove Serial definitions in _variant.h_?

DurandA:
Hi @fpistm, thank you for your patch.

There were many fixes! One of the issues I had is that I misunderstood the comment _All pins are digital, analog inputs are a subset of digital pins._ from _variant.h_ of the _board_template_. I interpreted it as `analogInputPin[]` is a subset of `digitalPin[]` so I thought that `analogInputPin[]` contained indices from `digitalPin[]`, or the starting index of `digitalPin[]` if `NUM_ANALOG_FIRST` was used.

I will do some testing in the next few days so you can merge this. If you already tested it and you think it is ready to merge, I marked the PR as ready for review.

Many thanks!

fpistm:
Hi  @DurandA 
I close this one as too much conflict.
I've made a new one with pin remap handling: #913

PR #866, opened 2020-01-13
Add generic F411Cx pinout

MCUdude:
This pinout is formatted the same way as the Generic F401Rx and F411Rx, and is used with the STM32F411RC and STM32F411RE chips.


MCUdude:
Does it look alright? Anything I can du to make your review easier/quicker?

fpistm:
> 
> 
> Does it look alright? Anything I can du to make your review easier/quicker?

Time is always the missing ingredients.

I will review it when got a slot.

PR #867, opened 2020-01-13
Print::printf() generic

### LINKED ISSUE #862: Print::printf on SerialUSB
Preconditions:
OS: Win10
Arduino 1.8.10
STM32 1.8.0
BluePill F103
HID Bootloader
USB CDC Serial

USB Serial output is not working when setting
HAL_UART_MODULE_ONLY
HAL_UART_MODULE_ENABLED
in build_opt.h

Target: LIN Serial implementeation on UART1 and USB Serial for debugging

Root cause:
https://github.com/stm32duino/Arduino_Core_STM32/blob/ea8664573c69acf7673103b6ef198ef4b22d00b9/cores/arduino/Print.cpp#L199-L220

Function write in Print.cpp gets disabled when setting UART into HAL mode.
But only uart_debug should be disabled. 

Working bugfix:

```C
extern "C"
{
  __attribute__((weak)) int _write(int file, char *ptr, int len)
  {
#if defined(HAL_UART_MODULE_ENABLED) && !defined(HAL_UART_MODULE_ONLY)
    switch (file)
    {
    case STDOUT_FILENO:
    case STDERR_FILENO:
      uart_debug_write((uint8_t *)ptr, (uint32_t)len);
      break;
    case STDIN_FILENO:
      break;
    default:
      ((class Print *)file)->write((uint8_t *)ptr, len);
      break;
    }
#else
    if (file != STDERR_FILENO)
      ((class Print *)file)->write((uint8_t *)ptr, len);
#endif
    return len;
  }
}
```

fpistm:
Before it was only linked to U(S)ART while `Print` is a Base class.

Fixes #862

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -200,22 +200,19 @@ extern "C" {
   __attribute__((weak))
   int _write(int file, char *ptr, int len)
   {
-#if defined(HAL_UART_MODULE_ENABLED) && !defined(HAL_UART_MODULE_ONLY)
     switch (file) {
       case STDOUT_FILENO:
       case STDERR_FILENO:
+#if defined(HAL_UART_MODULE_ENABLED) && !defined(HAL_UART_MODULE_ONLY)
+        /* Used for core_debug() */
         uart_debug_write((uint8_t *)ptr, (uint32_t)len);
-        break;
+#endif
       case STDIN_FILENO:
         break;
       default:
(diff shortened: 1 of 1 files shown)


jandrassy:
why the [SerialUSB] 'tag'?

fpistm:
Because Serial (UART) was already functional.
This PR enable the Print::printf for the Serial USB.
OK as it relies to Print class, all other class depending of Print would benefits from this but in this case this is mainly for the Serial USB. :wink:

PR #851, opened 2020-01-03
F103Zx support

fpistm:
Add Generic F103Zx support and VCCGND_F103ZET6 (Mini)

I have not this boards so if someone could do some tests that's would be fine.
Thanks.

PR #874, opened 2020-01-17
Add generic F401Cx pinout

MCUdude:
This pinout is formatted the same way as the Generic F411Cx pinout and is used with the STM32F401CB/CC/CD/CE chips.


variants/Generic_F401Cx/PeripheralPins.c
```diff
@@ -0,0 +1,228 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *******************************************************************************
+ * Automatically generated from STM32F401R(D-E)Tx.xml
```
fpistm: Even if this is the same, please update the comment to the correct value:

```suggestion
 * Automatically generated from STM32F401C(B-C|D-E)Ux.xml
```

MCUdude:
Thanks for pointing this out. Should be OK now 🙂

MCUdude:
Do you want every new pinout request to be a new PR?

At the moment I'm working on adding:
* Generic F446Rx
* Generic F410Rx
* Nucleo F410RB

Is it OK if I combine the F410 pinouts into the same PR? Oh, I do have physical hardware for all the pinouts I have created (and will create). I do test it on actual hardware before submitting PRs 😉

fpistm:
This should be fine to have a PR per series.
About Generic F410Rx and Nucleo F410RB, it could be in the same PR.
The Nucleo F410RB should also use the same variant directory and then define the pin definition under switch `ARDUINO_NUCLEO_F410RB`.
I think about refactor some variants to point on the generic variant.

MCUdude:
> The Nucleo F410RB should also use the same variant directory and then define the pin definition under switch ARDUINO_NUCLEO_F410RB.
I think about refactor some variants to point on the generic variant.

I didn't quite understand how this would be done. I would assume you want to keep compatibility with the other 64-pin STM32F4 Nucleo boards? If so, _all_  F4 Nucleo 64 boards would have to point to the generic pinouts.

Another thing; would it be a good idea that _all_ "Generic F*" pinouts use the same pinout style to keep all generic pinouts as similar as possible? If so, the *Generic_F103Rx* would need a little work; I can do that if you want. BTW your approach at #851!

variants/Generic_F401Cx/variant.cpp
```diff
@@ -0,0 +1,217 @@
...
+  RCC_OscInitStruct.OscillatorType      = RCC_OSCILLATORTYPE_HSE;
+  if (bypass == 0) {
+    RCC_OscInitStruct.HSEState          = RCC_HSE_ON; // External 8 MHz xtal on OSC_IN/OSC_OUT
+  } else {
+    RCC_OscInitStruct.HSEState          = RCC_HSE_BYPASS; // External 8 MHz clock on OSC_IN
+  }
+
+  RCC_OscInitStruct.PLL.PLLState        = RCC_PLL_ON;
+  RCC_OscInitStruct.PLL.PLLSource       = RCC_PLLSOURCE_HSE;
+  RCC_OscInitStruct.PLL.PLLM            = 8;             // VCO input clock = 1 MHz (8 MHz / 8)
```
fpistm: As discussed in other issue, please, use the  `HSE_VALUE` to be more generic.

PR #877, opened 2020-01-20
Add generic F446Rx pinout

MCUdude:
Layout wise this pinout is more or less identical with `Genric_F401Rx` and `Generic F411Rx`, but of course designed to work with the F446Rx series.


variants/Generic_F446Rx/PeripheralPins.c
```diff
@@ -0,0 +1,375 @@
...
+
+//*** I2C ***
+
+#ifdef HAL_I2C_MODULE_ENABLED
+WEAK const PinMap PinMap_I2C_SDA[] = {
+  {PB_3,  I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
+  {PB_4,  I2C3, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C3)},
+  {PB_7,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  {PB_9,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  {PC_7,  FMPI2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_FMPI2C1)},
```
fpistm: Currently FMPI2C is not managed byt the core so the best way is to comment it.

variants/Generic_F446Rx/PeripheralPins.c
```diff
@@ -0,0 +1,375 @@
...
+};
+#endif
+
+#ifdef HAL_I2C_MODULE_ENABLED
+WEAK const PinMap PinMap_I2C_SCL[] = {
+  {PA_8,  I2C3, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C3)},
+  {PB_6,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  {PB_8,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  {PB_10, I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
+  {PC_6,  FMPI2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_FMPI2C1)},
```
fpistm: dito

variants/Generic_F446Rx/variant.cpp
```diff
@@ -0,0 +1,238 @@
...
+  RCC_OscInitStruct.OscillatorType      = RCC_OSCILLATORTYPE_HSE;
+  if (bypass == 0) {
+    RCC_OscInitStruct.HSEState          = RCC_HSE_ON; // External 8 MHz xtal on OSC_IN/OSC_OUT
+  } else {
+    RCC_OscInitStruct.HSEState          = RCC_HSE_BYPASS; // External 8 MHz clock on OSC_IN
+  }
+
+  RCC_OscInitStruct.PLL.PLLState        = RCC_PLL_ON;
+  RCC_OscInitStruct.PLL.PLLSource       = RCC_PLLSOURCE_HSE;
+  RCC_OscInitStruct.PLL.PLLM            = 8;             // VCO input clock = 1 MHz (8 MHz / 8)
```
fpistm: As discussed in other issue, the clock config have to use `HSE_VALUE` instead of hardcoded value

MCUdude: Sure! So is the solution provided in #880 for HSE/HSI OK?

MCUdude: If so, I can update the other pinouts with this fix as well

fpistm: Yes, I've made a comment on this in the dedicated PR

MCUdude: Great! I'm a bit busy at the moment, but I'll try to get this all done and tested within this week. OK? 🙂

fpistm: Sure. Thanks

MCUdude:
As a note, I'm using CubeMX's clock config tool together with some of the [mbed F446Rx source files](https://github.com/ARMmbed/mbed-os/blob/bf21bac52d053da55d2b7bbe6f58417425fbf8fb/targets/TARGET_STM/TARGET_STM32F4/TARGET_STM32F446xE/TARGET_NUCLEO_F446RE/system_clock.c) to determine the right settings. I've also tested the internal and the external oscillator option on my Nucleo F446RE board

variants/Generic_F446Rx/PeripheralPins.c
```diff
@@ -0,0 +1,375 @@
...
+ * Note: Commented lines are alternative possibilities which are not used per default.
+ *       If you change them, you will have to know what you do
+ * =====
+ */
+
+//*** ADC ***
+
+#ifdef HAL_ADC_MODULE_ENABLED
+WEAK const PinMap PinMap_ADC[] = {
+  {PA_0,  ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 0, 0)}, // ADC1_IN0
```
fpistm: Why choose only ADC1?

MCUdude: To make it similar to the other generic F4 pinouts. What benefits will it have to use all three ADCs?

fpistm: Currently none😅
Just this allows to test all peripherals.

MCUdude: From the AVR world that I'm most used with, I generally think that "less is more". So if there's no real benefit from using all ADC's, I prefer not to do it 😉 It would definitely be different if it provided a major advantage though!

variants/Generic_F446Rx/variant.cpp
```diff
@@ -0,0 +1,267 @@
...
+    return 0; // FAIL
+  }
+
+  // Activate the OverDrive to reach the 180 MHz Frequency
+  if (HAL_PWREx_EnableOverDrive() != HAL_OK) {
+    return 0; // FAIL
+  }
+
+  // Select PLLSAI output as USB clock source
+  PeriphClkInitStruct.PLLSAI.PLLSAIM = 8;
```
fpistm: This config is correct only for 8MHz HSE.
```suggestion
  PeriphClkInitStruct.PLLSAI.PLLSAIM = HSE_VALUE / 1000000L;
```

variants/Generic_F446Rx/variant.h
```diff
@@ -0,0 +1,144 @@
...
+#define NUM_ANALOG_INPUTS       16
+
+// On-board LED pin number
+#ifndef LED_BUILTIN
+#define LED_BUILTIN             PA5
+#endif
+#define LED_GREEN               LED_BUILTIN
+
+// On-board user button
+#define USER_BTN                PC13
```
fpistm: This will allow to override this definition
```suggestion
#ifndef USER_BTN                
#define USER_BTN                PC13
#endif
```

Issue #878, opened 2020-01-20
TIM1_IRQn wrong definition for STM32F410x


MCUdude:
Hi!
I'm playing around with my Nucleo F410RE board to see if I can get it working with this core.
I've created a generic pinout for it (that I haven't pushed yet), but I'm not able to compile code for it due to a bug (?) in the core system files.

This is the error I get whenever I compile something:
```
In file included from C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\libraries\SrcWrapper\src\stm32\timer.c:14:
C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\libraries\SrcWrapper\src\stm32\timer.c: In function 'getTimerUpIrq':
C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\cores\arduino/stm32/timer.h:54:19: error: 'TIM1_UP_TIM10_IRQn' undeclared (first use in this function); did you mean 'TIM1_BRK_TIM9_IRQn'?
   54 | #define TIM1_IRQn TIM1_UP_TIM10_IRQn
      |                   ^~~~~~~~~~~~~~~~~~
C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\libraries\SrcWrapper\src\stm32\timer.c:374:16: note: in expansion of macro 'TIM1_IRQn'
  374 |         IRQn = TIM1_IRQn;
      |                ^~~~~~~~~
C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\cores\arduino/stm32/timer.h:54:19: note: each undeclared identifier is reported only once for each function it appears in
   54 | #define TIM1_IRQn TIM1_UP_TIM10_IRQn
      |                   ^~~~~~~~~~~~~~~~~~
C:\Users\user\AppData\Local\Arduino15\packages\STM32\hardware\stm32\1.8.0\libraries\SrcWrapper\src\stm32\timer.c:374:16: note: in expansion of macro 'TIM1_IRQn'
  374 |         IRQn = TIM1_IRQn;
      |                ^~~~~~~~~
```

However: if I change this line in stm32f410rx.h to `TIM1_UP_TIM10_IRQn` it does compile without any errors:
[...]

Issue #876, opened 2020-01-19
STM32F411CE doesn't seem to work


TzOk83:
I've tried to upload BLINK example (I've changed LED_BUILTIN to PC13) using "Generic F411CE" Board profile to the "Black Pill" WeAct V2.0 board (STM32F411CEU6), and it doesn't seem to work. I've also tried turning the USB support as CDC Generic Serial, but there was no serial port after uploading this firmware. I've tried uploading either by SWD (ST-Link v2) or by USB-DFU. Both times the upload was successful, but program didn't seem to start (of course I have tried to reset and power cycle the MPU).

fpistm:
Hi @TzOk83 
it seems a clock config issue.
You can try to reconfigure it defining the `SystemClock_Config `at sketch level:
```
extern "C" void SystemClock_Config(void)
{
...
}
```

flatpedals:
I have a WeAct 1.3 411CE and I can upload in Arduino IDE and PlatformIO via DFU (STM32Cube prog) and Serial(USB to TTL dongle). 
ArduinoIDE: 'Blackpill F401CC' under 'Generic STM32F4 series'
![generic](https://user-images.githubusercontent.com/20503511/72688018-3569e480-3b0c-11ea-971d-9c10a16c9654.png)

PlatformIO: edited 'BlackPill 401CC (ST)' to a 'BlackPill 411CE (ST)'. 
[blackpill_f411ceJSON(PIO).zip](https://github.com/stm32duino/Arduino_Core_STM32/files/4083534/blackpill_f411ceJSON.PIO.zip)

As I said DFU and Serial are working, then there is a working Serial Port available (STM Serial) while uploaded sketch is running. 

I updated the core and the new 'Generic F411CC' showed up in board manager for a short while but now it's gone, so I can't select it to see if it's a problem vs the 'Blackpill F401CC'.

All good with this board untill now, maybe a STM32DUINO Bootloader would be nice so BOOT0 isn't required at every upload. I also soldered a Winbond 25q32jv and the SMD cap but never got to test that SPI memory.

fpistm:
@flatpedals 
the Generic F411Cx has been added recently and is merged in the master: #866

MCUdude:
[...]

PR #880, opened 2020-01-21
Add generic F410Rx pinout

MCUdude:
This pinout is formatted the same way as the Generic F401Rx, F411Rx and 446Rx pinouts, and is used with the STM32F410R8 and STM32F410RB chips

Note that currently, this pinout does not compile due to a bug somewhere in the core. More info in #878.


fpistm:
Requires #882

MCUdude:
Is there any way we can re-run Travis now that #882 is merged?

fpistm:
Yes, anyway, I will not merge this one as it is.
As I explained in #876, the default config for Generic boards should be HSI.
https://github.com/stm32duino/Arduino_Core_STM32/issues/876#issuecomment-576676462

MCUdude:
This is no different than the other generic pinouts. First, it tries HSE. If no crystal/clock is detected, it will switch to HSI. This must be the most compatible option?

fpistm:
I will also change for other generic.
The main issue is that the config is linked to an HSE value.
As I point in the other issue, the clock config for the HSE is based on a 8 MHz value, while per default the `HSE_VALUE` for F4 is 25 MHz. So this is not correct.

MCUdude:
So would you accept the generic pinouts to use an HSE based clock/crystal of 25 MHz and automatically switch to HSI if not present?

And wouldn't it be better to merge my pending PRs like they are at the moment, and fix all generic F4 variants in the same commit/PR afterwards? After all their clock config is pretty much identical.

fpistm:
> So would you accept the generic pinouts to use an HSE based clock/crystal of 25 MHz and automatically switch to HSI if not present?

No it is too much support. Goal is to provide an unique working clock config for every case.

> And wouldn't it be better to merge my pending PRs like they are at the moment, and fix all generic F4 variants in the same commit/PR afterwards? After all their clock config is pretty much identical.

Why adding codes we know they will be removed.

MCUdude:
> No it is too much support. Goal is to provide an unique working clock config for every case.

So for all users that would like to use HSE they will have to redefine `SystemClock_Config()`?
How about making it so that the `SetSysClock_PLL_HSE()` function could have it's external clock speed specified as a parameter? something like this:

```c
void SystemClock_Config(void)
{
  SetSysClock_PLL_HSE(1 /* no bypass */, 25 /* MHz */);
}
```

This would actually be quite neat!

fpistm:
> So for all users that would like to use HSE they will have to redefine `SystemClock_Config()`?

Yes, at least value and config will be aligned as they will have also to define the proper `HSE_VALUE` if different from the default one.



> How about making it so that the `SetSysClock_PLL_HSE()` function could have it's external clock speed specified as a parameter?

The main issue is that the `HSE_VALUE` is a define so used at build time not execution.

MCUdude:
> The main issue is that the HSE_VALUE is a define so used at build time not execution.

Yes, but HSE_VALUE is just a constant that isn't directly used by `SetSysClock_PLL_HSE(uint8_t bypass)`. Couldn't the default value be loaded like so?

```c
SetSysClock_PLL_HSE(0 /* no bypass*/, HSE_VALUE /* default val */);
```

fpistm:
@MCUdude 
I didn't talk about the `SystemClock_Config()`, but all HAL RCC and more.

MCUdude:
OK. So basically it isn't that easy to just pick a random crystal and expect it to work by only tweaking the clock config function.

Hear me out here:

* It would be nice to be able to use an external clock/crystal without having to re-define the clock config function. This can easily be achieved if we standardize on one frequency, I suggest 25 MHz since this is default anyways.
  - A crystal/clock is automatically detected, **This works well!**
* If the user wants to use a clock/crystal with a different frequency the HSE_VALUE can be defined in the build process (i.e in platformio.ini). This will not mess up HAL RCC since HSE_VALUE can be defined "externally":
https://github.com/stm32duino/Arduino_Core_STM32/blob/07bb4f09109f3c00ff52f849bd8684195e3682ce/system/STM32F4xx/system_stm32f4xx.c#L65-L69
* Clock configuration would be transparent for the average user, but "pro users" will still be able to change HSE_VALUE.

If you want, I can submit a PR for this.

MCUdude:
Wait, how is this not a problem on the F407 DISCO board (which I do own)? This board has an external 8 MHz crystal, but the HSE_VALUE is not redefined anywhere.

fpistm:
In that case the clock config is correct but I guess if you use some dedicated feature, this would not work as expected, example RTC using HSE.

MCUdude:
> In that case the clock config is correct but I guess if you use some dedicated feature, this would not work as expected, example RTC using HSE.

Got it. So like mentioned in my previous post, why not "standardize" on 25 MHz for all generic F4 targets, and let the user redefine HSE_VALUE if needed?

fpistm:
Even if you redefine HSE_VALUE, it is need to redefine the `SystemClock_Config()`

MCUdude:
> Even if you redefine HSE_VALUE, it is need to redefine the SystemClock_Config()

Yes, but this is very trivial:

```c
RCC_OscInitStruct.PLL.PLLM  = HSE_VALUE/1000000L;
```

fpistm:
> > Even if you redefine HSE_VALUE, it is need to redefine the SystemClock_Config()
> 
> Yes, but this is very trivial:
> 
> ```c
> RCC_OscInitStruct.PLL.PLLM  = HSE_VALUE/1000000L;
> ```

Yes in that case, this work for the requested SYSCLK but for some other freq no.

MCUdude:
I've just done a pretty deep dive into how the clocks are initialized to get timing correctly. I've used CubeMX's graphical window to figure out what all the clock config values might represent.

Working on my Nucleo F410RB, I was only able to get accurate timing (delay() in this case) when using an 8 MHz external clock. The PLL settings didn't even matter, the LED was blinking way too fast when using a 25 MHz clock even though the PLL settings should result in a 100 MHz SYSCLK.

It turned out that the HSE_VALUE isn't 25000000 by default (For F410 and F411 at least), **it is 8000000**!

I believe it gets its definition here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/0ff0a2a45181f9788852c0075bf50c01c6f8fed9/system/STM32F4xx/stm32f4xx_hal_conf_default.h#L97-L99

And not from here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/4e1c22561ae15c976a42699c97e08c53dbb3f4b7/system/STM32F4xx/system_stm32f4xx.c#L67-L69

So then, 8 MHz _is_ actually the default HSE_VALUE for the F4 series. Is this a bug or a feature?

MCUdude:
But this was at least a great learning experience for me 😀. I now feel I understand more about how the whole clock configuration works. In the process, I found out that the clock settings for some of the other generic variants I've created aren't really ideal. When we figure out what the default HSE_VALUE is, then I can submit a PR where I clean up some of this.

I suggest we stick with either 8 MHz (default value for practically all other STM32 families) or 25 MHz, and then pass -DHSE_VALUE=25000000UL as a build option for targets/boards that require a special HSE_VALUE, such as the F401 blackpill board.

fpistm:
Right, header is used before the source, so yes it is 8MHz. Forgot this.

The default is the one in the `stm32f4xx_hal_conf_default.h`.
You always can do a PR but I can't tell if it will be merged.
As said the simpliest way it to provide a well know config using the HSI or other internal clock for generic.
Then a board can use the generic variant and adding extra config under `ARDUINO_<BOARDNAME>` definition to customize it properly (pinmaping, clock config, hal conf,...)

MCUdude:
> As said the simpliest way is to provide a well know config using the HSI or other internal clock for generic.

In my opinion, it is even better to first try to use an external clock/oscillator and fall back to HSI if not present. This will cover most cases. For known boards that use another crystal, it's just a matter of adding -DHSE_VALUE={someval} in boards.txt.

If you refuse this, another option is to rewrite SystemClock_Config into this:

```c
WEAK void SystemClock_Config(void)
{
  if (SetSysClock_PLL_HSI() == 0) {
      Error_Handler();
  }
}
```

... and let the user override it in the sketch, like so, if he/she wants to use an external clock/crystal:

```c
void SystemClock_Config(void)
{
  if (SetSysClock_PLL_HSE(0) == 0) {
      Error_Handler();
  }
}
```

variants/Generic_F410Rx/PeripheralPins.c
```diff
@@ -0,0 +1,216 @@
...
+
+#ifdef HAL_I2C_MODULE_ENABLED
+WEAK const PinMap PinMap_I2C_SDA[] = {
+  //  {PB_3,  FMPI2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_FMPI2C1)},
+  {PB_3,  I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF9_I2C2)},
+  {PB_7,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  {PB_9,  I2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C1)},
+  //  {PB_9,  I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF9_I2C2)},
+  {PB_11, I2C2, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_I2C2)},
+  {PB_14, FMPI2C1, STM_PIN_DATA(STM_MODE_AF_OD, GPIO_NOPULL, GPIO_AF4_FMPI2C1)},
```
fpistm: FMPI2C is not managed by the core so it should be better to comment all.

PR #883, opened 2020-01-22
I2C wait end of stop before start

### LINKED ISSUE #864: i2c problem with release 1.8.0
I was using Arduino_Core_STM32 release 1.5.0 with MPU9250 library by hideakitai.  It was working fine until I upgraded to the latest release of 1.8.0.  With 1.8.0, i2c bus was not able to locate the AK8963 device.

I've tested Arduino_Core_STM32 release 1.5.0, 1.6.0, 1.6.1, 1.7.0 & 1.8.0.  Only 1.8.0 has problem with the MPU9250 library.  All other releases worked fine.

Hardware Configuration:
Blue pill (stm32f103c8) with i2c (PB11, PB10) connected to a MPU9250 [link](https://www.ebay.com/itm/MPU-9250-SPI-IIC-9-Axis-Attitude-Module-Gyro-Accelerator-Magnetometer-ST-/172263235059)

Sample sketch to reproduce the problem:

```
#include "MPU9250.h"
MPU9250 mpu;

void setup()
{
    Serial.begin(115200);
    Wire.begin();
    delay(2000);
    mpu.setup();
}

void loop()
{
    static uint32_t prev_ms = millis();
    if ((millis() - prev_ms) > 16)
    {
        mpu.update();

        Serial.print("roll, pitch, yaw: ");
        Serial.print(mpu.getRoll());
        Serial.print("\t");
        Serial.print(mpu.getPitch());
        Serial.print("\t");
        Serial.println(mpu.getYaw());

        prev_ms = millis();
    }
}
```
Output with AK9963 not found (with rel 1.8.0):
```
MPU9250 WHO AM I = 71
MPU9250 is online...
AK8963  WHO AM I = 0
Could not connect to AK8963: 0x0
roll, pitch, yaw: 0.00.0.00.-3.00
roll, pitch, yaw: 0.00.0.00.-3.00
```

Expected output when working (with release 1.5.0 to 1.7.0):
```
MPU9250 WHO AM I = 71
MPU9250 is online...
AK8963  WHO AM I = 48
Calibration values: 
X-Axis sensitivity adjustment value 1.20
Y-Axis sensitivity adjustment value 1.21
Z-Axis sensitivity adjustment value 1.16
X-Axis sensitivity offset value 0.00
Y-Axis sensitivity offset value 0.00
Z-Axis sensitivity offset value 0.00
roll, pitch, yaw: 170.50.-67.37.-155.35
roll, pitch, yaw: 171.00.-68.81.-155.60
```

ABOSTM:
**Summary**

Wait for end of I2C STOP before starting a new transfer. 

Fixes #864

### DIFF
--- a/system/Drivers/STM32F1xx_HAL_Driver/Src/stm32f1xx_hal_i2c.c
+++ b/system/Drivers/STM32F1xx_HAL_Driver/Src/stm32f1xx_hal_i2c.c
@@ -3492,6 +3492,27 @@ HAL_StatusTypeDef HAL_I2C_Master_Seq_Transmit_IT(I2C_HandleTypeDef *hi2c, uint16
       while (__HAL_I2C_GET_FLAG(hi2c, I2C_FLAG_BUSY) != RESET);
     }
 
+    /* Before any new treatment like start or restart, check that there is no pending STOP request */
+    /* Wait until STOP flag is reset */
+    count = I2C_TIMEOUT_BUSY_FLAG * (SystemCoreClock / 25U / 1000U);
+    do
+    {
+      count--;
+      if (count == 0U)
+      {
+        hi2c->PreviousState       = I2C_STATE_NONE;
+        hi2c->State               = HAL_I2C_STATE_READY;
+        hi2c->Mode                = HAL_I2C_MODE_NONE;
+        hi2c->ErrorCode           |= HAL_I2C_ERROR_TIMEOUT;
+
+        /* Process Unlocked */
+        __HAL_UNLOCK(hi2c);
+
+        return HAL_ERROR;
+      }
+    }
+    while (READ_BIT(hi2c->Instance->CR1, I2C_CR1_STOP) == I2C_CR1_STOP);
+
     /* Process Locked */
     __HAL_LOCK(hi2c);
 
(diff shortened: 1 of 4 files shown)


matthijskooijman:
At the risk of muddying the waters with an unrelated issue, this might be related ;-). We had an issue where a I2c write followed by an I2c read would result in a timeout for the read for some reason. Adding a delay between the two transactions fixed the problem, so this sounds like it might at least be related.

Here's a trace I made:

image

Green dots are starts, red dots are stops. You can see a stop ollowed by a start at the right, which starts the read transaction. However, it is cut short after the first clock pulse where a stop is done and then nothing happens (until a timeout after 100ms). What's notable is that the one clockpulse that happens is actually 6us rather than 5us as all others are. I suspected this might be the slave stretching the clock, but that did not make too much sense. Reading this, I wonder if maybe somehow the start is done when a (second?) stop is still pending, causing the transaction to start and immediately stop again.

If this report helps clarify things: Great. If it does not make sense, then just ignore it, we'll be more thoroughly investigating this issue on our hardware later (but we need to get a usable release out first next week). I will see if I can quicktly test whether this PR fixes our problem next week.

ABOSTM:
@matthijskooijman,
Thanks for sharing your experiment. Your description corresponds to what I observed while analyzing #864.
So this draft PR could help you (if you are on STM32F1 or STM32F4). You can test it to confirm it is the same issue.

ABOSTM:
Official patch ready for all concerned STM32 families F1, F2, F4 and L1.

matthijskooijman:
We've just tested this on our hardware, and it seems this fix indeed fixes the problem we were seeing as well. Thanks!

PR #884, opened 2020-01-22
[CDC]: Do not claim AT-protocol in CDC interface descriptor

matthijskooijman:
The CDC code presents itself as a virtual serial port. However, it also
sets the "bFunctionProtocol" value to 1, which means it supports
AT-commands, which is not actually the case. This might cause problems
with some software, such as ModemManager.

Originally, ModemManager would be very liberal with probing serial
devices, using a blacklist to prevent probing non-modems such as
Arduinos.

Since version 1.7.990, it has supported a "strict" mode where it tries to be
more restrained in what devices it probes. For CDC ACM devices, this
means it will only probe devices that claim to support AT-commands.
However, it also stopped applying the blacklist (intending to eventually
remove the blacklist), meaning it would end up probing these serial ports.

This new strict policy is not the upstream default, but is enabled in
Debian (since Buster) and Ubuntu (since bionic 18.04.2). Later versions
of ModemManager make an exception for single-port Serial devices making this
less of a problem, but best to also fix the descriptor.

For the equivalent change in the Arduino AVR core, see:
https://github.com/arduino/ArduinoCore-avr/pull/92


fpistm:
Thanks @matthijskooijman 
I will tr to review it soon.

PR #888, opened 2020-01-24
I2C multi instances: Remove static qualifier on TwoWire attributes

### LINKED ISSUE #699: Using both I2C lines on STM32F103C8 results in NACKs after some time.
I'm using the I2C on PB6/PB7 for communicating with another controller - ESP8266 as master and STM32F103C8 as slave and the I2C on PB10/PB11 as master to communicate with sensors. But after a while (random) the ESP8266 receives only Wire error "2:received NACK on transmit of address" on Wire.endTransmission() until I reset the F103C8. I'm guessing there it's some kind of interrupt race condition happening but I thought that the controller should be able to manage both I2C without interference ? If I disable the second I2C (PB10/11) everything works forever but I want the sensors :)

PS1: May be mask  I2C1 interrupt vector when communicating on I2C2 - could that be a 
part of the Wire library ?

PS2: This made it work with frequent errors to the master but at least the F103C8 doesn't die horribly :)
```
    Wire.end();
    err = compass.read();
    Wire.begin(I2C_DRIVEMOD);         // join i2c bus
    Wire.onRequest(requestEvent); // register event
    Wire.onReceive(receiveEvent); // register event
```

Of course that is insane way of doing it and it's crazy slow - I'm sure there should be a better way of using both I2C lines :)

PS3: I found the same problem described on the STM's site and it doesn't seem like anybody from STM cared to reply for a year ...

https://community.st.com/s/question/0D50X00009yHD6SSAW/issue-when-using-i2c1-and-i2c2-simultaneously-with-stm32cubefwf0v190

ABOSTM:
**Summary**

I2C buffers, as well as some other attributes of the TwoWire class,
should not be static (shared between instances),
otherwise multi instance I2C doesn't work properly.
For example when MCU is master on one instance and slave on the other,
Slave reception under interruption can happen in the middle of master transmission,
and there is conflict on the shared resources.

fixes #699

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -27,22 +27,6 @@ extern "C" {
 
 #include "Wire.h"
 
-// Initialize Class Variables //////////////////////////////////////////////////
-uint8_t *TwoWire::rxBuffer = nullptr;
-uint8_t TwoWire::rxBufferAllocated = 0;
-uint8_t TwoWire::rxBufferIndex = 0;
-uint8_t TwoWire::rxBufferLength = 0;
-
-uint8_t TwoWire::txAddress = 0;
-uint8_t *TwoWire::txBuffer = nullptr;
-uint8_t TwoWire::txBufferAllocated = 0;
-uint8_t TwoWire::txBufferIndex = 0;
-uint8_t TwoWire::txBufferLength = 0;
-
-uint8_t TwoWire::transmitting = 0;
-void (*TwoWire::user_onRequest)(void);
-void (*TwoWire::user_onReceive)(int);
-
 // Constructors ////////////////////////////////////////////////////////////////
 
 TwoWire::TwoWire()
@@ -81,6 +65,13 @@ void TwoWire::begin(uint8_t address, bool generalCall)
   txBufferLength = 0;
   resetTxBuffer();
 
+  rxBuffer = nullptr;
+  rxBufferAllocated = 0;
+  txAddress = 0;
+  txBuffer = nullptr;
+  txBufferAllocated = 0;
+  _i2c.__this = (void *)this;
+  user_onRequest = NULL;
   transmitting = 0;
 
   ownAddress = address << 1;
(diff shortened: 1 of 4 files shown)


libraries/Wire/src/Wire.cpp
```diff
@@ -403,44 +394,48 @@ void TwoWire::flush(void)
 }
 
 // behind the scenes function that is called when data is received
-void TwoWire::onReceiveService(uint8_t *inBytes, int numBytes)
+void TwoWire::onReceiveService(i2c_t *obj, uint8_t *inBytes, int numBytes)
```
fpistm: I think it will be more concise to use the obj to get the inBytes and numBytes as they are fields of obj:
```
inBytes = (uint8_t *) obj->i2cTxRxBuffer
numBytes = obj->slaveRxNbData
```

```suggestion
void TwoWire::onReceiveService(i2c_t *obj)
```

libraries/Wire/src/utility/twi.c
```diff
@@ -1026,7 +1026,7 @@ void HAL_I2C_ListenCpltCallback(I2C_HandleTypeDef *hi2c)
   /*  Previous master transaction now ended, so inform upper layer if needed
    *  then prepare for listening to next request */
   if ((obj->slaveMode == SLAVE_MODE_RECEIVE) && (obj->slaveRxNbData != 0)) {
-    obj->i2c_onSlaveReceive((uint8_t *) obj->i2cTxRxBuffer, obj->slaveRxNbData);
+    obj->i2c_onSlaveReceive(obj, (uint8_t *) obj->i2cTxRxBuffer, obj->slaveRxNbData);
```
fpistm: According to my first remarks:
```suggestion
    obj->i2c_onSlaveReceive(obj);
```

libraries/Wire/src/utility/twi.h
```diff
@@ -99,15 +99,16 @@ struct i2c_s {
...
   PinName sda;
   PinName scl;
   IRQn_Type irq;
 #if !defined(STM32F0xx) && !defined(STM32G0xx) && !defined(STM32L0xx)
   IRQn_Type irqER;
 #endif /* !STM32F0xx && !STM32G0xx && !STM32L0xx */
   volatile int slaveRxNbData; // Number of accumulated bytes received in Slave mode
-  void (*i2c_onSlaveReceive)(uint8_t *, int);
-  void (*i2c_onSlaveTransmit)(void);
+  void (*i2c_onSlaveReceive)(i2c_t *, uint8_t *, int);
```
fpistm:
```suggestion
  void (*i2c_onSlaveReceive)(i2c_t *);
```

libraries/Wire/src/utility/twi.h
```diff
@@ -138,8 +139,8 @@ i2c_status_e i2c_master_read(i2c_t *obj, uint8_t dev_address, uint8_t *data, uin
 
 i2c_status_e i2c_IsDeviceReady(i2c_t *obj, uint8_t devAddr, uint32_t trials);
 
-void i2c_attachSlaveRxEvent(i2c_t *obj, void (*function)(uint8_t *, int));
-void i2c_attachSlaveTxEvent(i2c_t *obj, void (*function)(void));
+void i2c_attachSlaveRxEvent(i2c_t *obj, void (*function)(i2c_t *, uint8_t *, int));
```
fpistm: ditto
```suggestion
void i2c_attachSlaveRxEvent(i2c_t *obj, void (*function)(i2c_t *));
```

libraries/Wire/src/utility/twi.c
```diff
@@ -948,7 +948,7 @@ i2c_t *get_i2c_obj(I2C_HandleTypeDef *hi2c)
   * @param  function: callback function to use
   * @retval None
   */
-void i2c_attachSlaveRxEvent(i2c_t *obj, void (*function)(uint8_t *, int))
+void i2c_attachSlaveRxEvent(i2c_t *obj, void (*function)(i2c_t *, uint8_t *, int))
```
fpistm:
```suggestion
void i2c_attachSlaveRxEvent(i2c_t *obj)
```

libraries/Wire/src/Wire.h
```diff
@@ -37,28 +37,28 @@ extern "C" {
...
     i2c_t _i2c;
 
-    static void (*user_onRequest)(void);
-    static void (*user_onReceive)(int);
-    static void onRequestService(void);
-    static void onReceiveService(uint8_t *, int);
+    void (*user_onRequest)(void);
+    void (*user_onReceive)(int);
+    static void onRequestService(i2c_t *);
+    static void onReceiveService(i2c_t *, uint8_t *, int);
```
fpistm:
```suggestion
    static void onReceiveService(i2c_t *);
```

libraries/Wire/src/Wire.cpp
ABOSTM: Done

libraries/Wire/src/utility/twi.c
ABOSTM: Done

libraries/Wire/src/Wire.h
ABOSTM: Done

libraries/Wire/src/utility/twi.c
ABOSTM: Done

libraries/Wire/src/utility/twi.h
ABOSTM: Done

libraries/Wire/src/utility/twi.h
ABOSTM: Done

PR #890, opened 2020-01-24
Added support for BlackPill F411CE

TzOk83:
Added support for WeAct V2.0 STM32F411CEU6 dev. board a.k.a. BlackPill F411CE.


TzOk83:
I have used CubeMX to generate these clock dividers, checked on real board with 25MHz oscillator, and confirmed that I2C and USB work.

Target FCPU is 100 MHz for that board.

ag88:
well, i work the numbers based on your pll dividers:
25 / 12 * 96 / 2 ~ 100 Mhz - fcpu
then for usb
25 / 12 * 96 / 4 ~ 50 Mhz
usb will fail to work as usb specifically requires 48 Mhz

TzOk83:
You're right I wasn't aware of this, yet USB was working fine with 50 MHz clock, at least for a short while, as I didn't do any long term tests. I have changed the clock according to your suggestions.

ag88:
thanks! i think users would appreciate it as i think even at 50 mhz, usb 'may' work.
the trouble would be that 2 mhz difference would cause sampling to shift. usb is 48 mhz 4x oversampled, so the actual signal is 12 mhz. this shifting would seem to work initially as they are pretty close. however, as time goes on, erratic results will follow as the 50 mhz intervals will keep drifting and finally interpret the signals incorrectly

fpistm:
@TzOk83, @ag88, 

@Jozo132 made also a PR to add this support.
I will review them but I'm waiting  the system clock config update of the Generic F411Cx.
Then this board should use the generic variant to add its support.

fpistm:
This one is replaced by #909

PR #893, opened 2020-01-27
[HardwareSerial] Fix ambiguous call of overloaded function

fpistm:
```
error: call of overloaded 'HardwareSerial(int, int)' is ambiguous`

    1 | HardwareSerial Serialx(0, 1);
      |                                  ^
In file included from cores/arduino/WSerial.h:5,
                 from cores/arduino/wiring.h:47,
                 from cores/arduino/Arduino.h:32,
                 from sketch_jan27b.ino.cpp:1:
cores/arduino/HardwareSerial.h:106:5: note: candidate: 'HardwareSerial::HardwareSerial(void*, bool)'
  106 |     HardwareSerial(void *peripheral, bool halfDuplex = false);
      |     ^~~~~~~~~~~~~~
cores/arduino/HardwareSerial.h:104:5: note: candidate: 'HardwareSerial::HardwareSerial(uint32_t, uint32_t)'
  104 |     HardwareSerial(uint32_t _rx, uint32_t _tx);
      |     ^~~~~~~~~~~~~~
```

When the Rx Pin is 0 then the compiler doesn't know which constructor
to use as 0 can be interpreted like NULL. So both of them are valid:
```
 HardwareSerial(uint32_t _rx, uint32_t _tx);
 HardwareSerial(void *peripheral, bool halfDuplex = false);
```

Defining an explicit type avoid this issue.

Issue reported here:
https://www.stm32duino.com/viewtopic.php?f=13&t=137


fpistm:
WiKi updated:
https://github.com/stm32duino/wiki/wiki/API#enable-half-duplex-mode

PR #902, opened 2020-01-30
HardwareTimer: Fix ARR and CCRx computation

### LINKED ISSUE #897: Off-by one error in HardwareTimer ticks values?
The HardwareTimer class applies a -1 to the timer tick values it sets. e.g. in `setOverflow()`:

https://github.com/stm32duino/Arduino_Core_STM32/blob/cc05df90dc013c0da20a3cd29f0f889862fbaabc/cores/arduino/HardwareTimer.cpp#L465-L474

And similarly in [`setCaptureCompare()`](https://github.com/stm32duino/Arduino_Core_STM32/blob/cc05df90dc013c0da20a3cd29f0f889862fbaabc/cores/arduino/HardwareTimer.cpp#L690).

This is a bit unexpected. When using `TICK_FORMAT`, I would expect that the value you pass is set into the hardware directly. One might say that the tick number passed to these function identifies the number of the tick and is 1-based and must be translated to 0-based when writing to the register, but I did not expect this. I passed 0 to this function only to find that the capture never triggered, apparently since -1 was written to the register.

Furthermore, when using `TICK_MICROSECONDS`, it would also make sense to pass 0. If you do that now, the -1 also causes an underflow and the compare never triggers. This also happens when you pass a non-zero but low value (depending on the value of the prescaler), so just passing a fixed low value is not always helpful either.

So: I would think that all these -1's should just be removed? Or is there any compelling argument I'm missing?

ABOSTM:
**Summary**
HardwareTimer: Fix ARR and CCRx computation

* protect setOverflow against underflow.
* remove -1 for MICROSEC_COMPARE_FORMAT, HERTZ_COMPARE_FORMAT and TICK_COMPARE_FORMAT.
* remove -1 in setCount() too
* Manage special case where ARR = MAX_RELOAD and 100% expected.

Fixes #897

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -87,7 +87,7 @@ HardwareTimer::HardwareTimer(TIM_TypeDef *instance)
 
   /* Configure timer with some default values */
   _timerObj.handle.Init.Prescaler = 0;
-  _timerObj.handle.Init.Period = 0xFFFF; // 16bit max value
+  _timerObj.handle.Init.Period = MAX_RELOAD;
   _timerObj.handle.Init.CounterMode = TIM_COUNTERMODE_UP;
   _timerObj.handle.Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
 #if defined(TIM_RCR_REP)
@@ -452,6 +452,7 @@ uint32_t HardwareTimer::getOverflow(TimerFormat_t format)
 void HardwareTimer::setOverflow(uint32_t overflow, TimerFormat_t format)
 {
   uint32_t ARR_RegisterValue;
+  uint32_t PeriodTicks;
   uint32_t Prescalerfactor;
   uint32_t period_cyc;
   // Remark: Hardware register correspond to period count-1. Example ARR register value 9 means period of 10 timer cycle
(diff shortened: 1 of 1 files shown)


cores/arduino/HardwareTimer.cpp
```diff
@@ -474,6 +474,10 @@ void HardwareTimer::setOverflow(uint32_t overflow, TimerFormat_t format)
       break;
   }
 
+  // In case ARR computation give 0xFFFFFFFF (equal (0 - 1) ) set ARR register to 0 (the smallest achievable period)
+  if (ARR_RegisterValue == 0xFFFFFFFF) {
+    ARR_RegisterValue = 0;
+  }
```
matthijskooijman: An alternative that might be easier to read here, would be to move the -1 down a bit. So let the switch set a `PeriodTicks` value without the -1, e.g.:

    PeriodTicks = (period_cyc / Prescalerfactor);

And then something like:

    if (PeriodTicks > 0) {
      // The register specifies the maximum value, so the period is really one tick longer
      ARR_RegisterValue = PeriodTicks - 1;
    } else {
      // But do not underflow in case a zero period was given somehow.
      ARR_RegisterValue = 0;
    }

Alternatively, you could maybe trigger an error when the period would end up as 0 ticks (which means that either 0 microsecond, 0 ticks, or a frequency higher than the input clock was specified). But maybe just using a register value of 0 ticks (i.e. run as fast as possible even if not as fast as requested, as your PR does) is the better approach.

ABOSTM: I prefer 1st option as error check, is not always performed at application level.
Done

PR #908, opened 2020-02-03
Add generic F410Cx pinout

MCUdude:
This pinout is practically identical to the Generic F401Cx and F411Cx variants


MCUdude:
Damnit! Something weird happened to boards.txt. Any idea how this can be solved? All I want to do is to add the two targets

fpistm:
Update your fork with the original repo and then rebase on top.

PR #921, opened 2020-02-06
Fix SPI mapping for some generic F4 variants

MCUdude:
Make use of SPI3 on PA15, PB3, PB4, and PB5. This will make the generic F4 pinouts more similar, so it will be easier to migrate from one chip to another without breaking code changes due to the pin mapping.

F410Rx and F410Cx do not have SPI3 on these pins, but rather SPI1. These will stay untouched.


MCUdude:
Thanks for pointing this out. I totally forgot about this. A fix is on its way!

PR #922, opened 2020-02-06
Update variant.cpp

lab-mathias-claussen:
Removed entry of PA_3 for where PA_2 should be. Makes PA_2 now work

**Summary**

* [ x ] Fixes broken definition for PA_2 in variant Elektor LoRa Node Core F072

Due to a typo in variant.cpp PA_2 is not working and may lead also to undefined behaviour


fpistm:
Thanks @lab-mathias-claussen 
I will merge when the CI will be done.

PR #924, opened 2020-02-06
Add initial support for PlatformIO ecosystem

valeros:
This PR adds build script for @platformio and a new Travis CI environment for testing internal libraries using `platformio ci` command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,29 @@ matrix:
           else
             echo "Coding style check OK";
           fi
+
+#
+# PlatformIO test
+#
+  - env:
+      - NAME=PlatformIO
+      - CMSIS_VERSION=5.5.1
+    install:
+      # Install PlatformIO
+      - pip install -U platformio
+      # Install the development version of ststm32 platform
+      - platformio platform install https://github.com/platformio/platform-ststm32.git
+      # Prepare framework for CI
+      - python -c "import json; import os; fp=open(os.path.expanduser('~/.platformio/platforms/ststm32/platform.json'), 'r+'); data=json.load(fp); data['packages']['framework-arduinoststm32']['version'] = '*'; fp.seek(0); fp.truncate(); json.dump(data, fp); fp.close()"
+      - sudo ln -sf $TRAVIS_BUILD_DIR ~/.platformio/packages/framework-arduinoststm32
+      # Download and unpack CMSIS package
+      - wget https://github.com/stm32duino/ArduinoModule-CMSIS/releases/download/$CMSIS_VERSION/CMSIS-$CMSIS_VERSION.tar.bz2
+      - tar -xvjf CMSIS-$CMSIS_VERSION.tar.bz2
+    before_script:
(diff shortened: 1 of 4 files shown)


fpistm:
Hi @valeros and thanks

About the CI,did you plan or is it already done to move on GitHub actions ? I'm currently working to move our build and astyle checks on it.
Moreover, is it OK for you if there is an failed I ask you some support ?

valeros:
Hi @fpistm !
> Moreover, is it OK for you if there is an failed I ask you some support ?

Sure thing, ping me if there is something wrong with our build.

> About the CI, did you plan or is it already done to move on GitHub actions ?

It should't be hard to port PlatformIO env to Github Actions, we probably can create a standalone workflow for it. Let me know when you're ready to switch to Github Actions.

Thanks!

fpistm:
@valeros 
Looking at your script, it seems several things are hardcoded.
I guess this is the same for all other supported core but I guess it should be possible to move some of them generic by parsing the boards.txt and also the platform.txt ?
Is it something considered ?

valeros:
Hi @fpistm ! What would you like to parse from `boards.txt` ,`platform.txt`? Anyway, feel free to modify whatever you think best.

fpistm:
> Hi @fpistm ! What would you like to parse from `boards.txt` ,`platform.txt`? Anyway, feel free to 
modify whatever you think best.

@valeros 

I talk about those kind of hardcoded stuff:
https://github.com/stm32duino/Arduino_Core_STM32/pull/924/files#diff-a96e736475cb79393ad72ec653311807R182-R206

and several other hardcoded definition

ivankravets:
Hi @fpistm ,

We have a feature request for this https://github.com/platformio/platformio-core/issues/969.
There is a technical problem here. Arduino's formats are not intended for advanced embedded development. It's very difficult to parse all these flags which depending on Arduino IDE Menu. 

PlatformIO does not work with CC or C++ to build commands. We need a strict declaration of each flag in its special environment scope. Our users control the whole build process, they can add or remove any flag dynamically from a specific scope. The same, we have Library Dependency Finder which is built on top of virtual C Preprocessor based on Python. This is why PlatformIO is super fast and requires minimal memory and CPU resources. You can check:

- https://docs.platformio.org/en/latest/projectconf/section_env_build.html
- https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html

-------

We can actually use intermediate interface (JSON file or etc.) where you will provide all flags related to GCC Preprocessor scopes. Then you can use some script which will compress all of them into the 1 line for Arduino. 

P.S: We don't have the aim to reinvent a wheel. We resolve the problems around the project building and maintaining.

fpistm:
@ivankravets 
thanks for the explanation.
That was just a question 😉 Maybe a JSON file would be an intermediate step.

ivankravets:
Thanks !!! Please do not hesitate to contact us if something builds with Arduino IDE and does not build with PlatformIO. We will help ASAP.

PR #928, opened 2020-02-10
Add generic F413C/F423C pinout variant

MCUdude:
This pinout variant adds to the list of similarly formatted generic F4xxC pinout variants.


MCUdude:
I'd probably make a pinout for a 100-pin variant as well. I do have an F407 Disco board, so I might as well make a generic F407/417Vx pinout.

The reason why I'd like to make a generic 100-pin F4 variant is that if others want to make variants for other 100-pin F4's they can just follow the rules I've made 🙂

fpistm:
> I'd probably make a pinout for a 100-pin variant as well. I do have an F407 Disco board, so I might as well make a generic F407/417Vx pinout.
> 

I guess it should be fine to extend genpinmap script to generate the skeleton (i.e. prefilled variant.*)
It is possible as it is aware of all available pins.

uzi18:
so i'm waiting for F407ZET6 :)

PR #937, opened 2020-02-13
Add generic F407Vx/F417Vx pinout variant

MCUdude:
Pinout is based on the F446Rx pinout, but with different peripheral mapping and extra IOs.


MCUdude:
@fpistm I see now that STM32F405VG/415VG is identical to the F407Vx/417Vx, except that the F405/415 does not have ethernet. I verified this by comparing their (unmodified) PeripheralPins.c files.

The ethernet peripherals are wrapped around HAL_ETH_MODULE_ENABLED:

``` c
// Definition from F407Vz file
#ifdef HAL_ETH_MODULE_ENABLED
WEAK const PinMap PinMap_Ethernet[] = {
  {PA_0,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_CRS
  {PA_1,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_REF_CLK|ETH_RX_CLK
  {PA_2,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_MDIO
  {PA_3,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_COL
  {PA_7,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_CRS_DV|ETH_RX_DV
  {PB_0,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_RXD2
  {PB_1,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_RXD3
  {PB_5,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_PPS_OUT
  {PB_8,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TXD3
  {PB_10, ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_RX_ER
  {PB_11, ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TX_EN
  {PB_12, ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TXD0
  {PB_13, ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TXD1
  {PC_1,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_MDC
  {PC_2,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TXD2
  {PC_3,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TX_CLK
  {PC_4,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_RXD0
  {PC_5,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_RXD1
  {PE_2,  ETH, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF11_ETH)}, // ETH_TXD3
  {NC,    NP,    0}
};
#endif
```


Since this is the only difference, can this pinout be used with the 405/415Vx as well, without any modifications? If yes, what should the pinout be called? Generic_F4x5Vx_F4x7Vx?

MCUdude:
After playing around with CubeMX, it seems like the F427/437 can run at 180 MHz (the others at 168 MHz), plus the F427/437 also have slightly different clock setup settings.

fpistm:
> After playing around with CubeMX, it seems like the F427/437 can run at 180 MHz (the others at 168 MHz), plus the F427/437 also have slightly different clock setup settings.

Well what about the STM32F401V which exist and have less pins...
Ideally `Generic_4x[5-7]Vx` could be better I guess but I would avoid using special char.
Maybe `Generic_4x5-7Vx` is more explicit?

MCUdude:
> Well what about the STM32F401V which exist and have less pins...

It's also about on what pins the peripherals are routed to. I'll see what I can do. I guess I can use
 `#ifdef STM32F405xx`,  `#ifdef STM32F415xx`,  `#ifdef STM32F407xx`,  `#ifdef STM32F417xx`,  `#ifdef STM32F427xx` and `#ifdef STM32F437xx`, to separate them where needed?
Atleast that's what's done here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/7999a46413853f819cbc462eb8b279935d0c6002/variants/Generic_F103Rx/PeripheralPins.c#L29-L35

> Maybe Generic_4x5-7Vx is more explicit?

Sounds good to me!

fpistm:
Well, up to you sometimes it is too much hard to read if too many define.

FYI, I have rework src tree of the generated files to ease access.
https://github.com/stm32duino/Arduino_Tools/tree/master/src/genpinmap/Arduino

I'm also thinking about reorder `variants/` folder in the core because they are more and more variants.
`variants/<menu id>/<variant name>`

ex:
```
GenF4/Generic_F412Cx
     /Generic_F412Rx
     /... 
```

MCUdude:
> FYI, I have rework src tree of the generated files to ease access.
https://github.com/stm32duino/Arduino_Tools/tree/master/src/genpinmap/Arduino

I've been using this for at least a month now 😉 Would be difficult to do what I do without it.

OK, I see now that F429/F439 and F469/479 are also similar. Maybe I should divide this up. Especially if you're planning to group variants into folders. How about:

Generic_F4x5RG
Generic_F4x7Rx
Generic_F4x9Rx
?
This will make it easier to maintain.

fpistm:
> Especially if you're planning to group variants

This should not change the "problem".
About the group, it should not be so easy due to Arduino specification:

> {build.variant.path}      - The path to the selected board variant folder
>                             (for example hardware/arduino/avr/variants/micro)

MCUdude:
But from my point of view, it is certainly easier if we have three separate variants instead of one giant one. If you agree with this, this PR is already ready to be merged.

uzi18:
@MCUdude will you add also F407Zx with more pins?

MCUdude:
@uzi18 maybe later

fpistm:
Agreed.
After a quick test, as I thought, it will not be possible to change variants/ folder structure.
Arduino automatically include all sources files from the specified `{build.variant}` and it will be too risky to change it.

stas2z:
AFAIK upload.maximum_data_size=196608 is wrong here
these chips have 128K RAM + 64K CCM so it will not work imo

stas2z:
> AFAIK upload.maximum_data_size=196608 is wrong here
> these chips have 128K RAM + 64K CCM so it will not work imo

yep, just checked, it stuck ofter uploading
ive checked one of my cube projects and found this one

/* Memories definition */
MEMORY
{
  CCMRAM	(xrw)	: ORIGIN = 0x10000000,	LENGTH = 64K
  RAM	(xrw)	: ORIGIN = 0x20000000,	LENGTH = 128K
  FLASH	(rx)	: ORIGIN = 0x8000000,	LENGTH = 1024K
}
so additional 64K CCM block is not after main 128K block

BennehBoy:
I think both the PRNTR board variants also make this mistake of including CCMRAM in data size.  But I guess that needs a separate issue raising.

fpistm:
You're right @stas2z 
I didn't generate the LD script to check. 😕 
Seems @MCUdude use the same for all. It will require to check all GenF4 added.
Thanks for pointing this 👍

MCUdude:
Sorry about this. I checked the other variants I've added recently, and they do not have CCM, only this target.

fpistm:
No worries.
I should have seen it 😖

PR #942, opened 2020-02-17
Initial version of Globallogic ESK support

shchers:
Globallogic Embedded Starterkit is an Open Source
and Open Hardware solution that is based on
STM32F4Discovery board.

Schematics and BSP is available under this link:
https://github.com/GlobalLogicEdu/GL-SK-BSP

Signed-off-by: Sergey Shcherbakov <shchers@gmail.com>


stas2z:
Globallogic.menu.pnum.GLOBALLOGIC_ESK.upload.maximum_data_size=196608
wrong here
see https://github.com/stm32duino/Arduino_Core_STM32/pull/937#issuecomment-586575358

and, as i can understand it's not a devboard, but shield for F4Disco
does it really require separate variant defined?

fpistm:
Hi @shchers 
I'm agreed with @stas2z, it is a derivative of an existing board.
Yo should better create a library for it like it is done for all the X-Nucleo-Boards.

You can redefine all PinMap arrays as it is weak and define all alias for your pin naming.

shchers:
Hi @fpistm 
Right now it's derivative of STM32F4Disco as fast line to bring new education board for students and scholars.  
But the idea is to build new cost effective STM32F407-based board.
Of course, if adding this board will be rejected, so lib or parallel tree will be the single option for us.

@stas2z 
Thank you for pointing to your comment in parallel thread. You are absolutely right!
For some reason CCM and RAM was mixed together. It's definitely mistake.
I'll fix it in this patch, but the rest of configs need to be updated also.

With best regards,
Sergey

fpistm:
>  But the idea is to build new cost effective STM32F407-based board.

In this case it could be added in the Generic F4 menu.
But not while it is a "shield", in that case library is one option or this: 
https://github.com/stm32duino/wiki/wiki/Custom-board-based-on-a-core

Anyway I guess a library is the easiest way and match to your "shield".

PR #953, opened 2020-02-25
Wire: rxBuffer should be initialized before using it

ABOSTM:
**Summary**

Wire: rxBuffer should be initialized before using it
rxBuffer is used within `resetRxBuffer();`
Idem for txBuffer


uzi18:
nice catch

PR #963, opened 2020-03-03
Nucleo-144 F756ZG

hockeymikey:
**Summary**

Adds the Nucleo-144 F756ZG variant board. https://www.st.com/en/evaluation-tools/nucleo-f756zg.html

**Validation**

Tested on the board I had and blink ran fine. Would be great for someone to just double check my work since it's my first rodeo.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1736,8 +1736,8 @@ LoRa.menu.upload_method.dfuMethod.upload.tool=stm32CubeProg
 3dprinter.menu.pnum.FYSETC_S6.build.product_line=STM32F446xx
 3dprinter.menu.pnum.FYSETC_S6.build.variant=FYSETC_S6
 3dprinter.menu.pnum.FYSETC_S6.build.cmsis_lib_gcc=arm_cortexM4lf_math
-3dprinter.menu.pnum.MALYANM300_F070CB.build.flash_offset=0x10000
-3dprinter.menu.pnum.MALYANM200_F070CB.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET={build.flash_offset}
+3dprinter.menu.pnum.FYSETC_S6.build.flash_offset=0x10000
+3dprinter.menu.pnum.FYSETC_S6.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -DVECT_TAB_OFFSET={build.flash_offset}
 
 # Upload menu
 3dprinter.menu.upload_method.swdMethod=STM32CubeProgrammer (SWD)


fpistm:
Hi @hockeymikey 
Thanks for the PR.

variants/NUCLEO_F756ZG/variant.cpp
```diff
@@ -0,0 +1,218 @@
...
+  79, //A1
+  80, //A2
+  81, //A3
+  82, //A4
+  83, //A5
+  84, //A6
+  85, //A7
+  86, //A8
+  87, //A9
+  11, //A10
```
ABOSTM: analogInputPin[] to be updated after taking into account variant.h comment.

variants/NUCLEO_F756ZG/variant.cpp
```diff
@@ -0,0 +1,218 @@
...
+  *            SYSCLK(Hz)                     = 216000000
+  *            HCLK(Hz)                       = 216000000
+  *            AHB Prescaler                  = 1
+  *            APB1 Prescaler                 = 4
+  *            APB2 Prescaler                 = 2
+  *            HSE Frequency(Hz)              = 16000000
+  *            PLL_M                          = 8
+  *            PLL_N                          = 216
+  *            PLL_P                          = 2
+  *            PLL_Q                          = 9
```
ABOSTM: It looks like PLL_Q is 2 see code below. Comment is not up to date.

variants/NUCLEO_F756ZG/PeripheralPins.c
```diff
@@ -0,0 +1,488 @@
...
+  {NC,    NP,    0}
+};
+#endif
+
+//*** PWM ***
+
+#ifdef HAL_TIM_MODULE_ENABLED
+WEAK const PinMap PinMap_PWM[] = {
+  {PA_0,  TIM2,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF1_TIM2, 1, 0)}, // TIM2_CH1
+  {PA_0,  TIM5,   STM_PIN_DATA_EXT(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF2_TIM5, 1, 0)}, // TIM5_CH1
```
ABOSTM: dito.
balance for timer should take into account instance of timer + channel (try to have all instance/channel available at least once):
TIM5_CH1 is available only on PA_0 whereas TIM2_CH1 is also available on PA_15. So it is better to comment `// {PA_0, TIM2, ...`
and keep `{PA_0,  TIM5, ...`

The same for the all Timer pins

variants/NUCLEO_F756ZG/variant.cpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+  /** Initializes the CPU, AHB and APB busses clocks 
+  */
+  RCC_ClkInitStruct.ClockType = RCC_CLOCKTYPE_HCLK|RCC_CLOCKTYPE_SYSCLK
+                              |RCC_CLOCKTYPE_PCLK1|RCC_CLOCKTYPE_PCLK2;
+  RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_PLLCLK;
+  RCC_ClkInitStruct.AHBCLKDivider = RCC_SYSCLK_DIV1;
+  RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV4;
+  RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV2;
+
```
ABOSTM: Missing configuration of USB clock. Must be 48MHz.

variants/NUCLEO_F756ZG/variant.h
```diff
@@ -0,0 +1,182 @@
...
+#define PG14 1
+#define PF15 2
+#define PE13 3
+#define PF14 4
+#define PE11 5
+#define PE9  6
+#define PF13 7
+#define PF12 8
+#define PD15 9
+#define PD14 A10
```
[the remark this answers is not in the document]
hockeymikey: Well then the F767ZI variant must be wrong too. It has PD14 in there and they should be about identical.

variants/NUCLEO_F756ZG/variant.cpp
hockeymikey: How? Clock is set with RCC_OscInitStruct.PLL.PLLQ and that is 7. It's the same as F767ZI.

variants/NUCLEO_F756ZG/variant.h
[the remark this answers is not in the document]
fpistm: Right the F767ZI is wrong. When I've update for the new analog pins definition, I based my update on the comment which is wrong. We will fix it.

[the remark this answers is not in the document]
fpistm: This is now fixed

variants/NUCLEO_F756ZG/variant.cpp
ABOSTM: Generally speaking you can use CubeMX to generate code initialization and in this case code for clock configuration. see https://github.com/stm32duino/wiki/wiki/Add-a-new-variant-(board)#5---system-clock-configuration
Nevertheless if you look at NUCLEO_F767ZI/variant.cpp, you can found the following code at the end of `SystemClock_Config()`

```
  PeriphClkInitStruct.PeriphClockSelection = RCC_PERIPHCLK_CLK48;
  PeriphClkInitStruct.Clk48ClockSelection = RCC_CLK48SOURCE_PLL;
  if (HAL_RCCEx_PeriphCLKConfig(&PeriphClkInitStruct) != HAL_OK) {
    _Error_Handler(__FILE__, __LINE__);
  }
```
Also, in NUCLEO_F767ZI 
`  RCC_OscInitStruct.PLL.PLLQ = 9;`
whereas in this PR for \NUCLEO_F756ZG you set 
`  RCC_OscInitStruct.PLL.PLLQ = 2;`

variants/NUCLEO_F756ZG/PeripheralPins.c
```diff
@@ -24,46 +24,46 @@
...
-  {PA_1,  ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC1_IN1
-  {PA_1,  ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC2_IN1
-  {PA_1,  ADC3,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC3_IN1
-  {PA_2,  ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // ADC1_IN2
-  {PA_2,  ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // ADC2_IN2
-  {PA_2,  ADC3,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 2, 0)}, // ADC3_IN2
+  //{PA_0,  ADC3,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 0, 0)}, // ADC3_IN0
+  //{PA_1,  ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC1_IN1
+  //{PA_1,  ADC2,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC2_IN1
+  //{PA_1,  ADC3,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 1, 0)}, // ADC3_IN1
```
ABOSTM: This is true that we comment some lines to save space in Flash, bu we keep uncommented 1 line per pin, specially when there is quite a large Flash of 1MB
So one line for PA_1 should be uncommented. 
The same for PA_2 ...

variants/NUCLEO_F756ZG/PeripheralPins.c
```diff
@@ -199,17 +199,22 @@ WEAK const PinMap PinMap_PWM[] = {
 #ifdef HAL_UART_MODULE_ENABLED
 WEAK const PinMap PinMap_UART_TX[] = {
   {PA_0,  UART4,   STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF8_UART4)},
-  {PA_2,  USART2,  STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF7_USART2)},
-  {PA_9,  USART1,  STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF7_USART1)},
+  //{PA_2,  USART2,  STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF7_USART2)},
+  //{PA_9,  USART1,  STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF7_USART1)},
+  {PA_15, UART7,   STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF12_UART7)}, // D20
```
ABOSTM: It looks like you mixed with some F767:
On F756 PA_15 doesn't have UART7 alternate function.
I suggest that you restart from original generated PeripheralPins.c, and then comment the useless pin (keep 1 line per pin)
Note: It causes compilation error highlighted by Continuous Integration

variants/NUCLEO_F756ZG/PeripheralPins.c
hockeymikey: So which one should be uncommented?

variants/NUCLEO_F756ZG/PeripheralPins.c
hockeymikey: This is the original generated PeripheralPins.c, I only uncommented stuff.

variants/NUCLEO_F756ZG/variant.cpp
hockeymikey: Not even sure what I'm supposed to do.

variants/NUCLEO_F756ZG/PeripheralPins.c
fpistm: No, this is not the original file, this is the original one:
https://github.com/stm32duino/Arduino_Tools/blob/master/src/genpinmap/Arduino/STM32F7/STM32F756ZGTx/PeripheralPins.c
and as you can see PA_15 has not UART7.

variants/NUCLEO_F756ZG/variant.cpp
hockeymikey: I did use CubeMX that's the thing, lol. I switched it over anyways.

variants/NUCLEO_F756ZG/variant.cpp
ABOSTM: In variant.h I commented that PD14 is not an analog pin so it could not be named A10 (A means Analog in this define) . In the same way   `11, //A10` is not valid as A10 is not valid.

variants/NUCLEO_F756ZG/variant.cpp
ABOSTM: CubeMx only do what it is requested to. If you don't request USB clock, it will not configure USB clock.

variants/NUCLEO_F756ZG/variant.h
```diff
@@ -0,0 +1,182 @@
...
+#define PF1  69
+#define PF2  70
+// 71 is PA7 (11)
+// 72 is NC
+#define PB7  73 // LED_BLUE
+#define PB14 74 // LED_RED
+#define PC13 75 // USER_BTN
+#define PD9  76 // Serial Rx
+#define PD8  77 // Serial Tx
+#define PA3  A0
```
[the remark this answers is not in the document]
hockeymikey: Where can I find this?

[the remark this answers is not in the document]
ABOSTM: You can find this in Board user manual
https://www.st.com/resource/en/user_manual/dm00244518-stm32-nucleo144-boards-stmicroelectronics.pdf
Warning1: this user manual is valid for different board, so take care to check pin assigment for F756 (Table 21)
Warning2: take care to not redefine pin already defined for Arduino connector.
Note: there is no default Arduino naming for the morpho pins (Dx) so it is up to you. It should continue the previous numerotation, and then generally it is done in the order of the UserManual table

ABOSTM:
Hi @hockeymikey ,
I take the liberty to continue your Pull Request, specially to merge the support of NUCLEO_F746ZG  #999 which is very similar to NUCLEO_F756ZG and can use the same variant.
The only difference between both is that F756 embed cryptographic accelerators AES and CRYP,     that are not currently supported by Arduino.

variants/NUCLEO_F756ZG/variant.cpp
hockeymikey: Can't make heads or tails looking at the data sheets and even comparing against other datasheets and trying to reverse engineer in a way. Am I looking at the wrong documents? Or what should I be looking for that determines this? https://www.st.com/en/microcontrollers-microprocessors/stm32f756zg.html

ABOSTM: This as nothing to do with datasheet, it is linked to Arduino way of working.
All pins with ADC and DAC capabilities should be declared as Analog Ax. Thus it correspond to all pins listed in `PinMap_ADC[]` and `PinMap_DAC[]`. PD14 is not listed in those arrays.

variants/NUCLEO_F756ZG/variant.h
```diff
@@ -0,0 +1,431 @@
...
+#define PG1  64
+#define PG0  65
+#define PD1  66
+#define PD0  67
+#define PF0  68
+#define PF1  69
+#define PF2  70
+// 71 is PA7 (11)
+// 72 is NC/*
+  Copyright (c) 2011 Arduino.  All right reserved.
```
ABOSTM: Duplicated code: there is already copyright at the beginning of the file

variants/NUCLEO_F756ZG/variant.cpp
```diff
@@ -0,0 +1,451 @@
...
+  }
+  /** Activate the Over-Drive mode
+  */
+  if (HAL_PWREx_EnableOverDrive() != HAL_OK) {
+    Error_Handler();
+  }
+  /** Initializes the CPU, AHB and APB busses clocks
+  */
+  RCC_ClkInitStruct.ClockType = RCC_CLOCKTYPE_HCLK | /*
+  Copyright (c) 2011 Arduino.  All right reserved.
```
ABOSTM: Duplicated code: there is already copyright at the beginning of the file

fpistm:
@hockeymikey 
I've made a mistake while updating your PR.
I've sync your dork master with the original repo that's why this PR was closed automatically.
Anyway, I've submitted the clean PR in #1014.

PR #976, opened 2020-03-09
[CI] Build built-in libraries

fpistm:
Since {whole-archive} option has been removed the CI is  not fully relevant. So need to explicitly used basic functions of the built-in library.
    
The arduino-cli script have been updated to handle properly  region overflowed which is not a real issue while there is no other issue.

Fix 2 issues raised using the updated BareMinimum.ino:
* [F4] Fix HAL_RCC_USART3 definition for STM32F412Cx
* [Generic F410xx] Fix wrong TIMER_SERVO definition

### DIFF
--- a/variants/Generic_F410Cx/variant.h
+++ b/variants/Generic_F410Cx/variant.h
@@ -99,9 +99,8 @@ extern "C" {
 #define PIN_WIRE_SCL            PB8
 
 // Timer Definitions
-// Use TIM6/TIM7 when possible as servo and tone don't need GPIO output pin
 #define TIMER_TONE              TIM6
-#define TIMER_SERVO             TIM7
+#define TIMER_SERVO             TIM11
 
 // UART Definitions
 #define SERIAL_UART_INSTANCE    2 // Connected to ST-Link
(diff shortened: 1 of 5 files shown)


variants/Generic_F410Cx/variant.h
```diff
@@ -99,9 +99,8 @@ extern "C" {
 #define PIN_WIRE_SCL            PB8
 
 // Timer Definitions
-// Use TIM6/TIM7 when possible as servo and tone don't need GPIO output pin
 #define TIMER_TONE              TIM6
-#define TIMER_SERVO             TIM7
+#define TIMER_SERVO             TIM5
```
ABOSTM: I would prefer TIM11 which is the most basic timer (after TIM6/7). So that we leave more complex timers for user.

variants/Generic_F410Rx/variant.h
```diff
@@ -100,9 +100,8 @@ extern "C" {
 #endif
 
 // Timer Definitions
-// Use TIM6/TIM7 when possible as servo and tone don't need GPIO output pin
 #define TIMER_TONE              TIM6
-#define TIMER_SERVO             TIM7
+#define TIMER_SERVO             TIM5
```
ABOSTM: dito

fpistm: done

variants/Generic_F410Cx/variant.h
fpistm: done

PR #994, opened 2020-03-17
Included OpenOCD as a programmer

bdmihai:
OpenOCD is missing as a programmer tool from the Arduino STM32 board. I think this makes a good addition to the STM32 core.

There are some workarounds needed to get this fully working... as it is highly dependent on OpenOCD.

I have used the ST-Link V2 (cheap clone) debugger on linux with one of the latest openocd version (self compiled). 

OpenOCD config file used: https://github.com/ntfreak/openocd/blob/master/tcl/board/stm32f103c8_blue_pill.cfg

Other config files are available if needed.


fpistm:
Hi @bdmihai 

This would be fine to have it added and already think about that.

> There are some workarounds needed to get this fully working... as it is highly dependent on OpenOCD.

You're right. That's why I will not merge this PR as it it not enough generic:
* Linux only
* hardcoded  path to /usr/bin
* hard config (interface and target

PR #996, opened 2020-03-19
Support a .noinit section for variables

matthijskooijman:
This is inspired by the linker sripts on the AVR architecture, which
support a .noinit section (or any section starting with .noinit,
actually) for variables that should be allocated an address in RAM, but
not be initialized to any particular value (not even zero) on startup.
These can then be used to remember values across resets.

From the sketch perspective, this works exactly the same as on AVR: Just
annote a global variable with `__attribute__((__section__(".noinit")))`
and it will have an unpredictable value on power-up and retain its value
during resets.

To implement this without having to change all board-specific linker
scripts, the linker commandline is changed to pass the board-specific
linker script to the `--default-script` linker script, and change the
main linker script (passed to `--script`, previously `-T`) to a generic
"override" linker script. This new generic linker script contains an
`INSERT BEFORE` command, which causes the linker to load it *in addition
to* the default linker script, while adding an extra `.noinit` output
section in the right place.

Because these new variables take up RAM but have their own section in
the .elf file, they should be accounted for in the size summary after
compilation. This is done by adapting the `recipe.size.regex.data` entry
to include this new section.

I originally implemented this for the reset-to-bootloader in #710, but I think this is useful by itself as well, for sketches to use, even when #710 might not end up using this. Also, since this is distinct feature, it helps unclutter #710 by merging this separately.

To test it, consider the following sketch:

```
unsigned boot_count __attribute__((__section__(".noinit")));

void setup() {
    Serial.begin(115200);
    while (!Serial); // Wait for serial port open

    // Initialize the variable only on first power-on reset
    if (__HAL_RCC_GET_FLAG(RCC_FLAG_PORRST))
        boot_count = 1;
    __HAL_RCC_CLEAR_RESET_FLAGS();
    
    Serial.print("Boot number: ");
    Serial.println(boot_count);
    ++boot_count;
}

void loop() { }
```

This shows the number of boots since the last POR by incrementing a noinit variable across resets. Note that when you first upload this, it might not start at 1 but at some arbitrary value, because typically the first boot after an upload is not a power-on-reset. To start at 1, disconnect and reconnect power.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -128,7 +128,7 @@ recipe.S.o.pattern="{compiler.path}{compiler.S.cmd}" {compiler.S.flags} {build.i
 recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{archive_file_path}" "{object_file}"
 
 ## Combine gc-sections, archives, and objects
-recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} "-T{build.variant.path}/{build.ldscript}" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} {compiler.ldflags} {compiler.arm.cmsis.ldflags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" -Wl,--start-group {object_files} {compiler.libraries.ldflags} "{archive_file_path}" -lc -Wl,--end-group -lm -lgcc -lstdc++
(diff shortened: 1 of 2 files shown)


fpistm:
@matthijskooijman 
Please, could you add this to the wiki?

matthijskooijman:
Thanks for the quick merge :-)

>  Please, could you add this to the wiki?

Of course, any suggestion on where would be appropriate? I was thinking maybe here: https://github.com/stm32duino/wiki/wiki/API and then make a new section "Other" (on the same level as "Built-in library") with "Uninitialized variables" under there?

fpistm:
Sound good for me.

matthijskooijman:
Documented here: https://github.com/stm32duino/wiki/wiki/API#Remembering-variables-across-resets

PR #997, opened 2020-03-19
Refactor USB pullup handling

### LINKED ISSUE #885: Generalize USB pullup attach/detach support
I previously commented about this in the jump-to-bootloader, but I think this is something that warrants a bit more discussion, and can be solved separately as well. See https://github.com/stm32duino/Arduino_Core_STM32/pull/710#discussion_r343027463 This is getting a bit of a long post, but I'm also trying to capture my thoughts and, while writing this, moving towards a proposal as well.

USB needs a pullup on the D+ line (or D- for low speed), which indicates the "attachment" status (i.e. pullup present signals the USB host that the device is present, while lack of a pullup detaches from the bus) and allows discriminating between low speed and fullspeed (or higher).

There are various ways this pullup can be implemented in the circuit. In particular, I've seen:
 - A fixed pullup
 - A pullup that can be switched on or off using a transistor, which can be default on or off.
 - A pullup builtin to some MCUs.

Control of this pullup happens in the `USBD_reenumerate()` function, which currently handles two cases:
 - When USB_DISC_PIN is defined by the variant (e.g. on the maple mini), the pullup is assumed to be disabled by default, and can be enabled by making the defined pin OUTPUT LOW.
 - When USB_DISC_PIN is not defined, it is assumed that there is an external fixed pullup resistor. To force re-enumeration, the DP pin itself is briefly put into output low mode, which overrides the pullup and looks like an USB detach to the host (though it is a bit of a hack and probably not entirely standards-compliant).

These two cases seem to be somehwat mismatched: When USB_DISC_PIN is defined, the function actually only attaches to the USB bus (which causes enumeration to happen), while otherwise the function actually detaches and reattaches (which causes reenumeration to happen). So maybe for the USB_DISC_PIN case, it should be driven output high (or input/floating) for a short while before switching to low, so it actually causes reenumeration if already enumerated (e.g. when called later after startup).

Note that the internal pullup and default-on pullup configurations are not currently handled.

More generally, you might want to have an API that actually just supports "Attach" and "Detach" operations, so any caller can decide what they need. However, this is problematic because:
 - In the fixed pullup configuration, detaching is really a hack that should probably not last for more than a few ms.
 - In the internal pullup configuration, the USB core automatically controls the pullup state (in particular, on the F401 we're using, the pullup is enabled automatically when USB device mode is enabled and VBUS is detected). It can be disabled using the SDIS (soft disconnect) bit, so some control is still available.
 - In the different configurations, the default / startup state is different, so different sequences of attach/detach might be required.

So, it is probably better to not support separate attach/detach, but instead have `USB_reenumerate()` (possibly under a different name) just handle all usecases that need to be handled. Looking at those, I can see:
 - Startup, which needs to trigger USB attachment. At first glance, it seems this only needs work for the external switched pullup case (just enable the pullup), but for the fixed pullup and default on configurations, the pullup stays on through a reset, so this actually needs to *disable* the pullup before enabling it. For the internal configuration, nothing is needed (the pullup will be enabled as part of USB initialization). Note that this is the only usecase present in the current code (i.e. the only call to `USBD_reenumerate()` is at startup).
 - Jumping to the bootloader, which needs to detach from USB and then reattach to allow the bootloader to provide its own descriptors. This case is added by #710. In the current implementation, the bootloader happens after a reset, which makes this case essentially identical to the normal startup case.
 - Re-enumerating during runtime. I can imagine one wants to, as part of their sketch, force re-enumeration (e.g. to change device type or configuration) without resetting the MCU. This is pretty similar to the startup case, except that it might need to explicitly disable an external pullup, since no MCU reset has done so.
 - Detaching from the USB bus for a longer period of time. One might want to disable USB connectivity based on some trigger, user interaction, etc. This is something that is not feasible with fixed pullups, but is possible in the other configurations.

Looking at the above, it seems that the first three cases are pretty identical (the third one needs some extra work in case no reset happened beforehand, but that extra work should not hurt if a reset happened, so maybe best to just always do that). The fourth usecase is different, but also not currently supported. Hence, `USBD_reenumerate()` could be made to support the first three cases, and then we could (if we need it later) add `USBD_detach()` or something to support the fourth case.

This means that the contract of `USBD_reenumerate()` can be:
 - When internal pullups are used, do nothing
 - Otherwise, when pullups are enabled, disable them  for a few ms
 - Then (re-)enable pullups
This function should then be called from the startup/USB code, it is not intended to be called from user code directly.

This supports the first three usecases (though it might need some extra work with internal pullups for the third usecase, reenumerating during runtime, when that is implemented).

Specifying the configuration
---
Currently, only two configurations are handled, based on the `USB_DISC_PIN` macro: External fixed pullup, and external switched pullup that is detached by default and attaches on OUTPUT LOW. I do not really like the `USB_DISC_PIN` macro name here (it suggests "disconnect pin", while the pin really *connects* rather than *disconnects*). Also, it supports only one specific external pullup configuration. I would propose generalizing this by using a few macros. In particular, here's what I would add to the `variant.h` template:

```
// If the board has external USB pullup (on DP/DP depending on speed)
// that can be controlled using a GPIO pin, define these.
//  - If the the pullup is disabled (detached) by default, define
//    USB_ATTACH_PIN to the pin that, when written to USB_ATTACH_LEVEL,
//    attaches the pullup.
//  - If the the pullup is enabled (attached) by default, define
//    USB_DETACH_PIN to the pin that, when written to USB_DETACH_LEVEL,
//    detaches the pullup.
//#define USB_ATTACH_PIN x
//#define USB_ATTACH_LEVEL LOW
//#define USB_DETACH_PIN x
//#define USB_DETACH_LEVEL LOW
```

I think this could also be used to simplify the MALYANM200 variant, which currently has a customized `USBD_reenumerate()` function because, I think, it has an active-high attach pin (Maybe @xC0000005 can confirm that the pullup is indeed disabled by default and enabled by making PB9 high?).

Then, we also need to handle the internal pullup case. Ideally, we would autodetect that internal pullups are supported. We could check for the `USB_OTG_DCTL_SDIS` macro, which is the Soft Disconnect bit - if you can soft disconnect, that probably means that pullups are builtin. Note sure if this catches all of them, though.

If there are no internal pullups and `USB_ATTACH_PIN`/`USB_DETACH_PIN` are not defined, we can fall back to the DP-pin-low hack. Alternatively, we could say that such a hack should not be a default and thus must be explicitly enabled for a board, but I guess that does not add so much (we can always add a macro to disable it later when needed).

Summarizing
---
It's become a bit of a wall of text, but unless I missed something above, I think this should be something we can implement now. In particular, this needs:
 - Changes to the template variant.h to document the new macros
 - Changes to `USBD_reenumerate()` to:
    - Support different external switchable pullup configurations
    - Disable the external switchable pullup and wait a bit, just like with the fixed pullup hack.
    - Detect internal pullups and do nothing.

matthijskooijman:
Previously, variants could define USB_DISC_PIN when they had an USB
attach pullup that was disabled by default and needed a pin to
be written LOW to enable it. Other hardware configurations could only
overwrite the USBD_reenumerate function, like the M200 board did.

This commit makes the pullup configuration more flexible. By defining
the appropriate macros, enabled-by-default and disabled-by-default
pullups are both supported. The output level of the pin can also be
configured. Finally, for disabled-by-default, the output mode for the
enabled state can also be configured (INPUT to leave the pin floating
and let any external transistor gate pullup enable the USB pullup, or
OUTPUT to keep the pin as OUTPUT an actively drive the transistor gate).

The new macros also take PinName constants rather than pin numbers. This
allows merging the code that controls an external pullup with the code
that writes directly to the USB DP pin to fake a disabled pullup (this
code only knows the pin name of the DP pin, not the number).

Finally, for CPUs that have internal USB pullup (as indicated by the
presence of a SDIS config bit), the write-to-DP-trick is now not
performed (since the pullup is automatically managed by the USB hardware
already).

This fixes #885, also see that issue for discussion leading up to this
change.

### DIFF
--- a/cores/arduino/stm32/usb/usbd_if.c
+++ b/cores/arduino/stm32/usb/usbd_if.c
@@ -10,35 +10,147 @@
 #include "usbd_if.h"
 #include "usbd_cdc_if.h"
 
+#if !defined(USBD_REENUM_DISABLED)
+
+/*
+ * Below, support for re-attaching to USB is handled. USB-attachment is
+ * detected by the presence of a pullup on one of the USB datalines.
+ *
+ * To force the USB host to re-enumerate, we must detach (disable
+ * pullup), wait a bit, and re-attach (enable pullup). This can be done
+ * while running, but must also happen at startup (including the detach
+ * and delay, since USB might have been active just before reset, so we
+ * must force a re-enumeration). On a cold boot (e.g. USB just plugged
+ * in), the forced detach might not be needed, but will not be harmful
+ * either.
+ *
+ * For pullup control, a number of cases are supported:
+ *  1. An external pullup controllable through a GPIO (e.g. using an
+ *     external transistor). The default output state (e.g. when the
+ *     GPIO is floating) can be enabled or disabled, and the output
+ *     polarity can be configured.
+ *  2. A pullup built into the chip's USB core. This is automatically
+ *     enabled by hardware or USB HAL code, and can be manually
+ *     controlled using HAL functions.
+ *  3. A fixed (always enabled) external pullup. To still allow detaching,
+ *     a bit of a hack is used: By setting the D+ to OUTPUT LOW, it
+ *     *looks* like the pullup is removed (the host has own, bigger,
+ *     pulldown, to detect detaching, so writing LOW looks the same).
+ *     This probably violates USB specifications, but does seem to work
+ *     well in practice. It probably also should only be used briefly,
+ *     keeping the pin OUTPUT LOW for longer might be problematic.
+ *
+ * These scenarios are considered in order: If the variant defines an
+ * external pullup control pin, it is used, else if an internal pullup
+ * is present, that is used, else the D+ trick is used.
+ *
+ * To configure #1, the variant should either define USBD_ATTACH_PIN and
+ * USBD_ATTACH_LEVEL when the pullup is disabled by default and the pin
+ * can be used to *attach*, or USBD_DETACH_PIN and USBD_DETACH_LEVEL
+ * when the pullup is enabled by default and the pin can be used to
+ * *detach*.
+ */
+
+/* Compatibility with the old way to specify this */
+#if defined(USB_DISC_PIN) && !defined(USBD_ATTACH_PIN)
+#define USBD_ATTACH_PIN USB_DISC_PIN
+#define USBD_ATTACH_LEVEL LOW
+#warning "USB_DISC_PIN is deprecated, use USBD_ATTACH_PIN instead"
+#endif /* defined(USB_DISC_PIN) && !defined(USBD_ATTACH_PIN) */
+
+/* Some sanity checks */
+#if defined(USBD_ATTACH_PIN) && defined(USBD_DETACH_PIN)
+#error "Cannot define both USBD_ATTACH_PIN and USBD_DETACH_PIN"
+#endif /* defined(USBD_ATTACH_PIN) && defined(USBD_DETACH_PIN) */
+#if defined(USBD_ATTACH_PIN) && !defined(USBD_ATTACH_LEVEL)
+#error "USBD_ATTACH_PIN also needs USBD_ATTACH_LEVEL defined"
+#endif /* defined(USBD_ATTACH_PIN) && !defined(USBD_ATTACH_LEVEL) */
+#if defined(USBD_DETACH_PIN) && !defined(USBD_DETACH_LEVEL)
+#error "USBD_DETACH_PIN also needs USBD_DETACH_LEVEL defined"
+#endif /* defined(USBD_DETACH_PIN) && !defined(USBD_DETACH_LEVEL) */
+
+/* Either of these bits indicate that there are internal pullups */
+#if defined(USB_BCDR_DPPU) || defined(USB_OTG_DCTL_SDIS)
+#define USBD_HAVE_INTERNAL_PULLUPS
+#endif /* defined(USB_BCDR_DPPU) || defined(USB_OTG_DCTL_SDIS) */
+
+/* Figure out which USB instance is used. This mirrors the decision made
+ * in USBD_LL_Init in usbd_conf.c. */
+#if defined(USE_USB_HS)
+#define USBD_USB_INSTANCE USB_OTG_HS
+#elif defined(USB_OTG_FS)
+#define USBD_USB_INSTANCE USB_OTG_FS
+#elif defined(USB)
+#define USBD_USB_INSTANCE USB
+#endif
+
+/*
+ * Translate pin number to a pin name (using the same define for both
+ * attach and detach pins) and also define the complementary level. This
+ * allows using the same code for the default-enabled and
+ * default-disabled case as well as for the D+ trick (which does not
+ * know the pin number, only the pin name).
+ */
+#if defined(USBD_ATTACH_PIN)
+#define USBD_PULLUP_CONTROL_PINNAME digitalPinToPinName(USBD_ATTACH_PIN)
+#define USBD_DETACH_LEVEL !(USBD_ATTACH_LEVEL)
+#elif defined(USBD_DETACH_PIN)
+#define USBD_PULLUP_CONTROL_PINNAME digitalPinToPinName(USBD_DETACH_PIN)
+#define USBD_ATTACH_LEVEL !(USBD_DETACH_LEVEL)
+#elif !defined(USBD_HAVE_INTERNAL_PULLUPS)
+/* When no USB attach and detach pins were defined, and there are also
+ * no internal pullups, assume there is a fixed external pullup and apply
+ * the D+ trick. This should happen only for the USB peripheral, since
+ * USB_OTG_HS and USB_OTG_FS always have internal pullups. */
+#if !defined(USB)
+#error "Unexpected USB configuration"
+#endif
+#define USBD_PULLUP_CONTROL_PINNAME USB_DP
+#define USBD_DETACH_LEVEL LOW
+// USBD_ATTACH_LEVEL not needed.
+#define USBD_DP_TRICK
+#endif
+
 /**
-  * @brief  Force to re-enumerate USB
+  * @brief  Force to re-enumerate USB.
+  *
+  * This is intended to be called at startup by core code. It could be
+  * used at runtime, while USB is connected, to force re-enumeration
+  * too, but that does not work in all cases (when USB is enabled on an
+  * F103C8, setting output mode on the DP pin no longer has any effect).
+  *
   * @param  None
   * @retval None
   */
-
 WEAK void USBD_reenumerate(void)
 {
-#ifndef USBD_REENUM_DISABLED
-  /* Re-enumerate the USB */
-#ifdef USB_DISC_PIN
-  pinMode(USB_DISC_PIN, OUTPUT);
-  digitalWrite(USB_DISC_PIN, LOW);
-#else
-#ifdef USE_USB_HS_IN_FS
-  PinName pinDP = USB_OTG_HS_DP;
-#elif defined(USB_OTG_FS)
-  PinName pinDP = USB_OTG_FS_DP;
-#else /* USB */
-  PinName pinDP =  USB_DP;
-#endif
-  pin_function(pinDP, STM_PIN_DATA(STM_MODE_OUTPUT_PP, GPIO_NOPULL, 0));
-  digitalWriteFast(pinDP, LOW);
+#if defined(USBD_PULLUP_CONTROL_PINNAME)
+  /* Detach */
+  pin_function(USBD_PULLUP_CONTROL_PINNAME, STM_PIN_DATA(STM_MODE_OUTPUT_PP, GPIO_NOPULL, 0));
+  digitalWriteFast(USBD_PULLUP_CONTROL_PINNAME, USBD_DETACH_LEVEL);
+
+  /* Wait */
   delay(USBD_ENUM_DELAY);
-  pin_function(pinDP, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
-  /*delay(USBD_ENUM_DELAY);*/
-#endif /* USB_DISC_PIN */
-#endif /* USBD_REENUM_DISABLED */
+
+  /* Attach */
+#if defined(USBD_DP_TRICK)
+  /* Revert back to input (floating), needed for the D+ trick */
+  pin_function(USBD_PULLUP_CONTROL_PINNAME, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
+#else
+  digitalWriteFast(USBD_PULLUP_CONTROL_PINNAME, USBD_ATTACH_LEVEL);
+#endif /* defined(USBD_PULLUP_CONTROL_FLOATING) */
+#elif defined(USBD_HAVE_INTERNAL_PULLUPS)
+  USB_DevDisconnect(USBD_USB_INSTANCE);
+  delay(USBD_ENUM_DELAY);
+  USB_DevConnect(USBD_USB_INSTANCE);
+#else
+#warning "No USB attach/detach method, USB might not be reliable through system resets"
+#endif
 }
+
+#else /* !defined(USBD_REENUM_DISABLED) */
+WEAK void USBD_reenumerate(void) { }
+#endif
 
 #ifdef USBD_USE_CDC
 void USBD_CDC_init(void)
--- a/variants/MAPLEMINI_F103CB/variant.h
+++ b/variants/MAPLEMINI_F103CB/variant.h
@@ -108,8 +108,9 @@ extern "C" {
 #define PIN_SERIAL_RX           PA10
 #define PIN_SERIAL_TX           PA9
 
-// USB
-#define USB_DISC_PIN            PB9
+// USB, pull this pin low to enable the USB attach pullup
+#define USBD_ATTACH_PIN         PB9
+#define USBD_ATTACH_LEVEL       LOW
 
 #ifdef __cplusplus
 } // extern "C"
--- a/variants/board_template/variant.h
+++ b/variants/board_template/variant.h
@@ -162,6 +162,19 @@ extern "C" {
 // See AN4879 https://www.st.com/content/st_com/en/search.html#q=AN4879-t=resources-page=1
 //#define USBD_VBUS_DETECTION_ENABLE
 
+// If the board has external USB pullup (on DP/DM depending on speed)
+// that can be controlled using a GPIO pin, define these:
+//  - If the the pullup is disabled (USB detached) by default, define
+//    USBD_ATTACH_PIN to the pin that, when written to
+//    USBD_ATTACH_LEVEL, attaches the pullup.
+//  - If the the pullup is enabled (attached) by default, define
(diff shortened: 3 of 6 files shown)


matthijskooijman:
I've only done limited testing on this PR so far. I only have access to an F4 board with native USB, which has internal pullups and works well. I've ordered a blue pill and black pill to test, but it would be good to also test on a maple mini board (since that has external pullup) and the MALYAN M200 V1 (since that used to have special pullup handling). @xC0000005 maybe you could do the latter?

(I also did a few force pushes to fix style errors, should be good now).

xC0000005:
I'l test on V1 and V2 (pullup and no pullup) tonight.

fpistm:
And I will test on Maple. :wink:

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -10,33 +10,77 @@
...
-  digitalWriteFast(pinDP, LOW);
+#ifdef USBD_DETACH_PINNAME
+  // Default is attached, so detach, wait, and reattach.
+  pin_function(USBD_DETACH_PINNAME, STM_PIN_DATA(STM_MODE_OUTPUT_PP, GPIO_NOPULL, 0));
+  digitalWriteFast(USBD_DETACH_PINNAME, USBD_DETACH_LEVEL);
   delay(USBD_ENUM_DELAY);
-  pin_function(pinDP, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
-  /*delay(USBD_ENUM_DELAY);*/
-#endif /* USB_DISC_PIN */
+  if (USBD_ATTACH_MODE_INPUT == OUTPUT) {
```
xC0000005: I'm getting a compliation error here when compiling for the M200. Is this supposed to be USBD_ATTACH_MODE?

matthijskooijman:
Oh, remove the _INPUT, that should come from here: https://github.com/stm32duino/Arduino_Core_STM32/pull/997/files#diff-d707080ac76ca2646a4a2457bcb86ad9R44

variants/MAPLEMINI_F103CB/variant.h
```diff
@@ -108,8 +108,9 @@ extern "C" {
 #define PIN_SERIAL_RX           PA10
 #define PIN_SERIAL_TX           PA9
 
-// USB
-#define USB_DISC_PIN            PB9
+// USB, pull this pin low to enable the USB attach pullup
+#define USBD_ATTACH_PINNAME     PB_9
+#define USBD_ATTACH_LEVEL       LOW
```
xC0000005: I'm not entirely sure how this works, but shouldn't one of these be low and the other high?

matthijskooijman: What do you mean? The PINNAME indicates the pin to control the pullup, and the LEVEL indicates the level it needs to have to ATTACH. What should be HIGH here? Also see the comments in variants/board_template/variant.h.

matthijskooijman:
Sorry for the short comment earlier, I was just on my way out.

Turns out with my earlier compile-testing, I had forgotten to enable USB (for our custom board we have it enabled by default, didn't realize that the default boards did not). I did that now and found a dozen other preprocessor typos, I messed up the _PIN to _PINNAME rename and some other stuff. I should really have tested this more thoroughly, sorry.

Anyway, I pushed a fixup commit that at least compiles properly for the bluepill, maple mini and M200, so that should make testing easier.

Thanks!

xC0000005:
I pulled the most current code and USB is working for both M200V1 and V2 (103 and 070).

matthijskooijman:
I've done another push, to squash the fixup commit I pushed yesterday (only changes history, not the actual code).

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -10,33 +10,77 @@
...
+  if (USBD_ATTACH_MODE == OUTPUT) {
+    digitalWriteFast(USBD_DETACH_PINNAME, !USBD_DETACH_LEVEL);
+  } else {
+    pin_function(USBD_DETACH_PINNAME, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
+  }
+#elif defined(USBD_ATTACH_PINNAME)
+  // Default is detached, so just attach
+  pin_function(USBD_ATTACH_PINNAME, STM_PIN_DATA(STM_MODE_OUTPUT_PP, GPIO_NOPULL, 0));
+  digitalWriteFast(USBD_ATTACH_PINNAME, USBD_ATTACH_LEVEL);
+#elif !defined(USB_OTG_DCTL_SDIS)
```
ABOSTM: I didn't catch why we do not set then reset bit USB_OTG_DCTL_SDIS in case of Internal PullUp ? this should force reenumeration, isn't it ? otherwise, how reenumeration is performed ?

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -10,33 +10,77 @@
...
-  pin_function(pinDP, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
-  /*delay(USBD_ENUM_DELAY);*/
-#endif /* USB_DISC_PIN */
+  if (USBD_ATTACH_MODE == OUTPUT) {
+    digitalWriteFast(USBD_DETACH_PINNAME, !USBD_DETACH_LEVEL);
+  } else {
+    pin_function(USBD_DETACH_PINNAME, STM_PIN_DATA(STM_MODE_INPUT, GPIO_NOPULL, 0));
+  }
+#elif defined(USBD_ATTACH_PINNAME)
+  // Default is detached, so just attach
```
ABOSTM: Reading your comment,

> Looking at the above, it seems that the first three cases are pretty identical (**the third one needs some extra work in case no reset happened beforehand, but that extra work should not hurt if a reset happened, so maybe best to just always do that**).

I was expecting to see similar processing for both USBD_ATTACH_PINNAME and USBD_DETACH_PINNAME : 
First dettach then attach, thus you would cover the third case.

variants/board_template/variant.h
```diff
@@ -162,6 +162,21 @@ extern "C" {
...
+//    USBD_ATTACH_PINNAME to the pin that, when written to
+//    USBD_ATTACH_LEVEL, attaches the pullup.
+//  - If the the pullup is enabled (attached) by default, define
+//    USBD_DETACH_PINNAME to the pin that, when written to
+//    USBD_DETACH_LEVEL, detaches the pullup.
+// Note that these constants need PinName values (e.g. PB_5) not pin
+// numbers (e.g. PB5).
+//#define USBD_ATTACH_PINNAME x
+//#define USBD_ATTACH_LEVEL LOW
+//#define USBD_DETACH_PINNAME NAMEx
```
ABOSTM: dito

variants/board_template/variant.h
```diff
@@ -162,6 +162,21 @@ extern "C" {
...
+// that can be controlled using a GPIO pin, define these:
+//  - If the the pullup is disabled (USB detached) by default, define
+//    USBD_ATTACH_PINNAME to the pin that, when written to
+//    USBD_ATTACH_LEVEL, attaches the pullup.
+//  - If the the pullup is enabled (attached) by default, define
+//    USBD_DETACH_PINNAME to the pin that, when written to
+//    USBD_DETACH_LEVEL, detaches the pullup.
+// Note that these constants need PinName values (e.g. PB_5) not pin
+// numbers (e.g. PB5).
+//#define USBD_ATTACH_PINNAME x
```
ABOSTM: To reinforce your note, x could be advantageously replaced by PY_x format used in wiki for example
```suggestion
//#define USBD_ATTACH_PINNAME PY_x
```

fpistm:
Tested with a Maple mini and works as expected.

cores/arduino/stm32/usb/usbd_if.c
matthijskooijman: The pullup is automatically controlled by hardare, so it is already inactive during reset and activated once the USB core is initialized, so nothing is needed here during startup (SDIS won't actually *do* anything when the USB core is not initialized).

One case where SDIS could be useful is to re-attach during normal operation. This is the third usecase I described in #885, and though I planned to support (I see now), this PR does not implement it yet. Might be good to add anyway, which would involve triggering toggling SDIS, and also disabling the ATTACH pin (which is now only enabled, never disabled).

I'm not sure how the USB core handles a random detach / attach, but I guess it won't be much different from actually replugging USB while externally powered, which the code probably handles? I'll test :-)

cores/arduino/stm32/usb/usbd_if.c
matthijskooijman: Yeah, let's do that indeed.

matthijskooijman:
Thanks for the testing and reviews. I've also tested on a bluepill, which works as expected too.

> There is a remaining definition of USB_DISC_PIN in variant: Generic_F401Rx

Ah, I started working on top of an older branch where that variant did not exist yet. I'll convert it.

> The new macros also take PinName constants rather than pin numbers. This
allows merging the code that controls an external pullup with the code
that writes directly to the USB DP pin to fake a disabled pullup (this
code only knows the pin name of the DP pin, not the number).

I thought a bit more on this, but I think it might be nicer to specify pin numbers rather than pin names in the variant, which is more consistent and thus also less likely to cause mistakes. In terms of code that just means the PIN constants must be redefined as PINNAME constants in usbd_if.c, which should just be a few more lines but nothing spectacular (and just one extra lookup at runtime, but that's ok).

matthijskooijman:
Ok, I pretty much rewrote the entire thing. In a nutshelll:
 - All three hardware configurations (builtin pullups, external fixed pullups and external controllable pullups) are now handled in the same way: Detach, wait, attach.
 - A variant-specified control pin is now always push-pull, the "open-collector" treatment is applied only for the D+ output trick.
 - Variants can now just configure pin numbers instead of pin names.

I tested this on an F3 with internal pullups and on a bluepill. Testing on the MALYAN printers and a maple mini again would be appreciated.

For some more details about what I did and tried, read on.

> Nitpicking: those sanity check should be under switch : #ifndef USBD_REENUM_DISABLED

I now moved all of the code under this check, with just an empty `USBD_reenumerate()` function in the else.

### Internal pullups
I found that the internal pullups can be controlled portably using the `USB_DevConnect` and `USB_DevDisconnect` LL HAL functions, which the code now does.

Looking through the HAL code I found that the SDIS bit is used only by "USB_OTG_FS" and "USB_OTG_HS" peripherals, apparently some "USB" peripherals have an internal pullup controlled by the "USB_BCDR_DPPU" bit, so I adapted the code to check for both. I used `git grep -A 10 USB_DevConnect.*\)$` to check that no other bits are used for pullup control.

Even more, I found that *all* USB_OTG_HS and USB_OTG_FS peripherals have the SDIS bit, so there is really no need to have code to figure out the D+ trick for them (so I removed that code). I confirmed this observation using a [check script I wrote today](https://github.com/matthijskooijman/scripts/blob/master/Arduino/stm32-check-all) which compiles a short header file against all known STM32 cpus (see the script for info). I let the script compile this, which worked for all boards:

```
#include <stm32_def.h>

#if (defined(USB_OTG_FS) || defined(USB_OTG_HS)) && !defined(USB_OTG_DCTL_SDIS)
#error "Foo"
#endif
```

I have not been able to test the DPPU bit, in particular I'm not sure how it behaves on startup when the USB peripheral is not enabled yet (I think it will just be a no-op, but it might end up enabling the USB pullup earlier than normal, though that should not be a problem either. I see that the STM32F070x6/B has this bit, which includes the MALYAN M200 V2, so that provides a nice opportunity to test this (also, it explains why these V2 boards have no external pullup, since it is included in the CPU now).

I also considered the option of overriding the `USB_DevConnect` and `USB_DevDisconnect` functions with our custom code to control the external pullup (or apply the D+ trick), since they are *always* called from the PCD code (even on chips that do not have internal pullups). However, the dummy definitions of these functions on chips without internal pullups are not defined weak, so we cannot actually override them.

### Re-enumerating while connected
The code as it is now, should be usable to re-enumerate while running (instead of just at reset) too. I don't think we should actively advertise this, but it is interesting to know if it would work.

This works well with the F301RE with SDIS support I have, so it seems the USB stack handles external re-attachment fine. There did seem to be a small issue, possibly with buffering while disconnected (I got some duplicate serial output when the port was reconnected), but I did not investigate this.

I also tried on a bluepill, which has the D+ output trick. Switching to OUTPUT and back to INPUT would of course not work, since then the USB peripheral would lose control of the pin. So I tried using the USB pinmap to switch the pin back to USB AF mode.

However, detaching by switching to OUTPUT did not even work on F103C8, since it seems that the USB peripheral (when enabled) forcibly claims the DP pin, regardless of the AF settings. Even when I switch it to output and leave it at that, USB serial keeps working normally. Even more, the [pin map in the variant](https://github.com/stm32duino/Arduino_Core_STM32/blob/eeabfc10fc80b52c19ba6ce4d018c5d0a6e3c22e/variants/PILL_F103XX/PeripheralPins.c#L268-L269) does not even bother to set the pins into AF mode and just leaves them in input (verified using  debugger, so there is no bit of code elsewhere that *does* initialize to AF). I could not find any reference to this in the documentation, and [just one mention of it online](https://community.st.com/s/question/0D50X00009XkZP7SAN/stm32l151-usb-what-actually-switches-pa11pa12-from-gpio-to-usb), but that thread quickly focuses on getting things to work on hardware that does *not* seem to misbehave in this way. Since I could not test this properly, and the code was a bit fragile anyway (it would put pins in USB mode during startup, way before the USB peripheral was initialized), I reverted this, the pin mode is now reverted back to INPUT and intended to work at reset time only.

I have no board with DPPU support to test with, though. In case anyone wants to test, here is the sketch I used:

```
void setup() {
  Serial.begin(115200);
  while (!Serial);
}

unsigned i = 0;

extern "C" void USBD_reenumerate(void);

void loop() {
  Serial.println(i++);
  delay(1000);
  if (i % 10 == 0) {
      USBD_reenumerate();
  }
}
```

ABOSTM:
I tested your sketch with PNUCLEO_WB55RG which supports DPPU , it is working !!

fpistm:
> Testing on the MALYAN printers and a maple mini again would be appreciated.

@matthijskooijman I've tested the maple and it works as expected with your sketch.

@xC0000005 did you made the test on Mylyan ?

xC0000005:
I’ve just tested M200 V1, V2, and M300. USB works on all of them with this version.

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -10,36 +10,148 @@
...
+ *     external transistor). The default output state (e.g. when the
+ *     GPIO is floating) can be enabled or disabled, and the output
+ *     polarity can be configured.
+ *  2. A pullup built into the chip's USB core. This is automatically
+ *     enabled by hardware or USB HAL code, and can be manually
+ *     controlled using HAL functions.
+ *  3. A fixed (always enabled) external pullup. To still allow detaching,
+ *     a bit of a hack is used: By setting the D+ to OUTPUT LOW, it
+ *     *looks* like the pullup is removed (the host has own, bigger,
+ *     pulldown, to detact detaching, so writing LOW looks the same).
```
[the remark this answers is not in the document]
matthijskooijman: Fixed

matthijskooijman:
W00ps, forgot about the typo. Force pushed a version just fixing the typo, should be good to merge now.

PR #586, opened 2019-07-30
Add composite MSC CDC USB device

Bob-the-Kuhn:
**This PR should be considered experimental until more testing is done.**

I'm definitely not an expert in USB or the STM32 devices.  I'm sure there's lots of corrections/improvements that will be made before this PR is merged.

---

**Summary**

This PR implements a composite MSC CDC USB device.

The SD card on the STM32 device will appear as a USB drive on the host PC.  This allows file transfers between the PC and the STM32 device.  Hopefully this minimizes the need to physically move the SD card between the PC and the STM32 device when transferring files.

**Implementation**

The flag **USE_USB_COMPOSITE** is used to enable the new feature.

SDIO is used to access the SD card.

Almost all the changes are in the files in the new directory **msc_cdc_composite**.  There are some shared files that were also changed.

The biggest change is to modify the USBD_HandleTypeDef  structure so that the MSC and CDC classes have dedicated/unique data structures when USE_USB_COMPOSITE is defined.

The USBD_HandleTypeDef change is copied from the [GPSlogger repo](https://github.com/grafalex82/GPSLogger).  See the article [CDC + MSC USB Composite Device for STM32 HAL](https://habr.com/en/post/335018) for the theory behind the MSC CDC composite device as implemented in the GPSLogger repository.  Here is an [English version](https://weekly-geekly.github.io/articles/335018/index.html) of the article.

The most of the customization & new code are in the files:
- usbd_desc_msc_cdc_composite.c
- usbd_msc_cdc_composite_def.h
- usbd_msc_cdc_composite_if.c
- SdMscDriver
- usbd_def.h
- usbd_conf.c

---

**CPU support**

So far all the testing has been done on a BLACK_STM32F407ve board using Marlin software and PlatformIO

There are some hooks in it for use with the STM32F103Z cpu.  I have not been able to compile Marlin with this repo (Arduino_Core_STM32) for a F103 device as the Marlin F103 software uses STM32GENERIC.

There are also hooks in it for use with High Speed USB.  No testing has been done on that.  All testing so far has been done with the F407VE in Full Speed mode.

---

**Marlin sketch**

I used [Marlin 2.0](https://github.com/MarlinFirmware/Marlin/tree/bugfix-2.0.x) for testing.  The changes needed are:
file **platformio.ini** - under the **[env:black_stm32f407ve]** section add `-DUSBD_USE_CDC_COMPOSITE` to the **build_flags** list.
[...]

PR #1004, opened 2020-03-20
Add composite MSC CDC USB device (follow up to PR #586)

Bob-the-Kuhn:
This is a follow up to PR #586.  The design is the same but now uses the latest middle wares.

Please see PR #586 for details.

---

**This should be considered experimental until others have had a chance to play with it.**

---

This code was tested with two boards: Black STM32F407 and STEVAL

The Marlin sketch this was tested with is at [PR 17222](https://github.com/MarlinFirmware/Marlin/pull/17222).

---

OPEN ITEMS:
- General cleanup - I'm sure some of my comments to myself and test code weren't all purged.
- Initialization sequence needs to be reviewed in detail.  I may still have some redundant items in it.  The switching to 4 bit wide mode routine may be redundant.
- Look at a way to stay in 1 bit wide mode.  I have a vague recollection of one board not connecting all the SDIO data lines.
- The FIFO configuration needs more investigation.  Currently I can only set the PCD FIFOs to the minimum needed length.  Previously I was able to set FIFOs to use almost all the available space.
- My 4G SD card doesn't work with this software. I need to root cause that.

### DIFF
--- a/cores/arduino/stm32/usb/cdc/usbd_cdc.c
+++ b/cores/arduino/stm32/usb/cdc/usbd_cdc.c
@@ -52,6 +52,7 @@
 
 #ifdef USBCON
 #ifdef USBD_USE_CDC
+#ifndef USBD_USE_CDC_COMPOSITE
 
 /* Includes ------------------------------------------------------------------*/
 #include "usbd_cdc.h"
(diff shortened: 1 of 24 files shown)


fpistm:
Hi @Bob-the-Kuhn 
why not simply update the previous PR ?

I'm a little bit lost.

> now uses the latest middle wares

What did you mean by this?

Bob-the-Kuhn:
> why not simply update the previous PR ?

I was able to do this.   I'm going to close this PR.

> now uses the latest middle wares

Back in mid 2019 there was mention that the middle wares were going to be updated so I put this work aside.  I had some free time earlier this year so I started this work back up assuming that the updates had been done.

Issue #1006, opened 2020-03-22
Allow indenting preprocessor macros?


matthijskooijman:
While developing #997, I wrote some complicated and nested preprocessor statements. I originally tried making them clearer by applying indents, but that was not allowed by the astyle checks, so I removed them again. Now I wonder, should this astyle rule maybe be changed to allow (even if not require) indenting preprocessor directives?

e.g. consider:

```C
#if defined(USBD_ATTACH_PIN)
#define USBD_PULLUP_CONTROL_PINNAME digitalPinToPinName(USBD_ATTACH_PIN)
#define USBD_DETACH_LEVEL !(USBD_ATTACH_LEVEL)
#elif defined(USBD_DETACH_PIN)
#define USBD_DETACH_PINNAME digitalPinToPinName(USBD_DETACH_PIN)
#define USBD_ATTACH_LEVEL !(USBD_DETACH_LEVEL)
#elif !USBD_HAVE_INTERNAL_PULLUPS
/* When no USB attach and detach pins were defined, and the USB module
 * also does not have any builtin pullups, assume there is a fixed
 * external pullup and apply the D+ trick. */
#if defined(USE_USB_HS_IN_FS)
#define USBD_PULLUP_CONTROL_PINNAME USB_OTG_HS_DP
#elif defined(USB_OTG_FS)
#define USBD_PULLUP_CONTROL_PINNAME USB_OTG_FS_DP
#else /* USB */
#define USBD_PULLUP_CONTROL_PINNAME  USB_DP
#endif
/* Detach by writing LOW, but leave floating instead of writing
 * HIGH to attach. */
#define USBD_DETACH_LEVEL LOW
#define USBD_PULLUP_CONTROL_FLOATING
#endif /* !defined(USBD_ATTACH_PINNAME) && !USBD_HAVE_INTERNAL_PULLUPS */
```

vs:

```C
#if defined(USBD_ATTACH_PIN)
    #define USBD_PULLUP_CONTROL_PINNAME digitalPinToPinName(USBD_ATTACH_PIN)
    #define USBD_DETACH_LEVEL !(USBD_ATTACH_LEVEL)
#elif defined(USBD_DETACH_PIN)
    #define USBD_DETACH_PINNAME digitalPinToPinName(USBD_DETACH_PIN)
    #define USBD_ATTACH_LEVEL !(USBD_DETACH_LEVEL)
#elif !USBD_HAVE_INTERNAL_PULLUPS
[...]

PR #1007, opened 2020-03-22
Astyle indent preproc block

matthijskooijman:
Based on #1006, I quickly tried enabling the astyle ident-preproc-block to see what the impact would be. It produces a rather big diff, but on first glance does improve readability a lot.

Because it changes so much, this would probably mess with open pullrequests, so it might be good to merge those as far as possible (it is probably also possible to update them semi-automatically, but might still be a bit messy).


fpistm:
To have it passed, could you rebase on top of the master, please?
Thanks

fpistm:
Looking at all changes, I will have to update my scripts to generate wrapper files and PinNamesVar.h to indent properly by default.

matthijskooijman:
I pushed one more commit to let astyle enforce linux-style line endings. I noticed some CRLF-style files in the jumpsysBL branch, but when adding this option, notice @fpistm had already unified these in master. This just encodes this in the astyle config as well.

CI/astyle/.astylerc
```diff
@@ -13,6 +13,7 @@ style=kr
 # 2 spaces, convert tabs to spaces
 indent=spaces=2
 convert-tabs
+lineend=linux
```
fpistm: I will not add this.
For those working on Windows, files in the working tree will be in CRLF so the astyle will failed.

matthijskooijman: Ah, I see that the .gitattributes file indeed enables EOL translation/normalization for all text files. We could explicitly specify `eol=lf` as well, to checkout all files with lf as well (in practice, I think most tools on Windows do not really care about CRLF vs LF anymore, except maybe notepad). OTOH, with the current configuration, files should already be normalized to newlines by git on checkin, so this should usually be ok automatically.

One alternative would be to change the python script to apply lineend=linux only on non-Windows platforms and/or only in the CI-run, but that is probably too complicated and unexpected, so better leave that out.

I've removed the lineend commit again.

fpistm: I think about fforce LF on windows this would avoid any issue. We had the case in a PR, I guess user git config force to chieckin as CRLF

matthijskooijman:
What shall we do with this PR? I've just rebased it on top of master again, maybe it could just be merged? There's still a few open PRs that would produce conflicts from this, but most of them are experimental and would need to be mostly redone anyway, I think?

CI/astyle/.astylerc
matthijskooijman: Ah, in case we want that, I think it's a matter of adding `eol=lf` to all text lines in `.gitattributes`. But this can probably also be considered and tested separately of this PR, of course.

fpistm: Right. Done in #1021

PR #1014, opened 2020-03-26
Add support of NUCLEO_F746ZG and NUCLEO_F756ZG

fpistm:
The only difference between both is that F756 embed cryptographic accelerator AES and CRYP,
that are not currently supported by Arduino

Common PR from #963 and #999


hockeymikey:
@fpistm Thanks for the help!

PR #1019, opened 2020-03-29
F4x1 re update

fpistm:
Unified Nucleo F401RE and F411RE and use HSE BYPASS instead of HSI by default.
F401 runs at 84MHz and  F411 at 100MHz.


variants/NUCLEO_F4x1RE/variant.cpp
```diff
@@ -163,16 +153,15 @@ WEAK void SystemClock_Config(void)
   RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV2;
   RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV1;
 
+#ifdef ARDUINO_NUCLEO_F401RE
+  if (HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_2) != HAL_OK) {
+#else /* ARDUINO_NUCLEO_F411RE */
   if (HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_3) != HAL_OK) {
-    _Error_Handler(__FILE__, __LINE__);
```
ABOSTM: LSE is lost in this merge ?

variants/NUCLEO_F4x1RE/PeripheralPins.c
```diff
@@ -229,12 +235,18 @@ WEAK const PinMap PinMap_SPI_SCLK[] = {
 WEAK const PinMap PinMap_SPI_SSEL[] = {
   {PA_4,  SPI1, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF5_SPI1)},
   //  {PA_4,  SPI3, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF6_SPI3)},
-  {PA_15, SPI1, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF5_SPI1)},
-  //  {PA_15, SPI3, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF6_SPI3)},
+  //  {PA_15, SPI1, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF5_SPI1)},
+  {PA_15, SPI3, STM_PIN_DATA(STM_MODE_AF_PP, GPIO_PULLUP, GPIO_AF6_SPI3)},
```
ABOSTM: on F411 PA_15 was on SPI1, this may introduce regressions on users sketches

variants/NUCLEO_F4x1RE/variant.cpp
```diff
@@ -111,50 +111,40 @@ extern "C" {
...
-  *            APB2 Prescaler                 = 1
-  *            HSI Frequency(Hz)              = 16000000
-  *            PLL_M                          = 8
-  *            PLL_N                          = 100
-  *            PLL_P                          = 4
-  *            PLL_Q                          = 2
-  *            VDD(V)                         = 3.3
-  *            Main regulator output voltage  = Scale1 mode
-  *            Flash Latency(WS)              = 3
   * @param  None
```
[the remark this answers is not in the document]
fpistm: Right but those information are in the function itself and are (I think explicit). CubeMX does not generate those information.

variants/NUCLEO_F4x1RE/variant.cpp
fpistm: Yes, this is useless as the RTC will enable the LSE if requested.
https://github.com/stm32duino/Arduino_Core_STM32/blob/1b8664a815b8968bfb23a968b84e5b2fdc422ce4/libraries/SrcWrapper/src/stm32/rtc.c#L114-L122

variants/NUCLEO_F4x1RE/PeripheralPins.c
fpistm: I know, but it make sense to provide also an HW SSEL for SPI3.

variants/NUCLEO_F4x1RE/variant.cpp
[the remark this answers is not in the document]
ABOSTM: You don't have the info like SYSCL = 100000000Hz
It is not explicit in the function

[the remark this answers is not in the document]
fpistm: Done. Old comment was wrong for F411RE as it missed one 0 😉

PR #1032, opened 2020-04-16
Add more Generic_F103Rx targets

MCUdude:
This adds F103R4, F103R6, F103RD, F103RF and F103RG targets.
PeripheralPins.c is now based on STM32F103R(F-G)Tx, but is carefully modified to match all targets in this family.


MCUdude:
FYI I changed the analog pin part of the pinout. Now all pins use ADC1 by default, just like all generic F4 targets. This leaves ADC2 (and ADC3 on STM32F103xE and STM32F103xG) for "custom user" in the user application.

MCUdude:
BTW is there any good reason why the F103RC and F103ZC are listed with 49152 bytes instead of 65536 bytes of RAM?

fpistm:
Thanks @MCUdude for the PR

About:

> FYI I changed the analog pin part of the pinout. Now all pins use ADC1 by default, just like all generic F4 targets. This leaves ADC2 (and ADC3 on STM32F103xE and STM32F103xG) for "custom user" in the user application.

It is not really relevant, as said it is randomly set. As it can be customized like all PinMap array are weaked.

 

> ~BTW is there any good reason why the F103RC and F103ZC are listed with 49152 bytes instead of 65536 bytes of RAM?~
> 
> Nevermind, the datasheet says 48 kB, but the [spec-sheet over at the ST website says 64 kB.](https://www.st.com/en/microcontrollers-microprocessors/stm32f103.html)

About this,
the datasheet is correct


And yes the  "spec sheet' seems wrong as it should be 256 k / 48 k instead of  256 k / 64 k for the line.

MCUdude:
Let me know if it looks alright and can be merged. I also have a Generic F103Cx pinout ready, but due to me not being a git wizard, I will create a new PR when this one is merged. If you want, I can also create a generic F103Vx pinout so that the 1.9.0 release will contain pretty much all F103 chips available.

BTW when do you think 1.9.0 will be released?

fpistm:
I do my best to release the 1.9.0 soon.
It is a huge release and still some minor defects to fix.

fpistm:
@MCUdude 
FYI, the ST  STM32F103 product line image have been updated. Now it is well 48K 😉 
Thanks again for pointing this.

PR #1040, opened 2020-04-20
Fix DAC for F103RF and RG

MCUdude:
These lines prevented the DAC implementation from working on F103RF and F103RG.
Note that the definition is located in PeripheralPins.c.

### DIFF
--- a/variants/Generic_F103Rx/variant.h
+++ b/variants/Generic_F103Rx/variant.h
@@ -119,11 +119,6 @@ extern "C" {
 // Define here Serial instance number to map on Serial generic name
 #define SERIAL_UART_INSTANCE    1
 
-/* Extra HAL modules */
-#ifdef STM32F103xE
-#define HAL_DAC_MODULE_ENABLED
-#endif
-
 // Default pin used for 'Serial1' instance
 #define PIN_SERIAL_RX           PA10
 #define PIN_SERIAL_TX           PA9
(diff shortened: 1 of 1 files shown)


fpistm:
The define have to be in the variant

MCUdude:
There's an ifdef here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/6eaa8d20d69782ac346f6bc7d094cf078aba6d2c/variants/Generic_F103Rx/PeripheralPins.c#L89-L97

But this means we'll have to manually have to enable `HAL_ADC_MODULE_ENABLED ` and `HAL_DAC_MODULE_ENABLED`? if so, where is HAL_ADC_MODULE_ENABLED defined?

fpistm:
ADC is defined by default for all as all have an ADC.
DAC is not defined by default so it have to be defined for those which have DAC.
In the peripheral pins, this is not the definition, so if you remove it from variant DAC will not be usable unless user define it thanks build_opt.h or hal_conf_extra.h.

variants/Generic_F103Rx/variant.h
```diff
@@ -119,11 +119,6 @@ extern "C" {
 // Define here Serial instance number to map on Serial generic name
 #define SERIAL_UART_INSTANCE    1
 
-/* Extra HAL modules */
-#ifdef STM32F103xE
-#define HAL_DAC_MODULE_ENABLED
-#endif
-
```
fpistm, line 126: This would be this

```c
#if defined(STM32F103xE) || defined(STM32F103xG)
#define HAL_DAC_MODULE_ENABLED
#endif
```

PR #1043, opened 2020-04-23
Add support for F103Tx chips

MCUdude:
These are the only F103s that's currently not supported by this core.
This PR adds support for F103T4-6-8-B.


MCUdude:
> As mentioned in other PR for Generic variant, I will unify system core clock config

Yes, I'm aware of it. I figured I'd just use the current implementation since you're going to re-do them anyways.

fpistm:
No worries, this is mainly a note for me 😉

PR #1044, opened 2020-04-24
Improve supported boards list

ThomasGravekamp:
**Summary**

Add links to new STM32-base board pages.
Add links to manufacturers websites under the "More info" links.
Add device(s) column.

I took the liberty to do a bit of reformatting of the supported boards list. I've added the device(s) column so that you can search for specific devices.

There's one open question though. There was this line:

```
| :green_heart: | [Black F407VET6](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0) | *1.4.0* | VG/ZE/ZG in *1.5.0* |
```

The ZE/ZG variant of this board must be a completely different board, as those chips have got more pins. This may need to be split into two separate rows to prevent confusion about this one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ User can add a STM32 based board following this [wiki](https://github.com/stm32d
 | :green_heart:  | STM32F401CC | [WeAct Black Pill](https://stm32-base.org/boards/STM32F401CCU6-WeAct-Black-Pill-V1.2) | *1.7.0* | [More info](https://github.com/WeActTC/MiniF4-STM32F4x1) |
 | :yellow_heart: | STM32F411CE | [WeAct Black Pill](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0) | **1.9.0** | [More info](https://github.com/WeActTC/MiniF4-STM32F4x1) |
 | :green_heart:  | STM32F407VE<br>STM32F407VG | [STM32 F4VE / Black F407VET6](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0) | *1.4.0* | STM32F407VG support since *1.5.0* |
-| :green_heart:  | STM32F407ZE<br>STM32F407ZG | Black F407ZET6 | *1.5.0* |  |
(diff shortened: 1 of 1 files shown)


fpistm:
> ```
> | :green_heart: | [Black F407VET6](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0) | *1.4.0* | VG/ZE/ZG in *1.5.0* |
> ```
> 
> 
> The ZE/ZG variant of this board must be a completely different board, as those chips have got more pins. This may need to be split into two separate rows to prevent confusion about this one.

Right and it is managed. I guess it was left on the same line because they are all managed in the same variant folder:
Vx:
https://github.com/stm32duino/Arduino_Core_STM32/blob/1eac7c56ab65bd73b79d4237c00121a767cbcb24/variants/BLACK_F407XX/variant.h#L121
Zx:
https://github.com/stm32duino/Arduino_Core_STM32/blob/1eac7c56ab65bd73b79d4237c00121a767cbcb24/variants/BLACK_F407XX/variant.h#L251

The improvement seems nice for me, the main issue I can see is the link to the Board available which has been removed and it is used in several Wiki, docs,...
So I wonder if it will be possible to keep it for backward compatibility.

ThomasGravekamp:
Right, I'll add a new row to the table for the Zx variants of that particular board.

> The main issue I can see is the link to the Board available which has been removed and it is used in several Wiki, docs,...

I can't think of a way to keep the old anchor intact. The wiki can be updated easily by globally replacing the old link with the new link.

Which other docs rely on that anchor? Are these docs that are under control of stm32duino or do you mean external docs like websites linking to that anchor?

If you feel like changing the anchor breaks too many things, we can leave those changes out.

fpistm:
I guess it can be break, I will update in the wiki and in all other place I think (remember) 😉
For example here:
https://github.com/arduino/Arduino/wiki/Unofficial-list-of-3rd-party-boards-support-urls

ThomasGravekamp:
Force pushed a new commit with a new row for the Black STM32F407Zx board.

ThomasGravekamp:
Just added two new board pages to STM32-base. Added a link for the FK407M1 board to the list.

Issue #1029, opened 2020-04-06
probably USBD_reenumerate have issue after update


stas2z:
after updating local repo to the latest changes found my f405ve custom board (no special detaching/attaching pins, generic usb schematic like black f407) USB CDC stopped working properly

reverting e1d409f1203d5820e563e7be4bb084b57dc5b4b6 fixes it

i had no time to investigate it, but seems like reenumerate is not working properly after NVIC_SystemReset but i faced it a few times after detaching/attaching usb cable also
ive only tried to make CDC_deInit/CDC_Init to check if it come back to life, but it doesn't

@matthijskooijman 
can you check it?

if it will be ok from your side, i will try to find a time to sit with debugger to investigate

fpistm:
Hi @stas2z 
I will test on F407.
I guess you have this message during build?
https://github.com/stm32duino/Arduino_Core_STM32/blob/088081d09736a6c7479801bba2292579cd4dfed3/cores/arduino/stm32/usb/usbd_if.c#L147

stas2z:
> Hi @stas2z
> I will test on F407.
> I guess you have this message during build?
> https://github.com/stm32duino/Arduino_Core_STM32/blob/088081d09736a6c7479801bba2292579cd4dfed3/cores/arduino/stm32/usb/usbd_if.c#L147

oh, haven't seen it, project im working on atm have alot of warnings, cuz i always compile with -Wall)) so it's possible i missed it)
will check it a bit later

fpistm:
No worries 😉 
Probably in that case we should do also a toggling like before.

matthijskooijman:
The F407 has built-in USB pullups and an SDIS bit, so the commit will have changed behaviour to no longer apply the force-DP-low-trick, since that should not be needed anymore.

However, it seems the Black F407 actually has an external fixed pullup:

![image](https://user-images.githubusercontent.com/194491/78649661-09196500-78be-11ea-82e5-f4cb91048fa6.png)

@stas2z, can you confirm that you have the same on your board?

It's a bit weird to see a board like this, which actually reduces flexibility by fixing the pullup externally (and might actually also violate the USB spec since the two pullups combine to an effective 750Ω rather than the specified 1.5k). @stas2z if possible, you  might want to consider removing the external pullup.

[...]

PR #1048, opened 2020-04-28
Fix external and internal usb pullup

matthijskooijman:
This fixes boards that have external and internal USB pullups, which were broken by #997, as reported in #1029.

This needs a new define in the variant for these broken boards. I added this for two boards for which I could confirm that they had an external pullup. @stas2z, you have a black-based board, could you test this PR on your board?

I found that the CoreBoard F401RC had a switchable external pullup, but also internal pullups, so I changed that one to no longer use the external pullup and just the internal pullup. @mrguen, I think you have one of these boards, could you maybe test this PR on that board?

I tested this on a custom STM32F401 board, where I added an external pullup for testing. I also tested on an unmodified Blue pill, which still works as expected.

### DIFF
--- a/cores/arduino/stm32/usb/usbd_if.c
+++ b/cores/arduino/stm32/usb/usbd_if.c
@@ -68,6 +68,9 @@
 #if defined(USBD_DETACH_PIN) && !defined(USBD_DETACH_LEVEL)
   #error "USBD_DETACH_PIN also needs USBD_DETACH_LEVEL defined"
 #endif /* defined(USBD_DETACH_PIN) && !defined(USBD_DETACH_LEVEL) */
+#if (defined(USBD_DETACH_PIN) || defined(USBD_ATTACH_PIN)) && defined(USBD_FIXED_PULLUP)
+  #error "Cannot define both USBD_FIXED_PULLUP and USBD_ATTACH_PIN or USBD_DETACH_PIN"
+#endif /* (defined(USBD_DETACH_PIN) || defined(USBD_ATTACH_PIN)) && defined(USBD_FIXED_PULLUP) */
 
 /* Either of these bits indicate that there are internal pullups */
 #if defined(USB_BCDR_DPPU) || defined(USB_OTG_DCTL_SDIS)
@@ -78,10 +81,13 @@
  * in USBD_LL_Init in usbd_conf.c. */
 #if defined(USE_USB_HS)
   #define USBD_USB_INSTANCE USB_OTG_HS
+  #define USBD_DP_PINNAME USB_OTG_HS_DP
 #elif defined(USB_OTG_FS)
   #define USBD_USB_INSTANCE USB_OTG_FS
+  #define USBD_DP_PINNAME USB_OTG_FS_DP
 #elif defined(USB)
   #define USBD_USB_INSTANCE USB
+  #define USBD_DP_PINNAME USB_DP
 #endif
 
 /*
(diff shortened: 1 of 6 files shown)


stas2z:
Works for me

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -97,15 +103,13 @@
 #elif defined(USBD_DETACH_PIN)
   #define USBD_PULLUP_CONTROL_PINNAME digitalPinToPinName(USBD_DETACH_PIN)
   #define USBD_ATTACH_LEVEL !(USBD_DETACH_LEVEL)
-#elif !defined(USBD_HAVE_INTERNAL_PULLUPS)
-  /* When no USB attach and detach pins were defined, and there are also
+#elif !defined(USBD_HAVE_INTERNAL_PULLUPS) || defined(USBD_FIXED_PULLUP)
+ /* When no USB attach and detach pins were defined, and there are also
```
fpistm, line 107:
```suggestion
  /* When no USB attach and detach pins were defined, and there are also
```

matthijskooijman:
Heh, I had actually thought I *fixed* the indentation of that comment, didn't realize astyle had another opinion about it :-p Anyway, thanks for the fix, I've squashed it into my original commit.

@stas2z Thanks for testing!

cores/arduino/stm32/usb/usbd_if.c
```diff
@@ -97,15 +103,13 @@
...
   #define USBD_ATTACH_LEVEL !(USBD_DETACH_LEVEL)
-#elif !defined(USBD_HAVE_INTERNAL_PULLUPS)
+#elif !defined(USBD_HAVE_INTERNAL_PULLUPS) || defined(USBD_FIXED_PULLUP)
   /* When no USB attach and detach pins were defined, and there are also
   * no internal pullups, assume there is a fixed external pullup and apply
-  * the D+ trick. This should happen only for the USB peripheral, since
-  * USB_OTG_HS and USB_OTG_FS always have internal pullups. */
-  #if !defined(USB)
-    #error "Unexpected USB configuration"
-  #endif
```
fpistm, line 107: Maybe we should kept this if `USBD_FIXED_PULLUP` is not defined ?

`#if !defined(USB) && !defined(USBD_FIXED_PULLUP)`

fpistm: Well no. This should never be raised.

matthijskooijman: Yeah, the previous check was just the result of some simplification of the USB-to-DP-pin code, which can no longer be simplified like that because. So this check would no longer be needed, I think.

fpistm:
# Tested on:
* BluePill: with or without bootloader (USB)
* Maple Mini: with or without bootloader (USB with `USBD_ATTACH_PIN` and `USBD_ATTACH_LEVEL`) 
* Black F407VE (USB_OTG_FS with `USBD_FIXED_PULLUP`)
* Adafruit Feather F405 (USB_OTG_FS with `USBD_HAVE_INTERNAL_PULLUPS`)
* RobotdynF303 (USB)
* Disco F746NG (USB_OTG_FS  and USB_OTG_HS with `USBD_HAVE_INTERNAL_PULLUPS`)

matthijskooijman:
Thanks for the extensive testing!

fpistm:
Welcome.
Thanks too ;)

mrguen:
@matthijskooijman on the CoreBoard F401RC the pullup is not permanent but driven by PD2

matthijskooijman:
> @matthijskooijman on the CoreBoard F401RC the pullup is not permanent but driven by PD2

Yeah, that's what I already said (or at least tried to say :-P). Regardless, that pullup is still problematic when used together with the internal pullup, so this PR changes the code to not use the external pullup anymore. If you could confirm that USB still works on this board, that would be awesome!

mrguen:
@matthijskooijman I don't want to impact my install too much. Is changing these files
Arduino_Core_STM32/cores/arduino/stm32/usb/usbd_if.c
Arduino_Core_STM32/variants/Generic_F401Rx/variant.h

enough to test?

matthijskooijman:
@mrguen 

>Is changing these files
Arduino_Core_STM32/cores/arduino/stm32/usb/usbd_if.c
Arduino_Core_STM32/variants/Generic_F401Rx/variant.h

Yeah, I think it should be.

mrguen:
I tried it succesfully. I hope the test is significant. Here's what I did:

- Install the current release https://github.com/stm32duino/Arduino_Core_STM32 instead of 1.8.0

-  Configuration: CoreBoard F401RCT6, USART enabled Generic Serial, USB Support: CDC generic serial supersede usart,  USB low/full speed 

-  Uploaded in native DFU mode

- Tested the sketch SerialCallResponse working

matthijskooijman:
Awesome, thanks!

PR #1054, opened 2020-05-04
Added STM32F072CB varant for the Elektor LoRa Node

lab-mathias-claussen:
**Summary**

As the STM32F072C8 ( 64kB ) is not proviiding much space for user applications after the LoRaWAN Stack is inside, the first production run will include a STM32F072CB ( 128kB) with the same pinout but a bit more flash. The Variant has been renamed to reflect this change, also a new Board has added to the menu so you can choose what you have populated.

* [x] Renamed variant to "ELEKTOR_F072Cx"
* [x] Added "ELEKTOR_F072CB" as baord with name "Elektor LoRa Node Core F072CB (128kB)"
* [x] Renamed "ELEKTOR_F072C8" to show in menu as "Elektor LoRa Node Core F072C8 (64kB)"
* [x] Changed Linker-Skript to suhpport both Flashsizes


boards.txt
```diff
@@ -1736,17 +1736,29 @@ LoRa.menu.pnum.RHF76_052.build.cmsis_lib_gcc=arm_cortexM0l_math
 LoRa.menu.pnum.RHF76_052.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial} -D__CORTEX_SC=0
 
 # ELEKTOR_F072C8
-LoRa.menu.pnum.ELEKTOR_F072C8=Elektor LoRa Node Core F072
+LoRa.menu.pnum.ELEKTOR_F072C8=Elektor LoRa Node Core F072C8 (64kB)
 LoRa.menu.pnum.ELEKTOR_F072C8.upload.maximum_data_size=16384
-LoRa.menu.pnum.ELEKTOR_F072C8.upload.maximum_size=131072
+LoRa.menu.pnum.ELEKTOR_F072C8.upload.maximum_size=65536
```
fpistm, line 1741: So previous version does not defined the correct Flash size. Nice catch.

lab-mathias-claussen: The Linkerscript was doing things right, but the uploadsize was off, yes a nasty entry.

PR #1062, opened 2020-05-09
Fixed setting HardwareTimer interrupt priority

sjasonsmith:
**Summary**

HardwareTimer interrupt priority is currently set in the constructor inside the call to `HAL_TIM_Base_Init`. Setting the priority using `HardwareTimer::setInterruptPriority` stored the priority, but it was only applied if the timer was re-initialized. This would be unlikely to ever occur for typical HardwareTimer users.

This change applies the interrupt priority immediately in addition to storing it, so that it impacts already initialized timers.

The motivation for this change is to allow Marlin firmware to utilize the latest version of this framework, without having to maintain workarounds to ensure proper interrupt priorities.

**Validation**

I verified this fix by reworking Marlin to build with the latest code from this repo. With this fix I am able to delete Marlin's HardwareTimer priority workarounds and custom copy of SoftwareSerial.

Without this change SoftwareSerial bit timing was inconsistent due to conflicting interrupts, causing communication errors.

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -35,9 +35,6 @@
 /* Private Variables */
 timerObj_t *HardwareTimer_Handle[TIMER_NUM] = {NULL};
 
-IRQn_Type getTimerUpIrq(TIM_TypeDef *tim);
-IRQn_Type getTimerCCIrq(TIM_TypeDef *tim);
-
 /**
   * @brief  HardwareTimer constructor: set default configuration values
   * @param  Timer instance ex: TIM1, ...
@@ -913,6 +910,10 @@ void HardwareTimer::setPWM(uint32_t channel, PinName pin, uint32_t frequency, ui
   */
 void HardwareTimer::setInterruptPriority(uint32_t preemptPriority, uint32_t subPriority)
 {
+  // Set priority for immediate use
+  NVIC_SetPriority(getTimerUpIrq(_timerObj.handle.Instance), NVIC_EncodePriority(NVIC_GetPriorityGrouping(), preemptPriority, subPriority));
+
+  // Store priority for use if timer is re-initialized
   _timerObj.preemptPriority = preemptPriority;
   _timerObj.subPriority = subPriority;
 }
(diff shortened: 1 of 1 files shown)


cores/arduino/HardwareTimer.cpp
```diff
@@ -913,6 +910,10 @@ void HardwareTimer::setPWM(uint32_t channel, PinName pin, uint32_t frequency, ui
   */
 void HardwareTimer::setInterruptPriority(uint32_t preemptPriority, uint32_t subPriority)
 {
+  // Set priority for immediate use
+  NVIC_SetPriority(getTimerUpIrq(_timerObj.handle.Instance), NVIC_EncodePriority(NVIC_GetPriorityGrouping(), preemptPriority, subPriority));
+
```
ABOSTM, line 915: IRQ number may be different for Update interrupt and Capture/Compare interrupt (depending on timer instance). Thus it may be necessary to update also Caputure/Compare priority. Add something like this:
```suggestion
if (getTimerCCIrq(_timerObj.handle.Instance) != getTimerUpIrq(_timerObj.handle.Instance)) {
  NVIC_SetPriority(getTimerCCIrq(_timerObj.handle.Instance), NVIC_EncodePriority(NVIC_GetPriorityGrouping(), preemptPriority, subPriority));
}
```

fpistm:
@sjasonsmith 
any update on this ?

sjasonsmith:
> @sjasonsmith
> any update on this ?

I intend to come back to it, but this is hobby work for me and I am back to my day job at the moment.

Let me know if you were planning to make a new release or anything that would require me to rush this change in. Otherwise I expect to update it within the next couple days.

fpistm:
I'm currently preparing the next release and hope to release it this week or at least next week. That's why I ask. 😉

cores/arduino/HardwareTimer.cpp
sjasonsmith: I made this update, and also replace NVIC_SetPriority with HAL_NVIC_SetPriority. It's not entirely clear to me which entrypoint I should be using from here, but the HAL_ version took care of encoding the priority for me.

sjasonsmith:
I implemented the requested change to update the Capture/Compare interrupt. I'm only working with Update interrupts in Marlin, so I didn't have a way to test that change. This is still working as expected for the Update interrupt.

PR #1070, opened 2020-05-15
Add Daisy Seed variant [replace #1063]

fpistm:
This PR replace the PR #1063 from @stephenhensley

## Electrosmith Daisy Variant

**Summary**

Adds the Electrosmith Daisy as a variant

The [Daisy](https://www.electro-smith.com/daisy) is a small form factor, audio-centric SOM embedding an STM32H750

It features a high-quality,stereo audio codec, 64MB of external SDRAM, 8MB of external flash,

Support for the additional memories/audio has not been added yet (audio seems like it should be done in its own library). So, for now the variant has 512kB of internal SRAM, and 128kB of internal flash mapped as the default memories.

InitVariant() needs to be added to initialize the external memories, but all of the external pins, and several of their alternate functions have been tested (SPI, I2C, UART, ADC, DAC, PWM, etc.).


stephenhensley:
@fpistm 

Thanks! Sorry for not getting back fast enough.

I'll definitely keep up on it more in the future. 

Once I've retested the hardware/peripherals with new clock tree settings I'll have an update that adds SDRAM init, and moves CPU freq to the maximum 480MHz.

Do you have a standard for handling alternate memories? (i.e. should I have a variant that uses external SDRAM as default bss, or just provide a macro for the appropriate section attribute?)

fpistm:
No worries,

anyway the 1.9.0 will be released today.

So update will be done on the next release.
For the alternate memories, I never do it anyway you can do all you want in the variant and linker script, you can also change the default startup file if needed. 
IIRW Malyan variant do something like this.

PR #1091, opened 2020-06-05
[variant] Improve variant support and genericity

### LINKED ISSUE #855: STM32G031K8T6
Atmel has the most popular atmega328p microcontroller. In my opinion, the STM32G031K8T6 is the best alternative.
Is it possible to add this chip and is there any reason for this in view of memory?

### LINKED ISSUE #857: STM32G031 Support?
Is the STM32G031 line of MCUs supported?  Easy enough to add?

Datasheet is here:
https://www.st.com/resource/en/datasheet/stm32g031j6.pdf

I know the STM32G071RB is supported - perhaps I can modify that definition?

### LINKED ISSUE #1144: Add USER_BTN mapping to PA0 for BlackPill F401CC/F401CE boards
Since v1.2 BlackPill F401CC/F401CE boards have an user button (named "KEY") connected to PA0 added to the board. You can find more details about this board [here](https://github.com/WeActTC/MiniF4-STM32F4x1)

Please add the USER_BTN define mapping to PA0 for BlackPill F401CC and the [proposed F401CE](https://github.com/stm32duino/Arduino_Core_STM32/issues/1143) boards. That would be consistent with the already existing [BlackPill F411CE USER_BTN mapping](https://github.com/stm32duino/Arduino_Core_STM32/blob/59dda5c138badcb0a17a24f7a30fa81dececa0d7/variants/Generic_F411Cx/variant.h#L83)

I believe the following change in the [Generic_F401Cx/variant.h](https://github.com/stm32duino/Arduino_Core_STM32/blob/59dda5c138badcb0a17a24f7a30fa81dececa0d7/variants/Generic_F401Cx/variant.h#L82) file should do the job:

> // On-board user button
> #if defined(ARDUINO_BLACKPILL_F401CC) || defined(ARDUINO_BLACKPILL_F401CE)
> #define USER_BTN                PA0
> #elif !defined(USER_BTN) && !defined(ARDUINO_GENERIC_F401CC)
> #define USER_BTN                PC13
> #endif

### LINKED ISSUE #1180: UART1 on pin PA10, PA9 does not work on Nucleo-G071RB
I found issue where UART1 is not working on pin **PA10 & PA9** on **Nucleo-G071RB** but UART1 works fine on pin **PB7 & PB6**.
 
I verified with following code in Arduino IDE and PlatformIO

 ```c
  

HardwareSerial rfUart(PA10, PA9);  // do not work
//HardwareSerial rfUart(PB7, PB6); // work successfully
  
void setup()
{
  Serial.begin(115200);
  delay(100);

  rfUart.begin(115200);

  pinMode(LED_BUILTIN, OUTPUT);

  Serial.println("\nSG0 Begin....");
  rfUart.println("\nSG0 RF Begin....");
}

int count = 0;
void loop()
{
  delay(2000);

  Serial.printf("\nSG0 Count: %02d", count);
  rfUart.print("\nRF Count:"); rfUart.println(count++);

  digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN));
}
```

**Note:** I tried using Stm32Cube Framework, UART1 on pin PA10 & PA9 works fine.

### LINKED ISSUE #1276: Cannot use PG9 as digital pin with NUCLEO-H743ZI2
- OS: Mac OS X
 - Arduino IDE version: 1.8.13
 - STM32 core version: 1.9.0
 - Board Name: NUCLEO-H743ZI2

**Describe the bug**
I try to use PG9 as digital output :
- `pinMode (PG9, OUTPUT) ;` in `setup`;
- `digitalWrite (PE9, HIGH) ;` has no effect, pin is not a high level digital output.

**Bug Fix**
I think the bug is in the `~/Library/Arduino15/packages/STM32/hardware/stm32/1.9.0/variants/NUCLEO_H743ZI/variant.h` file.

Line 174 is : `#define NUM_DIGITAL_PINS  99`, and line 164 is `#define PG9  99`. `PG9` is the last pin, so `NUM_DIGITAL_PINS` should be equal to `100`.

So changing the line 174 to `#define NUM_DIGITAL_PINS  100` solves the bug.

**Note**
I think the line 176 `#define NUM_DIGITAL_PINS  97` (for NUCLEO_H743ZI) should be also changed to `#define NUM_DIGITAL_PINS  98`, for the same reason : last pin is `PG12` (#97).

Best regards,

Pierre Molinaro

### LINKED ISSUE #1277: Cannot use PA6 as analog input with a NUCLEO-H743ZI2
- OS: Mac OS X
 - Arduino IDE version: 1.8.13
 - STM32 core version: 1.9.0
 - Board Name: NUCLEO-H743ZI2

**Describe the bug**
I try to use PA6 as analog input. Calling `analogRead (PA6)` returns always zero.

**Bug Fix**
I think the bug is in the `~/Library/Arduino15/packages/STM32/hardware/stm32/1.9.0/variants/NUCLEO_H743ZI/PeripheralPins.c` file.

Line 57 `{PA_6,  ADC1,  STM_PIN_DATA_EXT(STM_MODE_ANALOG, GPIO_NOPULL, 0, 3, 0)}, // ADC1_INP3` is commented.

Uncommenting this line solves the bug.

Best regards,

Pierre Molinaro

### LINKED ISSUE #1302: variants/NUCLEO_F767ZI/variant.h seems incomplete
I get compile error when I try to use the pin name PD10 (e.g. in a list of FSMC data pins).
The Nucleo-144 STM32F767ZIT6 144 PINS has many more pins than available as names (#define) in variant.h
The Nucleo-144 STM32F746ZGT6 144 PINS seems to have these pins in variant.h (Morpho Pins).

### PR DESCRIPTION
**Pull request by fpistm**
As the number of variants continues to grow and in order to ease navigation, they have been grouped in sub-folders per 
series.

To improve variants support and generic ones several enhancements/fixes has been done:
- Each variant will be linked to its mcu/package
- All pins capabilities will be possible without modifying the `PeriperhalPins.c` PinMap array. ALTx pin naming will allow to use all alternative possibilities which use other HW peripheral instances.
- All remap pins will be supported (F0/G0)
- All dual pad pins will be supported (H7)
- Analog definition clean up
- Several fixes in existing variants

All generic variants are now automatically generated thanks the [STM32_open_pin_data](https://github.com/STMicroelectronics/STM32_open_pin_data) repository which provides all the information required for the pin configuration of products based on STM32 MCU. All the generic variants are generated in the variant folder, this means all the generic STM32 MCU are generated. Only the linker script and the clock config are missing. A board_entry.txt file is generated to ease board declaration.

#### Migrated STM32 series: 
- [x] STM32F0
- [x] STM32F1
- [x] STM32F2
- [x] STM32F3
- [x] STM32F4
- [x] STM32F7
- [x] STM32G0
- [x] STM32G4
- [x] STM32H7
- [x] STM32L0
- [x] STM32L1
- [x] STM32L4
- [x] STM32L5
- [x] STM32MP1
- [x] STM32WB

#### Issues linked to this PR:
Fix #855, fix #857, fix #1180, fix #1276, fix #1277, fix #1302, fix #1144

### PR for new variant included in this one;
- [x] Aurora One  #1176
- [x] Generic F030C8 #1080
- [x] Generic F042x #1103
- [x] NUCLEO_F070RB #1167
- [x] Nucleo L552ZE-Q #1256
- [x] Generic L010RB #1123
- [x] b-g431b-esc1 discovery kit #1236
- [x] STM32F303VC Discovery #1238 
- [x] Olimexino-STM32F3 #1115
- [x] Add SDIO support to FK407M1 #1337
- [x] STM32F407ZGT6 Mini #1130
- [x] STM32F100Cx #1208

---
That would be fine to define the variant in a different way, like this for example:
```
GenYY.build.variant={build.series}/{build.subvariant}
...
Nucleo_144.menu.pnum.NUCLEO_XYZ.build.subvariant=Generix_XYZx
```

but it does not work. It builds fine until the compile core step. The automatic default include is simply ignored while for other step it works fine.
I guess there is an issue around the way variant is managed in the Arduino IDE. 
Any input are welcome.
@matthijskooijman do you have any idea about this issue? --> https://github.com/arduino/arduino-cli/issues/762

### CODE REVIEW

**Comment by fpistm** [comment]
This change impact the PIO and I don't know how to solve this.
@valeros, I'm currently study how to better handle variant, this PR is one solution I've planned.
How can we handle this correctly with PIO to ensure backward compatibility?

**Comment by fpistm** [comment]
@matthijskooijman 

I saw in 1.8.13 your rework of the boards menu per architecture https://github.com/arduino/Arduino/pull/9238

has you any though on this PR and this question ?

> That would be fine to define the variant in a different way, like this for example:
> 
> ```
> Nucleo_144.build.variant=Nucleo_144/{build.subvariant}
> ...
> Nucleo_144.menu.pnum.NUCLEO_F207ZG.build.subvariant=NUCLEO_F207ZG
> ```
> 
> but it does not work. It builds fine until the compile core step. The automatic default include is simply ignored while for other step it works fine.
> I guess there is an issue around the way variant is managed in the Arduino IDE.
> Any input are welcome.
> @matthijskooijman do you have any idea about this issue?

I think about open an issue about that or a question on the devel group list.

The way to define a board/variant does not allows enough possibilities and seems very restricted. I guess it would be fine to allows build submenu more easily and have a variant path more "dynamic"

**Comment by valeros** [comment]
Thanks for keeping me informed. We're going to update STM32 platform with the v1.9.0 release, so we can take into account possible future changes of variant paths. Is it already decided to introduce a intermediary folder
`variants/ARMED_V1/PinNamesVar.h → variants/3dprinter/ARMED_V1/PinNamesVar.h`?

**Comment by fpistm** [comment]
> 
> 
> Thanks for keeping me informed. We're going to update STM32 platform with the v1.9.0 release, so we can take into account possible future changes of variant paths. Is it already decided to introduce a intermediary folder
> `variants/ARMED_V1/PinNamesVar.h → variants/3dprinter/ARMED_V1/PinNamesVar.h`?

Hi @valeros 

currently this is under investigation. 1.9.0 is not impacted.
Could you give me some tricks how to manage this during my testing?

**Comment by valeros** [comment]
Probably the best way is to fork `platform-ststm32` and change `variant` field in the board manifests that used in CI checks blackpill_f103c8 remram_v1), e.g.:
```
{
  "build": {
    "core": "stm32",
    ...
    "variant": "ARMED_V1"
  },
  ....
}
``` 

```
{
  "build": {
    "core": "stm32",
    ...
    "variant": "3dprinter/ARMED_V1"
  },
  ....
}
```

**Comment by matthijskooijman** [comment]
@fpistm, that `{build.variantvariant}` does not work seems to be a limitation or bug in the `arduino-cli` code (which is used by the Arduino IDE via `arduino-builder`). I reported a bug, look there for more details: https://github.com/arduino/arduino-cli/issues/762

A workaround would be what you've done now: Just repeat the group path in each menu option. Not elegant, but at least it works. Maybe if Arduino fixes this, it can be changed later. I can't comment on PlatformIO, but I think @valeros can :-)

Is this what you wanted feedback on? Or is there any other question I missed?

**Comment by fpistm** [comment]
Thanks @matthijskooijman !!

> Note that the compiler command has no -I option for the variant when compiling the core (but it is present for the sketch).

Yes that's what I saw during my test. For sketch it works but not for compiler while with the method used in this PR the -I option is available with the correct path.

In fact when I saw your rework of the menu, I'm thinking about why we do this kind of boards.txt with the boards group and then the real variant. This is due to the fact it is not possible to construct a submenu.

What would be fine would be to be able to construct a menu with submenu easily and then also be able to link upload method per submenu to avoid duplicate upload boards entries for each group.

Anyway, thanks again.

Just FYI, I've an empty menu with Roger's core. I guess this is due to empty boards.txt in drivers and tools folders:
image

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/drivers/boards.txt
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/tools/boards.txt

Maybe this file can now be safely deleted or simply Arduino should ignore empty file?

**Comment by matthijskooijman** [comment]
> In fact when I saw your rework of the menu, I'm thinking about why we do this kind of boards.txt with the boards group and then the real variant. This is due to the fact it is not possible to construct a submenu.

> What would be fine would be to be able to construct a menu with submenu easily and then also be able to link upload method per submenu to avoid duplicate upload boards entries for each group.

Well, reworking the nested menu was easy, that was just GUI code, it did not require changes to the `boards.txt` interpretation... Reading what you wrote, I thought it could be nice to allow grouping of boards (e.g. to have submenus within the "Boards" -> "STM32..." menu). However, that would probably only cause *more* duplication of stuff in `boards.txt` rather than less.

This is because the extra menus are now always defined per-board, so there is no way to have common (per-core) menus. I've realized this would be a good feature to add, but I haven't really sat down to think about this, let alone implement it (and it changes semantics of `boards.txt`, so that would not be possible in a backward-compatible manner).

> Just FYI, I've an empty menu with Roger's core. I guess this is due to empty boards.txt in drivers and tools folders:

Ah, that's a completely separate issue that is triggered by my nested boards change. I'm not sure what the best fix for this is, since Roger's core does sort of abuse the system by providing an empty `boards.txt`. Maybe you should file a bug at Roger's repository to further discuss this, since it's a bit off-topic for this issue.

**Comment by fpistm** [comment]
> Ah, that's a completely separate issue that is triggered by my nested boards change. I'm not sure what the best fix for this is, since Roger's core does sort of abuse the system by providing an empty `boards.txt`. Maybe you should file a bug at Roger's repository to further discuss this, since it's a bit off-topic for this issue.

As said it is just for your interest and of course off-topic ;)


>  However, that would probably only cause _more_ duplication of stuff in `boards.txt` rather than less.

OK. Just a thought as I think this could be probably better managed. Anyway, I guess this would be really useful only for this core. We planned to deployed more generic variant and so this will add a lot of entry in the boards.txt. Hope there is no limit in Arduino IDE (even the new pro IDE) and also with arduino-cli.

**Comment by fpistm** [comment]
Thanks @valeros  for the hints. I've made the patch in the GH action and it works well:
https://github.com/stm32duino/Arduino_Core_STM32/pull/1091/commits/38f6145fecd123b21dea9a267afa6d41a0350de9

I wonder if it could be possible to manage backward compatibilities in the script which manages the variant JSON entry to test if the path exists or not ?
Maybe having in the boards JSON:
```json
    "variant": "Generic_F1/PILL_F103XX",
    "alt_variant": "PILL_F103XX",
```

valeros:
@fpistm I think we can handle this is in the build script instead of adding a new filed `alt_variant` to the board manifests. If the path `Generic_F1/PILL_F103XX` doesn't exist, then we can use only the latter `PILL_F103XX`.

fpistm:
Thanks @valeros 
This would be fine and ensure backward compatibility.

ABOSTM:
I think that having 2 ways of categorization at the same level (by manufacturer and by board type) may be ambiguous, there may be some overlap. What to do when manufacturer XXX  introduce a 3D printer board ?
So I would prefer only 1 rule, or several rules at several level (ex manufacturer 1st level, board type second level).

That said, I would be in favor of 1st level being STM32 series (F1, F4, ...), this could help future enhancement where a board inherit from Generic implementation.

fpistm:
Thanks for you feedback @ABOSTM 
This make sense as the core is more MCU oriented.
I guess this need more investigation and probably this will help to factorize boards definitions, mainly for generic ones.

Any feedback are welcome. 😉

boards.txt
```diff
@@ -2014,88 +2012,88 @@ LoRa.menu.upload_method.dfuMethod.upload.tool=stm32CubeProg
 
 
 ###############################
-# Genericflight Flight Controllers
+# Generic Flight Controllers
 
-Genericflight.name=Generic flight controllers
+GenFlight.name=Generic Flight Controllers
```
per1234, line 2017: FYI, this board ID change is breaking because it changes the FQBN used to identify the board via the command line (both for Arduino IDE and Arduino CLI).

Before: `STM32:stm32.Genericflight`
After: ``STM32:stm32.GenFlight`

fpistm: Yes but I see no issue. It changes that's all or maybe I missed something ?

per1234: Well, I consider the FQBN part of the API of a boards platform, and thus cause for a major version bump.

Anyone who tries to use the old FQBN in a command will find it's now broken:

```
$ arduino-cli compile --fqbn STM32:stm32.Genericflight
Error during build: Error resolving FQBN: board STM32:stm32.Genericflight@1.9.1:Genericflight not found
```

This is especially problematic for automated systems like CI.

If you think it should be changed, fine, but I just want you to be aware that this isn't just cosmetic. It should certainly be prominently documented in the changelog.

fpistm: OK. Yes there will be a major version bum (2.0.0). I will also change the maintainer name as STM32 is not a vendor nor a maintainer and of course all will be documented in the changelog.

FQBN will looks like: `STMicroelectronics:stm32`

For our CI the change will not change anything as the script get the fqbn dynamically.

Thanks for your feedback.

per1234: A non-breaking option would be to make a copy of the board definition, so you have one definition with the deprecated `Genericflight` board ID and another with the new `GenFlight` ID, then add a `hide` property to the deprecated board definition:
https://arduino.github.io/arduino-cli/latest/platform-specification/#hiding-boards
```
Genericflight.hide=
```
This will cause the Arduino IDE to not display the deprecated board in its **Tools > Board** menu. The `hide` property is not yet supported by Arduino CLI or Arduino Pro IDE, so they will still show the deprecated board option.

fpistm: Thanks for pointing this, anyway I don't think the `hide` option is required in this case.

cypheron:
What's the ETA on still? Currently all new feature related PR's are blocked because of this. Reading this post I see lots of uncertainty, until a scheme is decided I therefore would suggest adding features and keep the contributions flowing.

dnandha:
What's the ETA on this? Currently all new feature related PR's are blocked because of this. Reading this post I see lots of uncertainty. Until a scheme is decided, I'd suggest adding features and keep the contributions flowing.

cypheron:
What's the ETA on this PR? There hardly has been any activity in the recent months and all new feature PR's are blocked because of this.. I personally think that the proposed grouping scheme by subvariant / model type is very intuitive and don't see a reason not to go on, once the board definitions pointed out by @per1234 have been added.

matthijskooijman (commented):
Just hade a look over these changes, looks quite good! Some nice cleanups and good that alternative pin mappings are now available :-D Did take me a while to figure out how it worked, but now I see how it works, I like it. I was initially assuming that the `ALTx` modifiers would correspond to the `AF` numbers, but now I see that these are just alternatives within a particular pin mapping, so most pins have only one alternative within a pin mapping so just use `ALT0 (or none at all). Seems like a good approach. One downside is that people must look inside the the variant files to see which alternatives are available (i.e. they cannot look this up in the datasheet), but that's probably fine (and can be improved by pin mapping documentation, I wonder if this could maybe even be autogenerated). Another downside could be that e.g. `PA_2_ALT0` refers to different functions in different contexts (i.e. when passing to adc or spi code), which might be a bit tricky to document, but I think that's ok.

I haven't actually tested the code yet, will do that next.

matthijskooijman:
Also tested this on Nucleo G031K8, worked like a charm. I used Serial, Wire and BUILTIN_LED, all of which seem to work fine.

cores/arduino/pins_arduino.c
```diff
@@ -101,11 +102,11 @@ uint32_t digitalPinToAnalogInput(uint32_t pin)
 #if NUM_ANALOG_INPUTS > 0
   if ((pin & PANA) == PANA) {
     /* PYn = Ax */
-    ret = pin & PANA_IDX;
+    ret = (pin & PANA_IDX) & (pin & ALTX);
```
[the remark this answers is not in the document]
fpistm: Nice catch! Thanks

cores/arduino/pins_arduino_analog.h
```diff
@@ -0,0 +1,313 @@
...
+#endif
+#if defined(ADC5) && defined(ADC_CHANNEL_TEMPSENSOR_ADC5)
+  #define ATEMP_ADC5   (NUM_ANALOG_INTERNAL_FIRST + 4)
+#endif
+
+/* If NUM_ANALOG_INPUTS is not defined there is no analog pins defined. */
+/* Anyway ADC internal channels are always available. */
+#if NUM_ANALOG_INPUTS > 0
+  #define PIN_A0       NUM_ANALOG_FIRST
+  static const uint8_t A0 = PIN_A0;
```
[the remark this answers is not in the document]
fpistm: 3 years back it was an enum but it was requested for compatibility to move as a macro in order to be able to use preprocessor directive which is not possible with an enum: https://github.com/stm32duino/Arduino_Core_STM32/issues/140

cores/arduino/pins_arduino_digital.h
```diff
@@ -0,0 +1,555 @@
...
+ *
+ *******************************************************************************
+ */
+#ifndef _PINS_ARDUINO_DIGITAL_H_
+#define _PINS_ARDUINO_DIGITAL_H_
+#include "variant.h"
+
+/* GPIO port (A to K + Z) * 16 pins: 192 */
+/* Arduino digital pins alias */
+#define D0       0
```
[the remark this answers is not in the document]
fpistm: AFAIK Using an enum or a macro does not change the issue if `Dx` is used in other code as it is global. I've changed to be inline with Ax definition and allow preprocor directive at build time.

fpistm:
> Just hade a look over these changes, looks quite good! Some nice cleanups and good that alternative pin mappings are now available :-D 

That's the goal 🤣 

> Did take me a while to figure out how it worked, but now I see how it works, I like it. 

I will add a dedicated section in the wiki.

I was initially assuming that the `ALTx` modifiers would correspond to the `AF` numbers, but now I see that these are just alternatives within a particular pin mapping, so most pins have only one alternative within a pin mapping so just use `ALT0` (or none at all). Seems like a good approach.

Right, that's why a good documentation will be required.
In fact, this ALTx is mainly inspired by STM32 support in MBED:
https://os.mbed.com/teams/ST/wiki/pinout_labels
Unfortunately, I had to manage also the Arduino layer with pin number definition.

> One downside is that people must look inside the the variant files to see which alternatives are available (i.e. they cannot look this up in the datasheet), but that's probably fine (and can be improved by pin mapping documentation, I wonder if this could maybe even be autogenerated). 

I agree with you. And I would like autogenerate this also when got some free time for that 😢 .

> Another downside could be that e.g. `PA_2_ALT0` refers to different functions in different contexts (i.e. when passing to adc or spi code), which might be a bit tricky to document, but I think that's ok.

Yes, but as you told it is in a different context. One per peripheral pinmap array and alway point to the correct GPIO port/pin.

cores/arduino/pins_arduino_analog.h
```diff
@@ -15,34 +15,38 @@
 #include "variant.h"
 #include "PinNames.h"
 
+/* Analog */
+#define PANA       0xC0
+/* Analog index */
+#define PANA_IDX   0x3F
```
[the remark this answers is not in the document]
fpistm: Right, I've made some update here : https://github.com/stm32duino/Arduino_Core_STM32/pull/1091/commits/f4a579d6e62625bb4cdbcf4e821f7f4d7d4148ef
Let me know if it is better/OK from your point of view 😉

fpistm:
> 
> 
> Maybe some macro/function to check PIN_IS_ANALOG?

```
bool digitalpinIsAnalogInput(uint32_t pin);
uint32_t digitalPinToAnalogInput(uint32_t pin);
```

cores/arduino/pins_arduino_analog.h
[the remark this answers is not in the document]
matthijskooijman: Yeah, perfect! Thanks!

cores/arduino/pins_arduino_analog.h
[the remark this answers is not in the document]
matthijskooijman: That issue is about `PIN_Ax`, which are now already macros and I was not suggesting to change those (because the names are longer, the chances of conflicts due to macros are a little smaller). My suggestion was to change the `Ax` constants to an enum. Not a very big deal, though.

cores/arduino/pins_arduino_digital.h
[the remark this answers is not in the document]
matthijskooijman: But for Ax, the `Ax` constants are *not* macros, only the (longer, thus less likely to cause problems) `PIN_Ax` constants are macros.

Consider a line of code like this:

```
void foo() {
  int D0 = 0;
}
```

With the changes in this commit (changing `D0` from an enum to a macro), that code no longer compiles, because the `D0` variable name is now expanded as a macro, yielding `int 0 = 0;`.

cores/arduino/pins_arduino_analog.h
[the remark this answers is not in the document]
fpistm: In fact we have mimic the way Arduino define Ax:
https://github.com/arduino/ArduinoCore-samd/blob/0b60a79c4b194ed2e76fead95caf1bbce8960049/variants/arduino_zero/variant.h#L94-L108
or 
https://github.com/arduino/ArduinoCore-mbed/blob/b86aaa794a3ca134ed727c710533ff253f3af539/variants/PORTENTA_H7_M4/pins_arduino.h#L26-L39

That's why I've aligned the Dx also on macro.
Anyway, I understood your point of view.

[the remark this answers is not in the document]
matthijskooijman: > I assume a rebase is in order @matthijskooijman after my merging of #137?

Yeah, but I'd think it could be a (small) improvement for Arduino as well to switch from `const` to `enum` for the Ax constants.

> Anyway, I understood your point of view.

Ok :-)

cores/arduino/pins_arduino_digital.h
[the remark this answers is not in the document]
fpistm: @matthijskooijman  I've revert Dx to an enum.

cores/arduino/pins_arduino_analog.h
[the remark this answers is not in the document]
fpistm: @matthijskooijman 
I've switched Ax to enum. :wink:

cores/arduino/pins_arduino_digital.h
```diff
@@ -535,4 +535,552 @@ enum {
...
+#ifdef PA13
+  #define PA13_ALT0      (PA13 | ALT0)
+#endif
+#ifdef PA14
+  #define PA14_ALT0      (PA14 | ALT0)
+#endif
+#ifdef PA15
+  #define PA15_ALT0      (PA15 | ALT0)
+#endif
+#ifdef PB1
```
ABOSTM, line 590: I looks like PB0_ALT0 is missing.
The same for all Py0_ALT0 (except for PA0_ALT0)

keywords.txt
```diff
@@ -726,6 +726,17 @@ PADC_TEMP	LITERAL1
...
+ALT1	LITERAL1
+ALT2	LITERAL1
+ALT3	LITERAL1
+ALT4	LITERAL1
+ALT5	LITERAL1
+ALT6	LITERAL1
+ALT7	LITERAL1
+PDUAL	LITERAL1
+PREMAP	LITERAL1
+
```
ABOSTM, line 745: Is it also required to add Pyx_ALTz  as literal ?

ABOSTM (dismissed):

keywords.txt
fpistm: I don't think

cores/arduino/pins_arduino_digital.h
fpistm: Right. Added.

matthijskooijman:
I worked a bit more with this code (defined my own variant on top of it), and had a few more thoughts:
 - ~~Would it make sense to rename `VARIANT_BOARD_H` to `SUBVARIANT_H` and `variant_*.h` to `subvariant_*.h`? It took me a few reads to see what was going on there, and I guess those are really board-specific subvariants? Or maybe not so much subvariants as actual board files, so maybe keeping `VARIANT_BOARD_H` (meaning the board header file within the variant) and renaming `variant_*.h` to `board_*.h`?~~
 - Looking a bit more closely at the `VARIANT_BOARD_H` stuff, it seems that this is not a board-specific *additional* include, but it replaces the `variant.h` completely. I wonder if would be better to make this more explicit. So instead of using `build.board` to construct a board-specific filename, use a `build.variant_header` variable or something like that. And then instead of having a generic version in `variant.h` and replacing that with an included file based on some (slightly hard to read) ifdefs checking board defines, just let *all* boards define a `build.variant_header` (including the generic boards) and have a `variant.h` that just does `#include VARIANT_H` or something like that (and has no content of its own, except maybe include guards). For the generic boards, you'd then just have a `variant_generic.h` file (with the default contents that is now in `variant.h`) and for specific boards, you could have a `variant_someboard.h` file (or, if the generic file is ok for a specific board, it could just refer to the generic file as well in `boards.txt`). This does leave the question what to do with `variant.cpp`, which currently checks the same defines. It could maybe check against `VARIANT_H`, (so the .h and .cpp files always match) except that strings comparisons are not supported by the preprocessor, I believe.
 - The new `ALTx` stuff is nice, since it offers full control over pin mappings. However, AFAICS do determine which pin constant you need, you have to look at the variant files, and for compatibility, we can later never change e.g. which peripheral corresponds to which `Pxx_ALTx` constant. How about defining additional pin aliases for each pin-peripheral combination? e.g. you could have `#define PB0_USART1 PB0` and ` #define PB0_USART2 PB0_ALT1`, and sketches can then use those constants based on information from the datasheet, without having to look at the variant. Additionally, if the `ALTx` mapping is ever changed, those sketches will stay working. Effectively the `ALTx` mapping and constants now become an implementation detail internal to the variant.
 -  `variant.h` files now contain an `extern "C"` block, but that (looking at the template, necessarily) only contains `#define`s, so I think these blocks can be omitted? Similarly, the `variant.cpp` file only contains variables, not functions, so I think the `extern "C"` block is not needed there either (alternatively, it could be renamed to `variant.c` maybe). Also note that for `PeripheralPins.c`, a different override mechanism is used (using `-DCUSTOM_PERIPHERAL_PINS` which disables the default `PeripheralPins.c`.
 - I was missing an LD script for the `G030C(6-8)Tx` variant, so I used the `G030K(6-8)Tx` one, which looks applicable. Are those still manually added and/or generated by CubeMX? Looking at the linker scripts, I see that flash and RAM sizes are now defined externally through commandline defsyms, so I wonder if there is still a reason for per-variant ldscripts? I suspect that for most boards, just a single generic ldscript could be sufficient? Looking at some ldscripts, I see that some of them already are identical, and some are only slightly different (extra section, or extra align statements, which suggest these are just generated by different version of CubeMX). It would be interesting to know what template CubeMX uses to generate these linker scripts and if there are any other differences besides the flash and ram sizes between chips.

fpistm:
>     * Looking a bit more closely at the `VARIANT_BOARD_H` stuff, it seems that this is not a board-specific _additional_ include, but it replaces the `variant.h` completely. I wonder if would be better to make this more explicit. So instead of using `build.board` to construct a board-specific filename, use a `build.variant_header` variable or something like that. And then instead of having a generic version in `variant.h` and replacing that with an included file based on some (slightly hard to read) ifdefs checking board defines, just let _all_ boards define a `build.variant_header` (including the generic boards) and have a `variant.h` that just does `#include VARIANT_H` or something like that (and has no content of its own, except maybe include guards). For the generic boards, you'd then just have a `variant_generic.h` file (with the default contents that is now in `variant.h`) and for specific boards, you could have a `variant_someboard.h` file (or, if the generic file is ok for a specific board, it could just refer to the generic file as well in `boards.txt`). This does leave the question what to do with `variant.cpp`, which currently checks the same defines. It could maybe check against `VARIANT_H`, (so the .h and .cpp files always match) except that strings comparisons are not supported by the preprocessor, I believe.

As usual several possibilities exists. I choose this one because the main entry point is variant.h and I do not see the interest to add one other files per variant folders and add a new build recipe in the boards.txt.

>     * The new `ALTx` stuff is nice, since it offers full control over pin mappings. However, AFAICS do determine which pin constant you need, you have to look at the variant files, and for compatibility, we can later never change e.g. which peripheral corresponds to which `Pxx_ALTx` constant. How about defining additional pin aliases for each pin-peripheral combination? e.g. you could have `#define PB0_USART1 PB0` and ` #define PB0_USART2 PB0_ALT1`, and sketches can then use those constants based on information from the datasheet, without having to look at the variant. Additionally, if the `ALTx` mapping is ever changed, those sketches will stay working. Effectively the `ALTx` mapping and constants now become an implementation detail internal to the variant.

Why not, this can be eventually added later and generated by the python script. But this will be in other PR as I need to release the 2.0.0.


>     * `variant.h` files now contain an `extern "C"` block, but that (looking at the template, necessarily) only contains `#define`s, so I think these blocks can be omitted? Similarly, the `variant.cpp` file only contains variables, not functions, so I think the `extern "C"` block is not needed there either (alternatively, it could be renamed to `variant.c` maybe). Also note that for `PeripheralPins.c`, a different override mechanism is used (using `-DCUSTOM_PERIPHERAL_PINS` which disables the default `PeripheralPins.c`.

Tight, nice catch. I will update the template and regenerate.

matthijskooijman:
>  As usual several possibilities exists. I choose this one because the main entry point is variant.h and I do not see the interest to add one other files per variant folders and add a new build recipe in the boards.txt.

Not sure I understand what you're saying here, why would you need additional build recipes? Also, recipes are defined in platform.txt, not board.txt? Did you mean additional variables in boards.txt maybe?

> Why not, this can be eventually added later and generated by the python script. But this will be in other PR as I need to release the 2.0.0.

Another PR sounds perfect, though maybe you want to tackle this before releasing 2.0.0 and people end up using the `ALTx` constants and you end up needing to support those indefinitely.

fpistm:
> Not sure I understand what you're saying here, why would you need additional build recipes? Also, recipes are defined in platform.txt, not board.txt? Did you mean additional variables in boards.txt maybe?

yes. :wink:

matthijskooijman:
Right, then my tendency is that explicit is better than implicit, but also concise is better than verbose, so there's a balance there. I guess the current solution is fine, it just takes a close reading to see what's happening exactly, but that's ok :-)

matthijskooijman:
> The main advantage now is that all "based files" are updated automatically in the core. Before we need to manually merged/update all file (mainly the PeripheralPins.c).

I wasn't suggesting to lose this, just to e.g. put the actual content that is now in `variant.h` in a file called `variant_generic.h` (or so) and then reference that explicitly from the boards that need it.

fpistm:
> I wasn't suggesting to lose this, just to e.g. put the actual content that is now in `variant.h` in a file called `variant_generic.h` (or so) and then reference that explicitly from the boards that need it.

@matthijskooijman 
I've clean up the extern C and include guards, replaced by `#pragma once` directive.
`variant.h` is now unique and include properly the right `variant_*.h` name (default `variant_generic.h`)

fpistm:
Last update introduce a new definition "VARIANT_H"  to specify which variant header file to include.
It impacts PIO build.

@valeros, could you help to solve this, please ?
https://github.com/stm32duino/Arduino_Core_STM32/pull/1091/checks?check_run_id=2120953299#step:3:817
```
 /github/home/.platformio/packages/framework-arduinoststm32/cores/arduino/variant.h:15:10: error: #include expects "FILENAME" or <FILENAME>
   15 | #include VARIANT_H
      |          ^~~~~~~~~
```

I've tried to solve it by modifying the  board JSON to add the VARIANT_H to the `data['build']['extra_flags']` but without success.
See https://github.com/stm32duino/Arduino_Core_STM32/pull/1091/commits/433fb777893b86b356d434df6b081bdc6f72c68d

Thanks in advance

valeros:
Hi @fpistm ! It seems we need to properly escape the file name, could you please try the following flags?
```
data['build']['extra_flags'] = '-DSTM32F765xx -DVARIANT_H=\\\"variant_REMRAM_V1.h\\\"'
```

fpistm:
> 
> 
> Hi @fpistm ! It seems we need to properly escape the file name, could you please try the following flags?
> 
> ```
> data['build']['extra_flags'] = '-DSTM32F765xx -DVARIANT_H=\\\"variant_REMRAM_V1.h\\\"'
> ```

@valeros  Unfortunately this does not change the issue. Like if the define was not take in account.

valeros:
Looks like we need even more backslashes:
```
-DVARIANT_H=\\\\\"variant_PILL_F103XX.h\\\\\"
```

fpistm:
> 
> 
> Looks like we need even more backslashes:
> 
> ```
> -DVARIANT_H=\\\\\"variant_PILL_F103XX.h\\\\\"
> ```

I will try anyway it was well formated with your first proposal.

jhleekr (changes requested):
Hi @fpistm, Sorry to disturb you. The link and email address have changed.
https://www.bfykorea.com/aurora-one/ (Not working) -> https://www.bfy.kr/aurora-one/
jhlee@bfykorea.com (Not working) -> jhlee@bfy.kr
If you can, please include this change. Thank you!

fpistm:
> 
> 
> Hi @fpistm, Sorry to disturb you. The link and email address have changed.
> https://www.bfykorea.com/aurora-one/ (Not working) -> https://www.bfy.kr/aurora-one/
> [jhlee@bfykorea.com](mailto:jhlee@bfykorea.com) (Not working) -> [jhlee@bfy.kr](mailto:jhlee@bfy.kr)
> If you can, please include this change. Thank you!

Hi @jhleekr, it has been updated

variants/STM32H7xx/H742Z(G-I)T_H743Z(G-I)T_H747A(G-I)I_H747I(G-I)T_H750ZBT_H753ZIT_H757AII_H757IIT/variant_NUCLEO_H743ZI.h
```diff
@@ -0,0 +1,407 @@
...
+  #define PD1                     66 // CAN_TX
+  #define PD0                     67 // CAN_RX
+  #define PF0                     68 // I2C_B_SDA
+  #define PF1                     69 // I2C_B_SCL
+  #define PF2                     70 // I2C_B_SMBA
+  // 71 is PE9                   (6)
+  // 72 is PB2                   (27)
+  #define PA3                     PIN_A0
+  #define PC0                     PIN_A1
+  #define PC3_C                   A2
```
ABOSTM, line 94:
```suggestion
  #define PC3_C                   PIN_A2
```

variants/STM32H7xx/H742Z(G-I)T_H743Z(G-I)T_H747A(G-I)I_H747I(G-I)T_H750ZBT_H753ZIT_H757AII_H757IIT/variant_NUCLEO_H743ZI.h
```diff
@@ -0,0 +1,407 @@
...
+  #define PB1                     PIN_A3
+  #define PC2_C                   A4
```
ABOSTM, line 96:
```suggestion
  #define PC2_C                   PIN_A4
```

variants/STM32H7xx/H742Z(G-I)T_H743Z(G-I)T_H747A(G-I)I_H747I(G-I)T_H750ZBT_H753ZIT_H757AII_H757IIT/variant_NUCLEO_H743ZI.h
```diff
@@ -0,0 +1,407 @@
...
+  #define PD1                     66  // CAN_TX
+  #define PD0                     67  // CAN_RX
+  #define PF0                     68  // I2C_B_SDA
+  #define PF1                     69  // I2C_B_SCL
+  #define PF2                     70  // I2C_B_SMBA
+  // 71 is PA7                   (11)
+  // 72 is                        NC
+  #define PA3                     PIN_A0
+  #define PC0                     PIN_A1
+  #define PC3_C                   A2
```
ABOSTM, line 214:
```suggestion
  #define PC3_C                   PIN_A2
```

variants/STM32H7xx/H742Z(G-I)T_H743Z(G-I)T_H747A(G-I)I_H747I(G-I)T_H750ZBT_H753ZIT_H757AII_H757IIT/variant_NUCLEO_H743ZI.h
```diff
@@ -0,0 +1,407 @@
...
+  #define PF3                     PIN_A3
+  #define PF5                     PIN_A4
+  #define PF10                    PIN_A5
+  #define PB1                     PIN_A6
+  #define PC2_C                   A7
```
ABOSTM, line 219:
```suggestion
  #define PC2_C                   PIN_A7
```

ABOSTM (changes requested):

matthijskooijman:
\o/

fpistm:
Finally done!

dreamcat4:
that was some huge merge. well done! very good work.

sorry if this is wrong place to ask. but had a simple little question (linked over on other issue), just if anybody here happens to know the answer

https://github.com/Serasidis/STM32_HID_Bootloader/issues/23#issuecomment-817866179

if these newly commited improvement can... lets me flash an hid bootloader and memory magic for flashing mode... without manually be pressing the reset button to reboot it every time. i think it means some special signal has to be sent over HID usb connection. to interrupt the running userland program, ad jump into the hid bootloader. which can then do all the flashing, reboot etc.

because for the life of me... i cannot find the answer to this question! yet the weact black pill f4xx is a very popular board. i searched and searched

fpistm:
@dreamcat4 

Thanks.

For HID. I saw several issue linked to the WeAct HID fork. I don't know why so many issue... Anyway, I do not use it and HID seems no more maintained by @Serasidis. 
I have to end #710 which is more generic (except for F1 which doesn't have USB support in builtin bootloader) and do not need extra stuff.

avtoneru:
> Wiki will be updated after official 2.0.0 core release

When can we expect an update?

ABOSTM:
@ridvanmelih

I looks like there is something wrong with github:
I receive a email notification about the following comment 
> Do we have a chance to define the SPI2 pins here?

related to this commit (part  of this PR)

> Re: [stm32duino/Arduino_Core_STM32] Add all generated STM32F1xx generic variant files (6b2b23d)

But I could not found the comment in github.
Consequently I don't know the context of this comment
* which directory is concerned
* which file  is concerned
* which line is concerned
* which MCU/board is concerned

What I can tell is that generic files are automatically generated, and all supported spi instances are made available.
So maybe the concerned MCU doesn't have spi2, or you are on a board/MCU which doesn't output spi2 pins

ridvanmelih:
Hello Alexandre,

Thank yo for your interest.

Yes, Yesterday I wrote a message and after that delete, because that’s true or not, I am not sure.



When I look at the MCU STM32F103C8T6 and the datasheet I am using, I see that there are two SPIs in this MCU and I made my PCB design considering this. 

 



 

Both of the ICs I have used on my card use the SPI communication protocol. For this reason, I had to make my software accordingly. 

One of the ICs I have used is the NRF24 module. I defined this module to SPI2 and connected it to SPI1 in my other integration. But I could not establish any communication with NRF24. 

Afterwards, I realized that there is no SPI2-related definition in the "variant_generic.h" code. 

 

PART OF variant_generic.h

 

As such, I defined the SPI2 pins to the SPI pins defined in the "variant_generic.h" code in order to at least determine the operability of my card and whether the SPI2 pins are usable. 

Later, I realized that my card was working, so I made sure that the SPI2 pins of the MCU I used were also working.

My current goal is to add the SPI1 and SPI2 pins to the relevant part (But I'm not sure which part, I just have a few thoughts.)

Maybe these libraries:

·         PeripheralPins.h

·         PinAF_STM32F1.h

 

You can find this codes: STMicroelectronics\hardware\stm32\2.0.0\cores\arduino\stm32

 

Thank you for your interest. I'd appreciate it if you could help me out or start a discussion on github. This may be of interest to other software developers.

 

Saygılarımla / Best regards 

 

Rıdvan Melih ŞAHİN

Elektronik Tasarım Mühendisi / Electronic Design Engineer

 

 <http://dets.com.tr/> Yatay Logo Yazı2

DETS DİNAMİK MÜHENDİSLİK TEKNOLOJİ SİSTEMLERİ SAN. TİC. LTD.ŞTİ.

ODTÜ Teknokent Ostim Merkezi Ostim Mahallesi 1308 Cadde No:6 

Zemin kat No: 6  06374 - Yenimahalle/ ANKARA 

T: +90 312 473 16 73 M: +90 507 733 22 34

 <http://www.dets.com.tr/> www.dets.com.tr 

 

Bu e-posta ile iletilen bilgiler ve dosyalar gönderici ve alıcı arasında gizlidir. E-postanın ulaşması gereken kişi siz değilseniz, e-posta ve tüm eklerini silerek gönderen kişiye yanlışlıkla ilgili bilgi veriniz. Alıcı ve gönderici haricinde mesajın içerdiği bilgiler kullanılamaz. Bu mesajın içerdiği bilgiler DETS Dinamik Mühendislik Teknoloji Sistemleri San. Tic. Ltd. Şti.’yi hukuki sorumluluk altında bırakmaz. 


The information and files transmitted by this e-mail are confidential between the sender and the receiver. If you are not the person this e-mail needs to reach, delete the e-mail and all of its attachments and let the sender know this mistake. The information contained in the message can not be used except the sender and the receiver. The information contained in this message does not leave DETS Dinamik Mühendislik Teknoloji Sistemleri San. Tic. Ltd. under legal responsibility.

ABOSTM:
As you can see, SPI1 and SPI2 pins are already defined:
https://github.com/stm32duino/Arduino_Core_STM32/blob/5e72ff95a6fe75a8f8f112c0ac3a55318306a5ac/variants/STM32F1xx/F103C8T_F103CB(T-U)/PeripheralPins.c#L176

Arduino create a default SPI instance (instance of the class)
In variant_generic.h you can see that default **SPI** instance used is SPI1 with PA5, PA6, PA7
https://github.com/stm32duino/Arduino_Core_STM32/blob/5e72ff95a6fe75a8f8f112c0ac3a55318306a5ac/variants/STM32F1xx/F103C8T_F103CB(T-U)/variant_generic.h#L107

If you need both SPI, you just need to declare, in you sketch, a new instance with the pins of SPI2 (hardware instance)
For example, (check the pin match your board):
```
SPIClass SPI_2(PB_15, PB_14, PB_13);
SPI_2.begin();
```

Issue #731, opened 2019-10-24
Variant.h for RUMBA32


MagoKimbra:
Please enter the TIMER_SERIAL define in variant.h of rumba32.
If the SoftwareSerial.h library is included, as is the case with the TMCStepper library, SoftwareSerial takes the timer 7 for the STM32F446 processor which is the same one used by the servo and in fact no longer works.
I put #define TIMER_SERIAL 8 in variant.h so the servo works, but I don't know if the SoftwareSerial library works, it should be verified ...

fpistm:
Hi @MagoKimbra 
in this case you could provide a PR.
This is well explained here:
https://github.com/stm32duino/Arduino_Core_STM32/blob/2f1fd8e19d35a9805f81e5054a726d3b0b6b748a/libraries/SoftwareSerial/src/SoftwareSerial.cpp#L42-L51

So I guess there is no TIM18 for F446 and then TIM7 is used.

MagoKimbra:
Yes, TIM18 not define for STM32F466 and use the first TIM7.
I tried to put the define in my HAL firmware, but it is not seen by SoftwareSerial.cpp.

fpistm:
Don't know your HAL firmware but if set in variant.h this should be ok

MagoKimbra:
Yes yes in variant.h the servo works perfectly. I tried to put the define in my firmware which includes the TMCstepper.h library which in turn includes the Softwareserial.h library, but unfortunately the define is not seen, I said this.

fpistm:
ok.
Other way is to use build_opt.h or maybe the hat_conf_extra.h.

fpistm:
@MagoKimbra 
which timer you want for TIMER_Serial per default ?

MagoKimbra:
TIM9. Thank's.
I use TIM2 for stepper, SysTick for the rest. There are still TIM5 32bit, TIM3, TIM4, TIM10-TIM14 at 16bit for other things.

fpistm:
Well,  I will not modify the variant as this will fix it for all users and they do not want necessarily to use this one.
In your case you have to add `hal_conf_extra.h` with 
`#define TIMER_SERIAL TIM9`

MagoKimbra:
Where i add hal_conf_extra.h??? in Myfirmware or in library STM32???

fpistm:
When using Arduino IDE, simply adding it at sketch level works.
While the fle in the include path this will work.

PR #1092, opened 2020-06-08
[RUMBA32] Minor Improvements and Fixes

chrissbarr:
## [RUMBA32] Minor Improvements and Fixes

**Summary**

The fixes relating to timer conflicts are raised as a result of the investigation conducted here: https://github.com/Aus3D/RUMBA32/issues/26

This PR fixes/implements the following changes:

* [x] Update PWM pin mapping
  * [x] Remove timer conflicts (refer to https://github.com/Aus3D/RUMBA32/issues/26)
  * [x] Enable PWM on pins in EXP3 header (refer to https://github.com/Aus3D/RUMBA32/issues/8)
  * [x] **breaking change** - PWM is removed from some pins it was available on previously. However, the RUMBA32 board could not take advantage of PWM on those pins / no reason to use PWM on those pins. Further, some of these PWM pin timer mappings were in conflict with other PWM pins - so it is likely that they would not have worked correctly if actually used.
* [x] Set TIMER_SERIAL to TIM9 to remove conflict with TIMER_SERVO (refer to #731 and https://github.com/Aus3D/RUMBA32/issues/26)
  * [x] Check if TIMER pin mapping can be overridden by user
* [x] Enable ADC mapping on all pins which map to ADC1 (original A0 - A6 pin mapping is preserved and there should be no impact to existing software that uses those pin names)
* [x] Enable UART5 on EXP1 header
* [x] Tidy variant files and add peripheral options from current genpinmap file (a few additional UART, SPI etc. options are in the current file, added)

I am curious, does anyone know if there's a way to define the timers in the variant.h file such that they can be overridden by the user's code? For instance, I tried setting up the variant.h file like:

```
#ifndef TIMER_SERIAL
  #define TIMER_SERIAL           TIM9
#endif
```

But it seems that the variant.h file gets compiled before anything I do in the Arduino sketch / header, and I see a warning that TIMER_SERIAL is redefined. Is there a way to make this work? Would be quite useful if the TONE, SERVO and SERIAL timers could all be redefined by the user, while still allowing for a variant default.

**Validation**

* Ensure CI build is passed.

**Code formatting**

**Closing issues**


variants/RUMBA32_F446VE/variant.h
```diff
@@ -27,14 +27,14 @@ extern "C" {
...
-#define PA6   6  //D6
-#define PA7   7  //D7
+#define PA0   0  //D0   //A7
+#define PA1   1  //D1   //A8
+#define PA2   2  //D2   //A9
+#define PA3   3  //D3   //A0
+#define PA4   4  //D4   //A1
+#define PA5   5  //D5   //A10
+#define PA6   6  //D6   //A11
+#define PA7   7  //D7   //A12
```
fpistm, line 37: You have to use Ax value else if you use the analogRead API with PYn this will not work.
```suggestion
#define PA0   A7  //D0
#define PA1   A8  //D1
#define PA2   A9  //D2
#define PA3   A0  //D3
#define PA4   A1  //D4
#define PA5   A10  //D5 
#define PA6   A11  //D6
#define PA7   A12  //D7
```
same for all other...

fpistm:
> I am curious, does anyone know if there's a way to define the timers in the variant.h file such that they can be overridden by the user's code? For instance, I tried setting up the variant.h file like:
> 
> ```
> #ifndef TIMER_SERIAL
>   #define TIMER_SERIAL           TIM9
> #endif
> ```
> 
> But it seems that the variant.h file gets compiled before anything I do in the Arduino sketch / header, and I see a warning that TIMER_SERIAL is redefined. Is there a way to make this work? Would be quite useful if the TONE, SERVO and SERIAL timers could all be redefined by the user, while still allowing for a variant default.

Well, in this case you could redefine it at the sketch level, using the `build_opt.h` with:
`-DTIMER_SERIAL=TIM9`

https://github.com/stm32duino/wiki/wiki/Customize-build-options-using-build_opt.h

variants/RUMBA32_F446VE/variant.h
chrissbarr: Thanks @fpistm, not sure what I was thinking there... 

Fixed and tested AnalogRead using both Ax and PYn naming. Results look correct (same readings for either naming). Gave this a quick test and results are consistent either way:

ADCtest

chrissbarr:
> 
> Well, in this case you could redefine it at the sketch level, using the `build_opt.h` with:
> `-DTIMER_SERIAL=TIM9`
> 

Hmm, it might be worth surrounding each of the timers defined in variant.h with an `#ifndef` check then, so that they can be overridden in this way if desired. Are you aware of any disadvantages to doing this? If not, I think I'll add a commit doing this for serial/servo/tone timers.

fpistm:
> Hmm, it might be worth surrounding each of the timers defined in variant.h with an `#ifndef` check then, so that they can be overridden in this way if desired. Are you aware of any disadvantages to doing this? If not, I think I'll add a commit doing this for serial/servo/tone timers.

From my point of view there is no issue to add `#ifndef`.

chrissbarr:
Thanks @fpistm. I've noticed a bit of weirdness when using SoftwareSerial with certain pins, and have opened an issue about this (https://github.com/stm32duino/Arduino_Core_STM32/issues/1093). Hope you can take a look when you get the opportunity. 

I might hold off on requesting a merge on this until I understand what's happening there, in case it's something I've introduced or can fix in the pin definitions.

chrissbarr:
All right, I'm satisfied with these changes and am happy for the PR to be merged. I'll move it from draft now and it is yours to review @fpistm!

PR #1097, opened 2020-06-12
Use more efficient mode for STM32L4XX stop mode

angelnu:
**Summary**

* [X] Feature 1 - save power in stop mode for STM32L4XX devices
* [X] Feature 2 - Use MSI as wake up clock when available
* [ ] Breaking changes

**Feature 1**
Currently the low power library uses HAL_PWR_EnterSTOPMode for all devices. This does not leverage the stop2 mode which is available in the STM32L4XX. This results in 4,6 uA instead of 1,35 uA in my nucleo_l412kb.

Stop1 is only required when the UART device needs to wake up. So when not used the patch calls HAL_PWREx_EnterSTOP2Mode directly.

**Feature 2**
When available use the MSI as wake up clock. This avoids starting the HSI which generates a short spike of ca 300 uA until the SystemClock_ConfigFromStop is called.

**Validation**

* Ensure CI build is passed.

Not adding a new sketch since no external APIs are changed by this patch. 

**Code formatting**

**Closing issues**
NA

### DIFF
--- a/libraries/SrcWrapper/src/stm32/low_power.c
+++ b/libraries/SrcWrapper/src/stm32/low_power.c
@@ -233,8 +233,7 @@ void LowPower_stop(serial_t *obj)
 
   /* Enter Stop mode */
 #if defined(STM32L4xx) 
-  if (WakeUpUart == NULL)
-  {
+  if (WakeUpUart == NULL) {
     // STM32L4xx supports STOP2 mode which halves consumption
     HAL_PWREx_EnterSTOP2Mode(PWR_STOPENTRY_WFI);
   } else
--- a/libraries/SrcWrapper/src/stm32/low_power.c
+++ b/libraries/SrcWrapper/src/stm32/low_power.c
@@ -233,12 +233,14 @@ void LowPower_stop(serial_t *obj)
 
   /* Enter Stop mode */
 #if defined(STM32L4xx)
-  if (WakeUpUart == NULL) {
+  if ((WakeUpUart == NULL) || (WakeUpUart->Instance == (USART_TypeDef *)LPUART1_BASE)) {
     // STM32L4xx supports STOP2 mode which halves consumption
     HAL_PWREx_EnterSTOP2Mode(PWR_STOPENTRY_WFI);
   } else
 #endif
-    HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
+  {
+      HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
+  }
 
   /* Exit Stop mode reset clocks */
   SystemClock_ConfigFromStop();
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @angelnu 
as you know I'm currently working on the SubSeconds managements and I'm studying to move the low power driver to the Low Power library. Anyway, I have to study the impact of the mode you propose.

angelnu:
Jep, I know (looking forward to it) :-)

This patch is independent from the sleep time. I made it specific for the STM32L4 since this is the only device I have that supports STOP2. It would be nice if there would be a define for all platforms with STOP2...

I hope I can get the style check to agree so the patch is ready. If you have any other feedback you would like to get here please let me know.

libraries/SrcWrapper/src/stm32/low_power.c
```diff
@@ -223,12 +223,22 @@ void LowPower_stop(serial_t *obj)
...
+  __HAL_RCC_WAKEUPSTOP_CLK_CONFIG(RCC_STOP_WAKEUPCLOCK_MSI);
+#else
   __HAL_RCC_WAKEUPSTOP_CLK_CONFIG(RCC_STOP_WAKEUPCLOCK_HSI);
+#endif
 #endif
 
   /* Enter Stop mode */
-  HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
+#if defined(STM32L4xx)
+  if (WakeUpUart == NULL) {
```
fpistm, line 236: Here I think we could check if it is the LPUART because it is managed in STOP2 mode.

libraries/SrcWrapper/src/stm32/low_power.c
```diff
@@ -223,12 +223,22 @@ void LowPower_stop(serial_t *obj)
...
+    // STM32L4xx supports STOP2 mode which halves consumption
+    HAL_PWREx_EnterSTOP2Mode(PWR_STOPENTRY_WFI);
+  } else
+#endif
+    HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
```
fpistm, line 241: We could use the LPUART1 in STOP2 mode
```suggestion
  if ((WakeUpUart == NULL) || (WakeUpUart->Instance == (USART_TypeDef *)LPUART1_BASE)) {
    // STM32L4xx supports STOP2 mode which halves consumption
    HAL_PWREx_EnterSTOP2Mode(PWR_STOPENTRY_WFI);
  } else
#endif
  {
      HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
  }
```

libraries/SrcWrapper/src/stm32/low_power.c
```diff
@@ -223,12 +223,24 @@ void LowPower_stop(serial_t *obj)
...
   /* Enter Stop mode */
-  HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
+#if defined(STM32L4xx)
+  if ((WakeUpUart == NULL) || (WakeUpUart->Instance == (USART_TypeDef *)LPUART1_BASE)) {
+    // STM32L4xx supports STOP2 mode which halves consumption
+    HAL_PWREx_EnterSTOP2Mode(PWR_STOPENTRY_WFI);
+  } else
+#endif
+  {
+      HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
```
fpistm, line 242:
```suggestion
    HAL_PWR_EnterSTOPMode(PWR_LOWPOWERREGULATOR_ON, PWR_STOPENTRY_WFI);
```

PR #1184, opened 2020-09-22
Add requestFrom(uint8_t address, size_t quantity, bool sendStop)

### LINKED ISSUE #1183: TwoWire requestFrom(uint8_t address, size_t size, bool sendStop) Implementation
I am having a Platformio build error with a library [(ArduinoECCX08)](https://github.com/arduino-libraries/ArduinoECCX08).

This is because it calls:

```while (_wire->requestFrom((uint8_t)_address, (size_t)responseSize, (bool)true) != responseSize && retries--);```

Where it s casting the sendStop paramater as a boolean, the compiler fails saying that it is ambiguous. The core does not appear to have the Arduino specific function as found here --> [Link](https://github.com/esp8266/Arduino/blob/master/libraries/Wire/Wire.cpp#L123)

Implement a handler that better aligns with the arduino specfification --> [Docs](https://www.arduino.cc/en/Reference/WireRequestFrom)

Are there any guides I need to follow to implement a PR for this repo?

porkyneal:
This PR adds a missing function declaration that aligns further with the arduino spec as specified in issue #1183 

This is to fix compilation with an external Arduino library [ArduinoECCX08](https://github.com/arduino-libraries/ArduinoECCX08)

Fixes #1183

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -172,6 +172,11 @@ uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint8_t sendStop
   return requestFrom((uint8_t)address, (uint8_t)quantity, (uint32_t)0, (uint8_t)0, (uint8_t)sendStop);
 }
 
+uint8_t TwoWire::requestFrom(uint8_t address, size_t quantity, bool sendStop)
+{
+  return requestFrom((uint8_t)address, (uint8_t)quantity, (uint8_t)sendStop);
+}
+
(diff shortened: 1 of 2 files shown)


ABOSTM (approved):
LGTM.
After discussion with @fpistm, we conclude that it is not worth today to implement the code in 
`:requestFrom(uint8_t address, size_t quantity, bool sendStop)`
instead of
 `requestFrom((uint8_t)address, (uint8_t)quantity, (uint8_t)sendStop)`
like others like megaavr. It won't change neither the API nor the behavior.

PR #1187, opened 2020-09-23
[Enhancement] Add date retention support to stm32f1xx boards

josh-gaby:
**Add date retention through power cycles on stm32f1xx based boards**

On these chips, the date details are only stored in SRAM while running, because of this, dates are not being retained after a power off even with VBAT connected.

This is fixed by using backup registers DR2 and DR3 to store the date data each time it is updated by HAL_RTC_SetDate and whenever RTC_DateUpdate is called. Doing this means that the date can then extracted when the RTC is next initialized.


fpistm:
Hi @josh-gaby 
Thnks forthis PR.
Anyway modifying the HAL is not correct. The best way to integrate this is to add it to the https://github.com/stm32duino/STM32RTC
Please could you integrate this PR in the rtc.c? 
Thanks in advance

fpistm:
Replaced by https://github.com/stm32duino/STM32RTC/pull/41

PR #1203, opened 2020-10-11
adding missing i2c2 SCL pinmap for DISCO_F407G

owennewo:
**Summary**
DISCO_F407G has a pinmap for i2c2 SDA but the matching i2c2 SCL pinmap is commented. This means i2c2 is unusable as there is not SCL pinmap for i2c2.

Simply uncommenting the line fixes the issue.

**Validation**
I have this board and have tested that the change allows me to use i2c2 with an AS5600 i2c device.

**Closing issues**
None


fpistm (approved):
Thanks.
I guess it was commented because the pin is the mic clk.
Anyway it can be used for other stuff if mic is not used.

PR #1209, opened 2020-10-19
Update HardwareTimer.cpp

patricklaf:
Fix typo on TIM13_BASE in IRQ handler. Tested with Nucleo-F429ZI.


ABOSTM (approved):
@patricklaf Thanks for the fix 
LGTM

PR #1212, opened 2020-10-20
Timer OPM

patricklaf:
Add One Pulse Mode functions. Tested with Nucleo-F429ZI.

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -748,6 +748,33 @@ void HardwareTimer::setPreloadEnable(bool value)
 }
 
 /**
+  * @brief  Enable or disable one pulse mode for timer operation.
+  *         When enabled, changes to the overflow value take effect
+  *         immediately. When enabled (the default), the value takes
+  *         effect only at the next update event (typically the next
+  *         overflow).
+  * @param  value: true to enable one pulse mode, false to disable
+  * @retval None
+  */
+void HardwareTimer::setOnePulseMode(bool value)
+{
+	if (value) {
+		LL_TIM_SetOnePulseMode(_timerObj.handle.Instance, LL_TIM_ONEPULSEMODE_SINGLE);
+	} else {
+		LL_TIM_SetOnePulseMode(_timerObj.handle.Instance, LL_TIM_ONEPULSEMODE_REPETITIVE);
+	}
+}
+
+/**
+  * @brief  Retrieves channel one pulse mode
+  * @retval returns true if one pulse mode is set 
+  */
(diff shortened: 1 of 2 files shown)


ABOSTM:
Hi @patricklaf,
It took me some time to review this PR because, even if it seems simple,
and just do what it is supposed to do (setting ab bit in a register),
it doesn't generate output signal as one can expect:
* With the all in one API `setPWM()`: there is only a rising edge and output signal remains high
* With a full configuration `setMode() `+ `setOverflow() `+` setCaptureCompare()` ...
	- with `TIMER_OUTPUT_COMPARE_PWM1`
		--> same as setPWM() which is normal as it used PWM1
	- with `TIMER_OUTPUT_COMPARE_PWM2`
		--> there is effectively a pulse, but the duration is not the one expected

The reason is because this feature is usually used with trigger (not supported today in arduino)
and specially with PWM2 in order to generate a pulse in response to a stimulus and after a delay.
Also because we enabled preload by default (of both ARR and CCRx registers) .
And also because STM32Cube HAL API used within HardwareTimer, configures preload and generates Update event on its own.

I study the solution to disable preload, but it is complex and it doesn't work properly, 
because STM32Cube HAL API within HardwareTimer, configures preload on its own.
A second option is to call `refresh()` at application level just before resume(), to force update generation.
It means it cannot be use with `setPWM()` as `resume(`) is included in `setPWM().`
Also the generated signal for PWM1 is not exactly as expected: there is a pulse, but at the end of timer count, there is a rising edge and signal remains high.
```
    _     ______________________
___| |___|
```
Only PWM2 generates a proper pulse.
So, because it relys on user to insertion of `refresh()` in the application and because only 1 case will work fine, I don't consider this as a proper solution.

From my point of view, best approach would be to create an all in one API that configures the hardware timer (disable preload or generate the Update Event, force PWM2, ...) and starts the onepulse.
And I would suggest parameters: 
* delay (delay before the pulse if any). Corresponding to CCRx in PWM2. Optional (if not set by user, 0 delay)
* duration of the pulse:  corresponding to time between CCRx and ARR.
restricted to `TICK_COMPARE_FORMAT `and `MICROSEC_COMPARE_FORMAT`
because other formats `HERTZ_COMPARE_FORMAT `and  `PERCENT_COMPARE_FORMAT `don't really make sense for a single pulse.
* channel: timer channel

Note: STM32Cube dedicated API: `HAL_TIM_OnePulse_Init() `cannot be used for such implementation, because it requires the use of a trigger (not yet supported in arduino).

ABOSTM (changes requested):
See my previous comment

fpistm:
No feedback from OP since a while so close it and considered as abandoned.

PR #1215, opened 2020-10-20
Add HSI clock support for STM32F103RX

sanyi:
Add HSI clock support for STM32F103RX

`build_flags = -D USE_HSI_CLOCK`


sanyi:
Apparently need styling fixes.

fpistm:
Will be added later when #1091 will be done

fpistm:
Hi @sanyi 
I'm currently working on #1091 
All clock config for generic will be aligned to try to use HSE, HSE bypass then HSI.
Like it is done for STM32F103Vx:

https://github.com/stm32duino/Arduino_Core_STM32/blob/caa8a28b281d14f91fbdda3744160fb5d40ca419/variants/Generic_F103Vx/variant.cpp#L210-L231

like this no need to add a specific define.
And if really the clock config doesn't fir the user need it should be redefined at sketch level.
So I close this PR as I will not merge it.

sanyi:
@fpistm I see. Thank you for the heads up. Looking forward to use/test the solution proposed by you. Cheers.

Issue #488, opened 2019-03-29
could'nt recieve integers at master from slave


DeadSHoT555:
well in 'Wire examples' for (stm32f10c bluepill) the character array works pretty good..., but when slave sender modified to send integer there is no serial print in master reader ... gets struck at some unknown point...,  works fine for one byte transfer....,and not any more for two bytes or more.. can anyone fix this bug???

fpistm:
@DeadSHoT555  could you provide more information.
Host OS, IDE version, core version, exact sketches you used, what is expected...
There an issue template for that, so pleas follow it else I will not be able to provide support.
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

DeadSHoT555:
OS: [ Windows10]
Arduino IDE version: [1.8.9]
STM32 core version: [1.5.0]
Tools menu settings [Newlib nano(default)]
Upload method: [serial]
Board (Generic STM32F1 Series):
Name: [BluePill F103C8]

DeadSHoT555:
// Wire Master Reader
// by Nicholas Zambetti <http://www.zambetti.com>

// Demonstrates use of the Wire library
// Reads data from an I2C/TWI slave device
// Refer to the "Wire Slave Sender" example for use with this

// Created 29 March 2006

// This example code is in the public domain.


#include <Wire.h>

void setup()
{
  Wire.begin();        // join i2c bus (address optional for master)
  Serial.begin(9600);  // start serial for output
}

void loop()
{
  Wire.requestFrom(8, 2);    // request 6 bytes from slave device #2

//  while(Wire.available())    // slave may send less than requested
//  { 
    byte c = Wire.read(); // receive a byte as character
    byte d = Wire.read(); // receive a byte as character
    Serial.print(c);         // print the character
//  }

  delay(500);
}

### master_sender

DeadSHoT555:
// Wire Slave Sender
// by Nicholas Zambetti <http://www.zambetti.com>

// Demonstrates use of the Wire library
// Sends data as an I2C/TWI slave device
// Refer to the "Wire Master Reader" example for use with this

// Created 29 March 2006

// This example code is in the public domain.


#include <Wire.h>

void setup() {
  Wire.begin(8);                // join i2c bus with address #8
  Wire.onRequest(requestEvent); // register event
}

void loop() {
  delay(100);
}

// function that executes whenever data is requested by master
[...]

PR #1221, opened 2020-10-29
Wire fixes for repeated start(slave) and multiple .write in onRequestEvent()

DanielLiebler:
**Summary**

* [x] when in slave mode and we are receiving a write/read combo with a repeated start(so no stop condition in between), the write part doesn't trigger an onReceiveEvent()
* [x] when doing multiple .write() calls in onRequestEvent(), only the last one is used. (as outlined in #488 )

**Description**

* The slave implementation doesn't seem to work with repeated starts, but this is also a quick fix by sending out an onReceiveEvent() when getting the new address callback.

* In #488 it was mentioned to just use one write, but I think this is not good behavior, since it's not mentioned in the documentation. Nevertheless, it seems to be a quick fix, so I just created this PR.

**How it works**

* When getting the address callback, the data which has been cached is sent to an onReceiveEvent, when still in SLAVE_MODE_RECEIVE mode.

* We are resetting i2cTxRxBufferSize before calling i2c_onSlaveTransmit() in HAL_I2C_AddrCallback()
and modify the function i2c_slave_write_IT() to start writing in the i2cTxRxBuffer with the size of the buffer as offset and add up the sizes.

### DIFF
--- a/libraries/Wire/src/utility/twi.c
+++ b/libraries/Wire/src/utility/twi.c
@@ -846,10 +846,10 @@ i2c_status_e i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint16_t size)
   } else {
     // Check the communication status
     for (i = 0; i < size; i++) {
-      obj->i2cTxRxBuffer[i] = *(data + i);
+      obj->i2cTxRxBuffer[obj->i2cTxRxBufferSize + i] = *(data + i);
     }
 
-    obj->i2cTxRxBufferSize = size;
+    obj->i2cTxRxBufferSize += size;
   }
   return ret;
 }
@@ -986,6 +986,7 @@ void HAL_I2C_AddrCallback(I2C_HandleTypeDef *hi2c, uint8_t TransferDirection, ui
       obj->slaveMode = SLAVE_MODE_TRANSMIT;
 
       if (obj->i2c_onSlaveTransmit != NULL) {
+        obj->i2cTxRxBufferSize = 0;
         obj->i2c_onSlaveTransmit(obj);
       }
 #if defined(STM32F0xx) || defined(STM32F1xx) || defined(STM32F2xx) || defined(STM32F3xx) ||\
--- a/libraries/Wire/src/utility/twi.c
+++ b/libraries/Wire/src/utility/twi.c
@@ -980,6 +980,11 @@ void i2c_attachSlaveTxEvent(i2c_t *obj, void (*function)(i2c_t *))
 void HAL_I2C_AddrCallback(I2C_HandleTypeDef *hi2c, uint8_t TransferDirection, uint16_t AddrMatchCode)
 {
   i2c_t *obj = get_i2c_obj(hi2c);
+  if ((obj->slaveMode == SLAVE_MODE_RECEIVE) && (obj->slaveRxNbData != 0)) {
+    obj->i2c_onSlaveReceive(obj);
+    obj->slaveMode = SLAVE_MODE_LISTEN;
+    obj->slaveRxNbData = 0;
+  }
 
   if (AddrMatchCode == hi2c->Init.OwnAddress1) {
     if (TransferDirection == I2C_DIRECTION_RECEIVE) {


libraries/Wire/src/utility/twi.c
```diff
@@ -846,9 +846,10 @@ i2c_status_e i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint16_t size)
   } else {
     // Check the communication status
     for (i = 0; i < size; i++) {
-      obj->i2cTxRxBuffer[i] = *(data + i);
+      obj->i2cTxRxBuffer[obj->i2cTxRxBufferSize + i] = *(data + i);
     }
 
+    size = size + obj->i2cTxRxBufferSize;
```
ABOSTM, line 852: Nitpick: Both lines could be simplified to 
`obj->i2cTxRxBufferSize += size;`

ABOSTM (commented):
Hi @DanielLiebler,
Thanks for your Pull Request.
After review, it looks good to me,
but I would like to test, specially because there are 2 major I2C Hardware implementation within STM32 families.
Can you tell me what you tested ? which boards ...
Can you tell me how you tested ? Which master with repeated start?
Can you share you sketch () ?

libraries/Wire/src/utility/twi.c
DanielLiebler: Hi @ABOSTM ,
thank you for the quick response.
To describe my main Setup:
I have 1 Raspberry Pi <- I2C Master --- I2C Slave -> 1 STM Nucleo F746ZG(stm32duino) <- I2C Master --- I2C Slave -> 4 STM Nucleo F746ZG(STMCube with MotorControl)
So the STM running the stm32duino is working with 2 I2C Instances. One as a master, one as a slave.
For the first tests, I stopped all communication and working with the 4 other Nucleos and only had the RPI<-> stm32duino communication running. I also tested with an Arduino Mega 2560 as the Master to the stm32duino to be able to control if the stop condition is sent(i couldn't do it easily with the rpi).

How I tested:
I had an Oscilloscope attached to look at the lines, and for the issue with repeated start, I could see that the transmission was there and acknowledged, but it didn't arrive in my Arduino code. (I think the Issue here is that HAL_I2C_ListenCpltCallback() isn't called when there is no stop condition).
For the issue with the write, it looked like the Nucleo was sending only 1 byte of data back and then held SCL low.

So the master which sends the repeated start is the RPI or the Arduino Mega.

Sadly, I don't have any other boards to test the functionality on other boards of the STM family.

My codebase has been getting quite big, so I don't know if you would want to have a look at it, but I will try to get a smaller example of the issue.
What would be the best way of sharing it?

DanielLiebler: **How to use it**
The useRestartCondition define in the Master code is for the issue with the restart condition,
For the other issue, in Slave there is respondMultiWrite.

To read data, I send via serial to the master as ASCII: `8,0,1,0\0A`
For the restart condition issue: if it works, you get on the Slave as Serial output:
received adress 0x1
transmitted data 0x5555

if it didn't work, the received adress line is missing.

For the other issue, if it didn't work, it will lock up the I2C bus. So a reboot of both devices is necessary afterward.

I didnt rewrite the Master code, so forgive me that its a little bit messy
**Master (Arduino Mega 2560)**
```#include <Arduino.h>
#include <Wire.h>

#define useRestartCondition true
#define WIRE_INTERFACE_HANDLE Wire
uint32_t noResponseCount = 0;
uint32_t writeFailedCount = 0;
uint32_t regAddrFailedCount = 0;
uint32_t errCount = 0;
uint32_t transactionCount = 0;
uint32_t lastPrint = 0;

uint8_t lastReadNum = 0;

void setup() {
  // initialize I2C
  WIRE_INTERFACE_HANDLE.begin();

  Serial.begin(115200);
  Serial.println("hello");
}

void sendReg(uint8_t address, uint8_t reg, uint16_t data) {
  uint8_t success;
  do {
    writeFailedCount++;
    // Start a Transmission to device with address 64
    WIRE_INTERFACE_HANDLE.beginTransmission(address);
    // Write reg&data to the device
    WIRE_INTERFACE_HANDLE.write(reg);
    WIRE_INTERFACE_HANDLE.write(data & 0xff);
    WIRE_INTERFACE_HANDLE.write((data >> 8) & 0xff);
    success = WIRE_INTERFACE_HANDLE.endTransmission();
  } while (success != 0);
  //subtract the one time which succeeded
  writeFailedCount--;
}
uint8_t readReg(uint8_t address, uint8_t reg, uint16_t *data) {
  uint8_t success;
  do {
    regAddrFailedCount++;
    // Start a Transmission to device with address 64
    WIRE_INTERFACE_HANDLE.beginTransmission(address);
    // Write reg&data to the device
    WIRE_INTERFACE_HANDLE.write(reg);
    success = WIRE_INTERFACE_HANDLE.endTransmission(!useRestartCondition);
  } while (success != 0);
  //subtract the one time which succeeded
  regAddrFailedCount--;
  
  lastReadNum = WIRE_INTERFACE_HANDLE.requestFrom(address, (uint8_t) 2);    // request 2 bytes from slave device #64

  if (lastReadNum != 2) {
    Serial.print("DATALEN:");
    Serial.println(lastReadNum);

    while(WIRE_INTERFACE_HANDLE.available()) {
      int c = Wire.read(); // receive a byte as character
      Serial.print(c);         // print the character
      Serial.print(',');         // print the character
    }

    return 1;
  }

  //receiving
  if (WIRE_INTERFACE_HANDLE.available() >= 2) {
    uint16_t r1 = WIRE_INTERFACE_HANDLE.read();
    uint16_t r2 = WIRE_INTERFACE_HANDLE.read();
    *data = ((r2 << 8) | r1);
    return 0;
  }
  return 2;
}

void loop() {
  while (Serial.available() > 0) {
    int addr = Serial.parseInt();
    int rw = Serial.parseInt();
    int reg = Serial.parseInt();
    int data = Serial.parseInt();
    if (Serial.read() == '\n') {
      if ((rw == 0 || rw == 1) && reg >= 0 && reg <= 255 && data >= 0 && (uint32_t)data <= 0xffffffff) { // (addr == 0x8) && 
        if (rw) {
          // write
          Serial.print("Sending:");
          sendReg(addr, reg, (uint16_t) data);
          Serial.print(addr, HEX);
          Serial.print("/");
          Serial.print(reg, HEX);
          Serial.print("/");
          Serial.println(data, HEX);
        } else {
          Serial.print("Receiving:");
          uint8_t err = readReg(addr, reg, (uint16_t*) &data);
          Serial.print(addr, HEX);
          Serial.print("/");
          Serial.print(reg, HEX);
          Serial.print("/");
          Serial.print(data, HEX);
          Serial.print("/");
          Serial.print(data);
          Serial.print("/");
          Serial.println(err, HEX);
        }
      } else {
        Serial.print("wrong param");
        Serial.print(addr, HEX);
         Serial.print("/");
        Serial.print(rw, HEX);
        Serial.print("/");
        Serial.print(reg, HEX);
        Serial.print("/");
        Serial.println(data, HEX);
      }
    }
  }
}```
**Slave (stm32duino)**
```#include <Arduino.h>
#include <Wire.h>

#define respondMultiWrite false

// Use Wire2, because Wire is used for other communication
//            SDA  SCL
TwoWire Wire2(PB11, PB10);

void i2cReceiveEvent(int numBytes);
void i2cRequestEvent();

volatile uint8_t i2cCommand; // register adress

void setup() {
  Serial.begin(115200);
  
  Wire2.begin(0x08, true); // boolean is general call enable/disable
  Wire2.onReceive(i2cReceiveEvent);  // register event
  Wire2.onRequest(i2cRequestEvent);  // register event
  Serial.println("Started arduino firmware!");
}

void loop() {
}

// function that executes whenever data is received from master
// this function is registered as an event, see setup()
void i2cReceiveEvent(int numBytes) { 
  // there are two cases:
  // 1. user sends a 16-bit write command, then the transmission is 1+2 bytes long
  // 2. user sends a 16-bit read command, then the transmission is 1 byte long. After that a 2 byte request is received.

  // case 1
  if ((3 == numBytes) && (Wire2.available() > 0)) {
    i2cCommand = Wire2.read();  // First byte
    uint8_t b = Wire2.read();   // Second byte
    uint8_t c = Wire2.read();   // Third byte

    // Get the I2C Value
    uint16_t i2cVal = ((c << 8) & 0xFF00) | (b & 0xFF);

    Serial.print("recevied adress 0x");
    Serial.print(i2cCommand, HEX);
    Serial.print(" with data 0x");
    Serial.println(i2cVal, HEX);
  }

  // case 2
  // User issues i2c 16-bit register read command
  if ((1 == numBytes) && (Wire2.available() > 0)) {
    i2cCommand = Wire2.read();
    Serial.print("received adress 0x");
    Serial.println(i2cCommand, HEX);
  }
}

// function that executes whenever data is requested by master
// this function is registered as an event, see setup()
void i2cRequestEvent() {

  if (respondMultiWrite) {
    // The way i would like it to also work
    int16_t regVal = 0x5555;
    //Wire2.write(regVal & 0xFF);         // Send data back to the Master
    //Wire2.write((regVal >> 8) & 0xFF);  // Send data back to the Master
  } else {
    // The old way to write data
    const uint8_t data[2] = {0x55, 0x55}; 
    const uint8_t* dataPtr = (const uint8_t*) &data;
    Wire2.write(dataPtr, (size_t) 2);
  }  

  Serial.print("transmitted data 0x");
  Serial.println(regVal, HEX);
}```

**Comment by ABOSTM** [comment]
Thank I will test that.

**Comment by fpistm** [comment]
Thanks @DanielLiebler for this PR.
Could you squash your last commit with the first one please.

Note: The CI core compilation issue is normal as I'm updating some dependencies and will be back to normal soon.

**Comment by DanielLiebler** [comment]
@fpistm should be done. I didnt know you could do such things, even though I guess you can mess up things really bad with squashing^^

**Comment by fpistm** [comment]
Thanks @DanielLiebler 
I've finished to update dependencies anyway to have the build OK, could you rebase on top of the master. Thanks in advance.

File: libraries/Wire/src/utility/twi.c [hunk 2]
```diff
@@ -846,9 +846,10 @@ i2c_status_e i2c_slave_write_IT(i2c_t *obj, uint8_t *data, uint16_t size)
   } else {
     // Check the communication status
     for (i = 0; i < size; i++) {
-      obj->i2cTxRxBuffer[i] = *(data + i);
+      obj->i2cTxRxBuffer[obj->i2cTxRxBufferSize + i] = *(data + i);
     }
 
+    size += obj->i2cTxRxBufferSize;
     obj->i2cTxRxBufferSize = size;
```
ABOSTM, line 853: What I meant was to merge both lines, as there is no need to update `size`:
```
obj->i2cTxRxBufferSize += size;
```

ABOSTM (commented):
LGTM
Tested on Nucleo F429ZI (I2C V1) and Nucleo F746ZG (I2C V2)

libraries/Wire/src/utility/twi.c
fpistm: Will be ready for merge when fixed in the first commit.

DanielLiebler: now I understood what you meant!

fpistm (commented):
LGTM
Waiting @ABOSTM approval

ABOSTM (approved):
LGTM

PR #1227, opened 2020-11-02
HW UART - Remove repetitive NVIC SetPriority calls

AbsoluteCatalyst:
Hi,
Is there any purpose behind the NVIC UART IRQ priority being repetitively set within uart_attach_tx_callback and uart_attach_rx_callback rather than only at initialization? 

Unfortunately this prevents any runtime/external changes done to the IRQ priority since it'd constantly be reset.

If there's no problem having it this way: 
I've quickly removed the two extra setPriority calls and replaced them with a single call at init().


ABOSTM (approved):
LGTM

AbsoluteCatalyst:
Thanks!

fpistm:
Welcome @AbsoluteCatalyst 
Thanks for the PR too 😉

PR #1238, opened 2020-11-14
Support for STM32F303VC Discovery

dnandha:
**Summary**
Adds support for STM32F303VC (STM32F3) Discovery:  https://www.st.com/en/evaluation-tools/stm32f3discovery.html.
Variant `DISCO_F303VC` is added using the wiki guide. Board/Variant is similar to `DISCO_F407VG` but comes with decreased specs and slightly different pin-out.

Blink program for the boards wonderful 8 LED's successfully compiles / runs:
```
/*
  Blink

  Turns an LED on for one second, then off for one second, repeatedly.

  http://www.arduino.cc/en/Tutorial/Blink
*/

int LED_SEQ[8] = {LED_BLUE, LED_BLUE2, LED_RED, LED_RED2,
                  LED_ORANGE, LED_ORANGE2, LED_GREEN, LED_GREEN2};

// the setup function runs once when you press reset or power the board
void setup() {
  // initialize digital led pins as an output.
  for (int i = 0; i < 8; i = i + 1) {
    pinMode(LED_SEQ[i], OUTPUT);
  }
}

// the loop function runs over and over again forever
void loop() {
  for (int i = 0; i < 8; i = i + 1) {
    digitalWrite(LED_SEQ[i], HIGH);   // turn the LED on (HIGH is the voltage level)
    delay(500);                       // wait for a second
  }
  for (int i = 0; i < 8; i = i + 1) {
    digitalWrite(LED_SEQ[i], LOW);    // turn the LED off (LOW is the voltage level)
    delay(500);                       // wait for a second
  }
}
```


fpistm:
Will be added later when #1091 will be done

dnandha:
@fpistm Ok, when is that? #1091 has been open for a long time. Also adding all open PR's **after** #1091 means that all PR's have to be updated to the new structure. I'd suggest adding all new variants now and update #1091 once it's ready.

fpistm:
Hi @dnandha 
I've included your PR in #1091, it has been reformatted to the new variant definition way and made some fixes.
So I close this one as it is superseded by #1091.
Thanks again for your contribution.

PR #1258, opened 2020-12-04
Add clear pending interrupt feature with new clearPendingInterrupt API

cparata:
Hi Frederic,
I reopened this PR referring on the new branch that I created on my fork to test it in a clean way.
Best Regards,
Carlo

### DIFF
--- a/cores/arduino/WInterrupts.cpp
+++ b/cores/arduino/WInterrupts.cpp
@@ -87,13 +87,3 @@ void detachInterrupt(uint32_t pin)
   UNUSED(pin);
 #endif
 }
-
-void clearPendingInterrupt(uint32_t pin)
-{
-#if !defined(HAL_EXTI_MODULE_DISABLED)
-  PinName p = digitalPinToPinName(pin);
-  stm32_interrupt_clear_pending(STM_GPIO_PIN(p));
-#else
-  UNUSED(pin);
-#endif
-}
--- a/cores/arduino/WInterrupts.cpp
+++ b/cores/arduino/WInterrupts.cpp
@@ -87,3 +87,13 @@ void detachInterrupt(uint32_t pin)
   UNUSED(pin);
 #endif
 }
+
+void clearPendingInterrupt(uint32_t pin)
+{
+#if !defined(HAL_EXTI_MODULE_DISABLED)
+  PinName p = digitalPinToPinName(pin);
+  stm32_interrupt_clear_pending(STM_GPIO_PIN(p));
+#else
+  UNUSED(pin);
+#endif
+}
--- a/cores/arduino/WInterrupts.h
+++ b/cores/arduino/WInterrupts.h
@@ -33,6 +33,4 @@ void attachInterrupt(uint32_t pin, void (*callback)(void), uint32_t mode);
 
 void detachInterrupt(uint32_t pin);
 
-void clearPendingInterrupt(uint32_t pin);
-
 #endif /* _WIRING_INTERRUPTS_ */
--- a/cores/arduino/WInterrupts.h
+++ b/cores/arduino/WInterrupts.h
@@ -33,4 +33,6 @@ void attachInterrupt(uint32_t pin, void (*callback)(void), uint32_t mode);
 
 void detachInterrupt(uint32_t pin);
 
+void clearPendingInterrupt(uint32_t pin);
+
 #endif /* _WIRING_INTERRUPTS_ */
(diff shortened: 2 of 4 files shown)


fpistm:
For ref #1168

ABOSTM:
Hi @cparata,
I had in mind (according to #1168 discussion) that we go toward a symmetric approach, meaning that 
in `stm32_interrupt_disable()` we should disable EXTI too (in addition to NVIC).
And instead of calling HAL_GPIO_DeInit()/HAL_GPIO_Init() to keep GPIO mode,
I suggested to use LL EXTI API (even if EXTI API was not directly used in stm32_interrupt_enable(), it is used indirectly in HAL_GPIO_Init(), thus no symmetric break).

cparata:
Hi @ABOSTM ,
in fact, this is what I would like to test. I will update the PR asap I have the opportunity to implement the stm32_interrupt_disable() with the LL HAL.
Best Regards,
Carlo

cparata:
My only concern about the usage of LL HAL is portability. It is not said that LL HAL is available for all STM32 families; this is the reason why I suggested the usage of the HAL_DeInit that instead should be available for all STM32 families.

cparata:
Hi all,
I remanaged the PR, modifying only the "stm32_interrupt_disable". I implemented it using the gpio DeInit because I guess it is the simplest way to do it and it is also the most portable. Let me know if you have any comment. Unfortunately, I cannot test it with motor driver until Wednesday, when I will be in the office.
Best Regards,
Carlo

cparata:
I was able to test the new PR with the motor driver and it works as expected on my side. So, for me this implementation is functional.

ABOSTM (commented):
Hi Carlo,
Thanks for this PR,
I feel like this is quite heavy to simply disable exti.
I wonder whether we could make that simpler:
* either with LL, using directly LL_EXTI_DisableIT_0_31()
* or with HAL, using HAL_EXTI_GetConfigLine(), changing EXTI mode to EXTI_MODE_NONE and HAL_EXTI_SetConfigLine()
Note: this way we keep pin configuration (no resore of default input mode) and simply disable exti.
What do you think ?

cparata:
Hi @ABOSTM ,
I'm not really expert on LL APIs. Unfortunately, at the moment I do not have band to investigate better on this topic. Anyway, from the testing that I did, it is sure that if we are able to reset in some way the EXTI settings during the stm32_interrupt_disable(), the issue that I see on motor driver will be solved. So, if you are able to find a more efficient way to do that, it is fine for me as well.
Best Regards,
Carlo

battosai30:
Hi guys,

I was meeting troubles with a STM32L072 just trying to use interrupt with a button generating some rebonds and I found this PR. I applied modifications but it's not enough for me. I had to add __HAL_GPIO_EXTI_CLEAR_IT() in order to get the expected behaviour.

My (working) code : 

```
const int butPpin = PA1;
volatile byte flag = 0;
volatile byte flag2 = 0;
volatile int count = 0;


void setup() {

  Serial.begin(115200);
  while (!Serial);
  delay(500);
  pinMode(butPpin, INPUT_PULLUP);

  attachInterrupt(digitalPinToInterrupt(butPpin), intInt, FALLING);

}

void loop() {

  if (flag) {

    flag = 0;
    Serial.println("ping ! ");
    Serial.println(count);
    delay(1000);

    attachInterrupt(digitalPinToInterrupt(butPpin), intInt, FALLING);

  }

}


void intInt()
{

  detachInterrupt(digitalPinToInterrupt(butPpin));

  __HAL_GPIO_EXTI_CLEAR_IT(butPpin);
  flag = 1;
  count++;
}
```

If __HAL_GPIO_EXTI_CLEAR_IT(butPpin); is commented, I get an extra interrupt

cparata:
Hi @battosai30 ,
actually for debouncing it is not needed to use this API. You can try something like this:

```
const int butPpin = PA1;
volatile byte flag = 0;
volatile int count = 0;
int PushButtonState = 0;


void setup() {

  Serial.begin(115200);
  while (!Serial);
  delay(500);
  pinMode(butPpin, INPUT_PULLUP);

  /* Check what is the Push Button State when the button is not pressed. It can change across families */
  PushButtonState = (digitalRead(butPpin)) ?  0 : 1;

  attachInterrupt(digitalPinToInterrupt(butPpin), intInt, FALLING);

}

void loop() {

  if (flag) {
    /* Debouncing */
    delay(50);

    /* Wait until the button is released */
    while ((digitalRead( butPpin) == PushButtonState));

    /* Debouncing */
    delay(50);

    /* Reset Interrupt flag */
    flag = 0;
    Serial.println("ping ! ");
    Serial.println(count);
  }

}


void intInt()
{
  count++;
  flag = 1;
}
```

Let me know if it works on your side.
Best Regards,
Carlo

battosai30:
Thanks for your reply :)

Yeah it could be a solution but as I'm designing a very low power device, spend 100ms just for debounce is a pretty big waste ... And  it's not the behaviour you usually get in Arduino environnement so ...

PR #1269, opened 2020-12-22
SoftwareSerial stopListening at the begining

FRASTM:
When begining the transaction, in the SoftwareSerial::begin
, the active_listener is not yet assigned, but null.
Then it differs from _this_ but should not act as if it was listening.

This is the case when testing on nucleo_l552ze board.

Signed-off-by: Francois Ramu <francois.ramu@st.com>

### DIFF
--- a/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -148,7 +148,9 @@ bool SoftwareSerial::listen()
   if (active_listener != this) {
     // wait for any transmit to complete as we may change speed
     while (active_out);
-    active_listener->stopListening();
+    if (active_listener != nullptr) {
+      active_listener->stopListening();
+    }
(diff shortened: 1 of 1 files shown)


libraries/SoftwareSerial/src/SoftwareSerial.cpp
```diff
@@ -148,7 +148,9 @@ bool SoftwareSerial::listen()
   if (active_listener != this) {
     // wait for any transmit to complete as we may change speed
     while (active_out);
-    active_listener->stopListening();
+    if (active_listener != 0) {
```
fpistm, line 151:
```suggestion
    if (active_listener != nullptr) {
```

fpistm (changes requested):

libraries/SoftwareSerial/src/SoftwareSerial.cpp
FRASTM: changed

fpistm (approved):

ABOSTM (approved):
LGTM

PR #1287, opened 2021-01-28
fix: wrong timer source for frequency calculation with STM32G0xx controllers

ahessling:
This PR fixes the wrong returned timer source in STM32G0xx MCUs. The G0xx controllers also have only PCLK1 (like the F0 familiy), so the function should always return 1 in this case. Otherwise, e.g. if TIM1 in combination with PWM is used, 2 is returned which will crash the function HardwareTimer::getTimerClkFreq because of an unexpected timer source.


ABOSTM (approved):
LGTM

fpistm:
Thanks @ahessling for the fix.
Thanks @ABOSTM for review.

PR #1290, opened 2021-02-02
Use Semaphore in this variant to control M0/M4

FRASTM:
To prepare the stm32duinoBLE running on the stm32wb55 to control the clock of the M4
Especially the CFG_HW_RCC_SEMID (3) is used when enabling the BLE and preventing the M0 to access the RCC.

This is required for the https://github.com/stm32duino/STM32duinoBLE/pull/25

Signed-off-by: Francois Ramu <francois.ramu@st.com>

PR #1291, opened 2021-02-02
WB enhancements

fpistm:
- [x] Remove OTP API from the variant and provide it as a bsp.
- [x] Enable HSEM HAL module by default
- [x] Move HSE Capacitor Tuning api to clock api file
- [x] Add required semaphore managements

Supersede #1290


ABOSTM (commented):
LGTM, minor question ?

libraries/SrcWrapper/src/stm32/clock.c
```diff
@@ -160,6 +162,18 @@ void enableClock(sourceClock_t source)
   }
 }
 
+void configHSECapacitorTuning(void)
+{
+#if defined(OTP_AREA_BASE) && defined(STM32WBxx)
```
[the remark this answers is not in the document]
fpistm: I asked myself the same question. So I've checked WL serie and saw it have it that's why I added this guard. :wink:

fpistm:
@ABOSTM 
note that the PR is no ended.

FRASTM (approved):

cores/arduino/stm32/lock_resource.h
```diff
@@ -35,15 +30,134 @@ typedef enum {
...
+#define CFG_HW_CLK48_CONFIG_SEMID                               0U
+#define CFG_HW_RCC_CRRCR_CCIPR_SEMID                            0U
+#define CFG_HW_ENTRY_STOP_MODE_SEMID                            0U
+#define CFG_HW_RCC_SEMID                                        0U
+#define CFG_HW_FLASH_SEMID                                      0U
+#define CFG_HW_PKA_SEMID                                        0U
+#define CFG_HW_RNG_SEMID                                        0U
+#define CFG_HW_GPIO_SEMID                                       0U
+#define CFG_HW_EXTI_SEMID                                       0U
+#endif /* STM32WBxx || STM32H7xx */
```
ABOSTM, line 118: comment should match #if

cores/arduino/stm32/lock_resource.h
```diff
@@ -35,15 +30,134 @@ typedef enum {
 } LockResource_Status_t;
 
 /* Exported constants --------------------------------------------------------*/
-#define LOCK_RESOURCE_TIMEOUT   100U /* timeout in ms */
+#if defined(STM32WBxx) // || defined(STM32H7xx)
```
ABOSTM, line 33: remove comment 
`// || defined(STM32H7xx)`

variants/PNUCLEO_WB55RG/variant.cpp
```diff
@@ -99,9 +100,15 @@ WEAK void SystemClock_Config(void)
   RCC_ClkInitTypeDef RCC_ClkInitStruct = {};
   RCC_PeriphCLKInitTypeDef PeriphClkInitStruct = {};
 
+  /* This prevents the CPU2 (M0+) to configure RCC */
```
ABOSTM, line 103: It does not prevent M0+ to configure RCC, it prevents M0+ to configure RCC at the same time. It can still happen later, after semaphore release.

ABOSTM (changes requested):

cores/arduino/stm32/lock_resource.h
fpistm: OK. originally I add it for future use in case of H7 dual core support introduction :wink:
Done

cores/arduino/stm32/lock_resource.h
fpistm: ok

variants/PNUCLEO_WB55RG/variant.cpp
fpistm: Cuopy from MBed clock config. 😜 
Anyway at this time when doing the lock the comment is valid while the lock is kept.
And of course release will allow M0+ to configure it.
What do you propose instead ?

ABOSTM: "This prevents concurrent access to RCC registers by CPU2 (M0+)"

ABOSTM (changes requested):
I think HSEM is necessary in analog.cpp  when calling
__HAL_RCC_ADC_CONFIG()

fpistm:
> 
> 
> I think HSEM is necessary in analog.cpp when calling
> __HAL_RCC_ADC_CONFIG()

Right nice catch 👍

ABOSTM (approved):
LGTM

PR #1292, opened 2021-02-07
Optinally use TIM6 as timebase for the HAL, to allow usign the core inside FreeRTOS

ghent360:
**Summary**

RepRap firmware (RRF) is being ported to STM32. RRF is using FreeRTOS for multitasking.
When using FreeRTOS, the CubeMX suggests not to use SysTick as timebase for the HAL. I believe this is because there is conflict between the HAL and RTOS use of the SysTick timer. This PR is a simple hack that allows using TIM6 as timebase for the HAL for the PRNTR_Vx variant.

I propose this code to be used as basis for a more generic approach to allow configurable timebase.

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -42,6 +42,11 @@ timerObj_t *HardwareTimer_Handle[TIMER_NUM] = {NULL};
   */
 HardwareTimer::HardwareTimer(TIM_TypeDef *instance)
 {
+#ifdef USE_TIM6_TIMEBASE
+  if (instance == TIM6) {
+    Error_Handler();
+  }
+#endif
   uint32_t index = get_timer_index(instance);
   if (index == UNKNOWN_TIMER) {
     Error_Handler();
@@ -1421,6 +1426,12 @@ extern "C" {
 
   void HAL_TIM_PeriodElapsedCallback(TIM_HandleTypeDef *htim)
   {
+#ifdef USE_TIM6_TIMEBASE
+    if (htim->Instance == TIM6) {
+      HAL_IncTick();
+      return;
+    }
+#endif
     HardwareTimer::updateCallback(htim);
   }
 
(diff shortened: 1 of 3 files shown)


ABOSTM (changes requested):
Hi @ghent360,
Thanks for this proposal, nevertheless I have several remarks:
1) As far as HardwareTimer.cpp modifications are concerned: this is purely application code, there is no reason to add this code in HardwareTimer driver. 
This should be in application (or FreeRtos port, ...)

2) This is not generic enough:
The choice of TIM6 (and use of switch USE_TIM6_TIMEBASE) for such usage should be part of variant and not part of drivers (clock or HardwareTimer).
Warning, TIM6 is often used for TONE or SERVO HardwareTimer instance, and as I can see variant PRNTR_Vx use it
`#define TIMER_TONE              TIM6`

3) Concerning variants modifications: this is your variant so in principle, you do what you want, but from my point of view this is also some application code that should not be part of the variant, only choice of TIM instance may eventually remain.

Finally I wonder whether all that is required as there is a dedicated FreeRTOS port for STM32:
https://github.com/stm32duino/STM32FreeRTOS
and this take care of tick conflict between HAL and FreeRTOS.
Are you aware of this library?
Did you concider it ?
Is there a specific reason to not use **STM32FreeRTOS** library ?

ghent360:
Hi Alexandre,

Thank you for the feedback. I was not aware of the *STM32FreeRTOS* library.
I'll take a look at it. The RRF code is meant to be portable across
multiple Arm MCUs (Atmel, LPC and STM32). Custom vendor RTOS code may
complicate things, but these are solvable issues. I'm not sure how the tick
conflict would be resolved, I'll dig through the code.

In general all your points are valid, the proposal is not general at all.
However I think it should serve as a basis for discussion on a better more
generic approach. For example the timer interrupt routine is currently
inside the HardwareTimer.cpp and if I want to have a custom handler for
TIM6, I have to make modifications there. This is not ideal.

~V

ABOSTM:
```
For example the timer interrupt routine is currently
inside the HardwareTimer.cpp and if I want to have a custom handler for
TIM6, I have to make modifications there. This is not ideal.
```
Yes, `TIM6_IRQHandler()` is defined in HardwareTimer.cpp, but you can register your own callbacks thanks to attachInterrupt():
see various example:
https://github.com/stm32duino/STM32Examples/tree/master/examples/Peripherals/HardwareTimer

PR #1298, opened 2021-02-09
[USB] Manage internal pull-up

fpistm:
On STM32L1xx, an internal pull-up resistor (1.5 kΩ) can be connected by software on the USB data + (DP) line.
This internal pull-up resistor is enabled if the USB is not in power-down mode and if the USB_PU bit is set.


fpistm:
Tested with Nucleo L152RE.

PR #1311, opened 2021-03-01
Enable PWR clock systematically when available

ABOSTM:
**Summary**

Enable PWR clock systematically when available

This avoid to miss it in system clock configuration,
    as it is not systematically done by cubeMX

### DIFF
--- a/cores/arduino/stm32/usb/usbd_conf.c
+++ b/cores/arduino/stm32/usb/usbd_conf.c
@@ -53,16 +53,7 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd)
   const PinMap *map = NULL;
 #if defined(PWR_CR2_USV)
   /* Enable VDDUSB on Pwrctrl CR2 register*/
-#if !defined(STM32WBxx)
-  if (__HAL_RCC_PWR_IS_CLK_DISABLED()) {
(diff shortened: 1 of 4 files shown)


cores/arduino/stm32/usb/usbd_conf.c
```diff
@@ -55,9 +55,7 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd)
   /* Enable VDDUSB on Pwrctrl CR2 register*/
 #if !defined(STM32WBxx)
   if (__HAL_RCC_PWR_IS_CLK_DISABLED()) {
-    __HAL_RCC_PWR_CLK_ENABLE();
     HAL_PWREx_EnableVddUSB();
-    __HAL_RCC_PWR_CLK_DISABLE();
   } else
 #endif
   {
```
fpistm, line 61: As the PWR clock is considered enabled no need to test

fpistm (changes requested):

cores/arduino/stm32/usb/usbd_conf.c
ABOSTM: Right. Fixed.

fpistm (approved):
LGTM

PR #1312, opened 2021-03-02
spi: wait for end of transfer before returning from spi_transfer()

ABOSTM:
**Summary**
spi: wait for end of transfer before returning from spi_transfer()

This guarantee to application that transfer is finished,
and avoid for example that application releases NSS pin
while transfer is still ongoing.


fpistm (approved):
LGTM

PR #1313, opened 2021-03-03
HardwareSerial: don't call STM32 HAL for every bytes to transfer

### LINKED ISSUE #1309: STM32F1 UART TX idles after every byte
**Describe the bug**
When using UART1 on STM32F1, TX seem to idle after every byte sent.  Don't expect this idle period. 

**To Reproduce**
Source from Arduino IDE:

```
void setup() {
  Serial.begin(1000000);
}

void loop() {
  Serial1.printf("Test");
  Serial1.flush();
  delay(100);
}
```

Screenshot from Scope:
[
Capture_2021_02_24_18_50_39
](url)

**Expected behavior**
Expect to see continuous output from TX, such as:

Capture_2021_02_24_19_02_24

**Screenshots**
Screenshots above

**Desktop (please complete the following information):**
 - OS: Win 10
 - Arduino IDE version: 1.8.13
 - STM32 core version: 1.9.0
 - Tools menu settings if not the default: Default
 - Upload method: Serial

 - Name: STM32F103C Blue Pill

ABOSTM:
**Summary**

HardwareSerial: don't call STM32 HAL for every bytes to transfer

Instead of calling HAL for each and every bytes,
try to call it only once for a string.
Several call may be needed to manage internal circular buffer (tx_buff)

Fixes #1309

### DIFF
--- a/cores/arduino/HardwareSerial.cpp
+++ b/cores/arduino/HardwareSerial.cpp
@@ -327,7 +327,7 @@ int HardwareSerial::_tx_complete_irq(serial_t *obj)
 {
   // If interrupts are enabled, there must be more data in the output
   // buffer. Send the next byte
-  obj->tx_tail = (obj->tx_tail + 1) % SERIAL_TX_BUFFER_SIZE;
+  obj->tx_tail = (obj->tx_tail + obj->tx_size) % SERIAL_TX_BUFFER_SIZE;
 
   if (obj->tx_head == obj->tx_tail) {
     return -1;
@@ -467,8 +467,12 @@ void HardwareSerial::flush()
   // the hardware finished tranmission (TXC is set).
 }
 
-size_t HardwareSerial::write(uint8_t c)
+size_t HardwareSerial::write(const uint8_t *buffer, size_t size)
 {
+  tx_buffer_index_t i;
+  size_t size_tmp;
+  size_t ret = size;
+
   _written = true;
   if (isHalfDuplex()) {
     if (_rx_enabled) {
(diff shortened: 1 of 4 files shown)


cores/arduino/HardwareSerial.cpp
```diff
@@ -477,24 +480,60 @@ size_t HardwareSerial::write(uint8_t c)
...
-  _serial.tx_buff[_serial.tx_head] = c;
-  _serial.tx_head = i;
+  _serial.tx_size = size;
 
   if (!serial_tx_active(&_serial)) {
-    uart_attach_tx_callback(&_serial, _tx_complete_irq);
+    uart_attach_tx_callback(&_serial, _tx_complete_irq, size);
   }
 
   return 1;
```
fpistm, line 528: It should return the real size wrote?

cores/arduino/HardwareSerial.cpp
```diff
@@ -477,24 +480,60 @@ size_t HardwareSerial::write(uint8_t c)
...
+    write(buffer, size_intermediate);
+    size -= size_intermediate;
+    buffer += size_intermediate;
+  }
+
+  // Here size if less or equal to SERIAL_TX_BUFFER_SIZE, but SERIAL_TX_BUFFER_SIZE is not possible as tx_head = tx_tail is ambiguous empty or full
+  if (size == SERIAL_TX_BUFFER_SIZE) {
+    size_t size_intermediate = SERIAL_TX_BUFFER_SIZE - 1;
+
+    write(buffer, size_intermediate);
```
fpistm, line 496: The returned size should be checked ?

fpistm (changes requested):

cores/arduino/HardwareSerial.cpp
ABOSTM: Right, I have done this, even if there is no error management, so return the requested size

cores/arduino/HardwareSerial.cpp
fpistm: OK

ABOSTM:
**Performances measurements:**
Note: This PR subserve, by implementation, sending multi char string instead of single char.
This is the major usage.

Tests performed with UART @115200

| Test | Before PR | After PR| Performance|
| --- | --- | --- | --- |
| Nucleo_l476RG   <br> sending string of 20 char | 1,80 ms  | 1,72 ms  | gain: 4,4% |
| Nucleo_l476RG  <br>  sending 20 char one by one | 1,80 ms  | 1,88 ms  | loss: 4,4% |
| BluePill       <br>          sending string of 20 char | 1,90 ms  | 1,73 ms  | gain: 8,9% |
| BluePill       <br>          sending 20 char one by one | 1,90 ms  | 1,90 ms  | stable |

PR #1326, opened 2021-03-15
DAC fixes

### LINKED ISSUE #1324: DAC1 channel 2 and DAC2 channel 1 are not working on STM32F303
Only the first channel on DAC1 is working on STM32F303 (k8) as it's the only one with output buffer:
https://github.com/stm32duino/Arduino_Core_STM32/blob/6ec82f6f8df0bf56b2abd82998fea29196a7b279/libraries/SrcWrapper/src/stm32/analog.cpp#L406

The other two channels are using output switch instead of output buffer enable:  `dacChannelConf.DAC_OutputSwitch = DAC_OUTPUTSWITCH_ENABLE;`

As the HAL is using only correct struct member, it should be possible to set both without any interference:
`dacChannelConf.DAC_OutputBuffer = DAC_OUTPUTBUFFER_ENABLE;`
`dacChannelConf.DAC_OutputSwitch = DAC_OUTPUTSWITCH_ENABLE;`
This solves issue that DAC1 channel 2 and DAC2 channel 1 are not getting routed out to the output pins.

However there is another issue with initialization (after the previous fix):
`  analogWriteResolution(12);`
`  analogWrite(A4,512); // output is active for 5 seconds`
`  delay(5000);`
`  analogWrite(A3,1512); // now HAL_DAC_MspDeinit is called prior to initialize second channel and A4 is now floating`
`  analogWrite(A5,2512); // this channel is working after the fix`

Changing the order of first usage A3/A4 changes which one stops working (first one used stops working after the other channel gets "initialized" too)

ABOSTM:
**Summary**
* DAC: STM32F3 dac may have output switch instead of output buffer
  For STM32F3 serie, depending on instance/channel used,
  output buffer may be replaced by output switch
  Both being exclusive, they can be both set in HAL structure.

* DAC: remove HAL_DAC_DeInit() to avoid other channel deconfiguration
    
Fixes #1324

### DIFF
--- a/libraries/SrcWrapper/src/stm32/analog.cpp
+++ b/libraries/SrcWrapper/src/stm32/analog.cpp
@@ -389,12 +389,6 @@ void dac_write_value(PinName pin, uint32_t value, uint8_t do_init)
     return;
   }
   if (do_init == 1) {
-
-    if (HAL_DAC_DeInit(&DacHandle) != HAL_OK) {
-      /* DeInitialization Error */
-      return;
-    }
-
     /*##-1- Configure the DAC peripheral #######################################*/
     g_current_pin = pin;
     if (HAL_DAC_Init(&DacHandle) != HAL_OK) {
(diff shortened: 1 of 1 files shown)


fpistm (approved):
LGTM

kiiv-cz (approved):
Just tested, all three channels working now.

PR #1336, opened 2021-03-24
fix: Compile-time error caused by non-weak yield

Thor-x86:
**Summary**

Error message to be fixed:

```
wiring_time.c:(.text.delay+0xc): undefined reference to `yield'
exit status 1
Error compiling for board Generic STM32F1 series.
```

* [x] Bug

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -47,10 +47,12 @@ extern void initVariant() __attribute__((weak));
 extern void setup(void) ;
 extern void loop(void) ;
 
-static void __empty() {
+static void __empty()
+{
   // Empty
 }
-void yield(void) __attribute__ ((weak, alias("__empty")));
+  
+void yield(void) __attribute__((weak, alias("__empty")));
 
 #ifdef __cplusplus
 } // extern "C"
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -47,7 +47,11 @@ extern void initVariant() __attribute__((weak));
 extern void setup(void) ;
 extern void loop(void) ;
 
-void yield(void);
+static void __empty() {
+  // Empty
+}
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @Thor-x86 
how you met this issue?
the yield is already defined as empty in the hooks.c.

https://github.com/stm32duino/Arduino_Core_STM32/blob/01a1aa3ff3c1eb768389fb0b3f4b97b154e63c99/cores/arduino/hooks.c#L30-L34


Which IDE you used ?

Thor-x86:
The error exist on Arduino IDE with multiple .cpp and single .h for binding. Perhaps, is there any `something.h` to be included other than `#include<Arduino.h>`?

fpistm:
I never met this issue. Don't know why you got it. Include is correct. Could you please give all information of your setup. Sketch, ide version, target, ....

fpistm:
Moreover the `yield()` function is used by `delay()` function so really I don't know why you get this issue because a simple blink program should failed if it was not properly defined.
So please give us a detailed description on how you met this.

fpistm:
The only reasons I can see is that the core is not properly installed. Maybe an archive corruption... already saw that before...

Thor-x86:
> Could you please give all information of your setup. Sketch, ide version, target, host OS

Full source code [here](https://github.com/Thor-x86/arduino_cli).

I'm currently using latest stable Arduino IDE on Ubuntu 20.04, the problem detected while compiling for [STM32F103C8T6](https://www.st.com/en/microcontrollers-microprocessors/stm32f103c8.html) via HID bootloader.

fpistm:
I've built your sketch and have no issue as expected.
Really don't know how you met this issue anyway as stated before if it was a real issue it would be seen since more than 3 years.
So I close this PR as from my point of view it is not valid.

PR #1348, opened 2021-04-02
[F4] Update STM32F4xx HAL Drivers to v1.7.12

fpistm:
Update from [STM32CubeF4 FW v1.26.1](https://github.com/STMicroelectronics/STM32CubeF4/releases/tag/v1.26.1)


fpistm:
Core compilation failed due to new arduino-cli which is not backward compatible. See #1349

PR #1408, opened 2021-06-03
build options in source comments

koendv:
This PR allows embedding build options in project .ino source comments, like this:
```
// build-opt -DUSB_PRODUCT_STRING=\"device\ name\"

void setup() {
}

void loop() {
}
```
This way there is no need for a separate build_opt.h file. This also works when using arduino ide 2.0.0.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -111,8 +111,8 @@ build.opt.path={build.path}/sketch/{build.opt.name}
 extras.path={build.system.path}/extras
 
 # Create empty {build.opt} if not exists in the output sketch dir and force include of SrcWrapper library
-recipe.hooks.prebuild.1.pattern="{extras.path}/prebuild.sh" "{build.path}" "{build.source.path}"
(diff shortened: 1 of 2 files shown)


fpistm:
Hi @koendv 
thanks for this proposal.
At this time I will not merge this as I'd like to have Arduino feedback about https://github.com/arduino/arduino-ide/issues/363

fpistm:
Hi @koendv 
Thanks for this PR, it was a great idea anyway I will not merge them.
- PR https://github.com/stm32duino/Arduino_Core_STM32/pull/1442 solves the issue bring by the arduino-CLI
- Comment is only on one line and number of option can be high.
- Syntax have to be strict and I know this will bring support 😉 

When Arduino will fix the original issue then I will review this part moreover Arduino study/implement a way to "_Provide dedicated sets of "extra flags" properties for platform developers and users_"
https://github.com/arduino/arduino-cli/issues/846

PR #1431, opened 2021-06-25
HardwareTimer: update function header and parameter description

### LINKED ISSUE #1430: The function's help textes are wrong
The manual of many functions of the HardwareTimer.cpp are wrong. (https://github.com/stm32duino/Arduino_Core_STM32/blob/633606a4664b81a81042db3496d88629c39aaf39/cores/arduino/HardwareTimer.cpp) 

The manual of the functions:
`void HardwareTimer::setPWM(uint32_t channel, PinName pin, uint32_t frequency, uint32_t dutycycle, callback_function_t PeriodCallback, callback_function_t CompareCallback)
void HardwareTimer::setPWM(uint32_t channel, uint32_t pin, uint32_t frequency, uint32_t dutycycle, callback_function_t PeriodCallback, callback_function_t CompareCallback)`
are listing the parameter _format_ which does not exist! Instead the paramters _callback_function_t PeriodCallback_ and _callback_function_t CompareCallback_ exist but don't have any help text!?

Also this function:
_uint32_t HardwareTimer::getCaptureCompare(uint32_t channel,  TimerCompareFormat_t format)_ 
It does have a return value but the info text says that it does not have any return values.

Maybe there are some more wrong helping textes. These are the first one I found.

ABOSTM:
**Summary**

HardwareTimer: update function header and parameter description

Fixes #1430

### DIFF
--- a/cores/arduino/HardwareTimer.cpp
+++ b/cores/arduino/HardwareTimer.cpp
@@ -826,7 +826,7 @@ void HardwareTimer::setCaptureCompare(uint32_t channel, uint32_t compare, TimerC
   *           TICK_FORMAT:     return value is the number of tick for Capture/Compare value
   *           MICROSEC_FORMAT: return value is the number of microsecondes for Capture/Compare value
   *           HERTZ_FORMAT:    return value is the frequency in hertz for Capture/Compare value
-  * @retval None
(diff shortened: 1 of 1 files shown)


fpistm (approved):
Thanks @ABOSTM

PR #1438, opened 2021-07-06
library: Wire/SPI: change default pins type

### LINKED ISSUE #1432: SPI library: can't accept pin numbers with _ALTx in constructor
Version2 core introduced pin numbers with _ALTx (ex. PB5_ALT1) and  I can use SPI3 with pins PB5:MOSI, PC11:MISO and PC10:SCLK in this way:
```
SPIClass SPI_3rd;// set pin numbers later.

SPI_3rd.setMOSI(PB5_ALT1);
SPI_3rd.setMISO(PC11);
SPI_3rd.setSCLK(PC10);
SPI_3rd.begin();
SPI_3rd.setClockDivider(SPI_CLOCK_DIV16);
SPI_3rd.transfer(0x10);// any data
```

But I can't:
```
SPIClass SPI_3rd(PB5_ALT1, PC11, PC10);// passing all pin numbers.

SPI_3rd.begin();
SPI_3rd.setClockDivider(SPI_CLOCK_DIV16);
SPI_3rd.transfer(0x10);// any data
```

When I call transfer(), the program keeps waiting TXE here.
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/libraries/SPI/src/utility/spi_com.c#L443

**Probably the problem is:**
The constructor takes pin numbers as uint8_t, so I pass PB5_ALT1 (= PB5 | ALT1), but mosi's value still equals PB5.
```
SPIClass(uint8_t mosi, uint8_t miso, uint8_t sclk, uint8_t ssel = (uint8_t)NC);
```

**Versions:**
- OS: Windows 10 Pro
- PlatformIO IDE: 2.3.2
- VSCode: 1.57.0
- platform-ststm32: 14.0.0
- STM32 core version: 2.0.0

**Board:**
 - Nucleo L476RG

fpistm:
Since `_ALTx` introduction, default Arduino API using `uint8_t` type for pin number could not use them and raised raised this error:

```log
warning: unsigned conversion from 'int' to 'uint8_t' {aka 'unsigned char'} changes value from '260' to '4' [-Woverflow]
   91 | #define PB5_ALT1                (PB5  | ALT1)
      |                                 ~~~~~~^~~~~~~
sketch_jul06a.ino:10:20: note: in expansion of macro 'PB5_ALT1'
   10 |   SPIClass SPI_3rd(PB5_ALT1, PC11, PC10);
      |                    ^~~~~~~~
```

Moving API to uint32_t allows to use `_ALTx`, anyway some libraries used the default pin definition as an uint8_t from the variant definition, for example: 
```C
#ifndef PIN_SPI_MOSI
  #define PIN_SPI_MOSI          PB5_ALT1
#endif
```

Which is stored in:
```C
static const uint32_t MISO = PIN_SPI_MISO;
```

With Arduino SD library:
https://github.com/arduino-libraries/SD/blob/a64c2bd907460dd01cef07fff003550cfcae0119/src/utility/Sd2Card.h#L75-L83

```log
libraries\SD\src\utility\Sd2PinMap.h:28:28: warning: conversion from 'uint32_t' {aka 'long unsigned int'} to 'uint8_t' {aka 'unsigned char'} changes value from '260' to '4' [-Woverflow]
   28 |   uint8_t const MOSI_PIN = MOSI;
      |                            ^~~~
```

In this case this is not risky as those conversion will use the correct pin for SOFTWARE_SPI (GPIO toggling) not related to a peripheral and in major way default pins uses value without `_ALTx` which mean an unit8_t size and so does not raised warning.
Anyway not all libraries could use `_ALTx` feature.

Fixes #1432

### DIFF
--- a/cores/arduino/pins_arduino.h
+++ b/cores/arduino/pins_arduino.h
@@ -80,8 +80,8 @@ static const uint32_t SCK  = PIN_SPI_SCK;
   #define PIN_WIRE_SCL              15
 #endif
 
-static const uint8_t SDA = PIN_WIRE_SDA;
-static const uint8_t SCL = PIN_WIRE_SCL;
+static const uint32_t SDA = PIN_WIRE_SDA;
+static const uint32_t SCL = PIN_WIRE_SCL;
 
 #ifdef __cplusplus
 extern "C" {
--- a/cores/arduino/pins_arduino.h
+++ b/cores/arduino/pins_arduino.h
@@ -64,13 +64,13 @@ _Static_assert(NUM_ANALOG_INPUTS <= MAX_ANALOG_INPUTS,
   #define PIN_SPI_SCK               13
 #endif
 
-static const uint8_t SS   = PIN_SPI_SS;
-static const uint8_t SS1  = PIN_SPI_SS1;
-static const uint8_t SS2  = PIN_SPI_SS2;
-static const uint8_t SS3  = PIN_SPI_SS3;
-static const uint8_t MOSI = PIN_SPI_MOSI;
-static const uint8_t MISO = PIN_SPI_MISO;
-static const uint8_t SCK  = PIN_SPI_SCK;
+static const uint32_t SS   = PIN_SPI_SS;
+static const uint32_t SS1  = PIN_SPI_SS1;
+static const uint32_t SS2  = PIN_SPI_SS2;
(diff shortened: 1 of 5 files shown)


ABOSTM (approved):
LGTM

PR #1445, opened 2021-07-13
fix: library: EEPROM: use correct flash size

### LINKED ISSUE #1316: Wrong flash size for eeprom emulation
Hey,

I am trying to use eeprom on the STM32L432.  It should be pretty straight forward but I dont appear to be having any luck.  Is someone able to confirm it works?

Cheers

fpistm:
`FLASH_SIZE` defined by the CMSIS Device Peripheral Access Layer Header File is the maximum one of a product line.
So depending of the mcu this is not the correct one.

As an example, L432KC have 256 KB while L432Kb has 128 KB.

Using `LL_GetFlashSize()` allows to get the correct FLASH size.

Fixes #1316

### DIFF
--- a/libraries/EEPROM/src/utility/stm32_eeprom.c
+++ b/libraries/EEPROM/src/utility/stm32_eeprom.c
@@ -17,6 +17,7 @@
   */
 
 #include "stm32_eeprom.h"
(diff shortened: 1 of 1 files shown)


ABOSTM (approved):
LGTM

PR #1446, opened 2021-07-13
config: add debug information menu

### LINKED ISSUE #873: Improvement: Enable -g on all builds
Currently, the `-g` compiler option for adding debug symbols is only passed when the optimization level is "Debug". However, I think that passing -g is *always* a good idea, even with higher optimization levels.

Of course, when you expect to be debugging, you can simply enable the debug optimization level for a build. However, sometimes you need to debug a production build, in particular when you cannot reproduce the problem using a debug build, or when you want to analyze a stack dump collected from some production system.

It is true that aggressive optimizations might interfere with debugging, but in practice debugging still works reasonably well (and certainly better than having *no* debugging symbols at all).

Also, AFAIK adding debug symbols does not affect the produced machine code, so the only downside of adding debug symbols might be that compilation is a little slower.

Implementing this would, I think, be a matter of removing the `-g` option from the Debug optimization levels in `boards.txt` (and replacing `-g` with `-Og` in its description, which is a good idea regardless). Then, the `-g` option can be added unconditionally in `platformt.txt`.

fpistm:
This PR add a new menu `Debug information`.
By default it is not enabled. Enable it will produce debugging information.

image

It also fix the `Optimize` menu which displayed `Debug (-g)` instead of "Debug (-Og)".

Fixes #873

For further reading: https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#Debugging-Options

/CC @matthijskooijman

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -8,6 +8,7 @@ menu.xusb=USB speed (if available)
 menu.virtio=Virtual serial support
 
 menu.opt=Optimize
+menu.dbg=Debug symbols
 menu.rtlib=C Runtime Library
 menu.upload_method=Upload method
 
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Patch looks good to me. Maybe name it "Debug symbols" instead of "Debug information"? I can imagine the latter makes people expect that additional debug info will be printed?

fpistm:
Thanks @matthijskooijman for your feedback.
I agree with you, information can be confusing even if it's purely the name used by gcc for the debugging options.
I will do the change.

PR #1482, opened 2021-08-22
Fix `bitWrite` with parenthesis

positron96:
Current version does not wrap arguments in "()", so works incorrectly when used like this:
```
bitWrite(var, bit, cond?1:0);
```

Same as in here: https://github.com/espressif/arduino-esp32/issues/4466


fpistm (approved):
LGTM, even if I think `value` and `bit` does not need extra parenthesis.

fpistm:
Hi @positron96,

thanks for this PR. 
To go further, I would raised also an issue in the ArduinoCore-API:
https://github.com/arduino/ArduinoCore-API/blob/d06d9b27c9de71b871a6b80ca9bc67e5e7a6c603/api/Common.h#L68

PR #1489, opened 2021-09-08
chore: update SPI pins for Swan_R5

zfields:
**Summary**

update SPI pins (MOSI, MISO, SCK, SS) for Swan_R5

[Bug] SPI does not trigger the appropriate pins.

I work at Blues Wireless. During preparation and testing of the Swan, I discovered the SPI pins were not functional through the Feather pinouts.

**Validation**

* I have used the code to drive an MCP23S17 from the Swan.

**Code formatting**

It's all in the preprocessor, and it was copied from the existing pattern.

**Closing issues**

### DIFF
--- a/variants/STM32L4xx/L4R5Z(G-I)Y_L4R9Z(G-I)Y_L4S5ZIY_L4S9ZIY/variant_SWAN_R5.h
+++ b/variants/STM32L4xx/L4R5Z(G-I)Y_L4R9Z(G-I)Y_L4S5ZIY_L4S9ZIY/variant_SWAN_R5.h
@@ -181,9 +181,23 @@
   #define USER_BTN              PC13
 #endif
 
-// On-board user button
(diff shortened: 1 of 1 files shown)


fpistm:
Hi @zfields ,
In fact default SPI instance pins were defined to use PA5/PA6/PA7 by default using SPI peripheral.
So with your change it will use SPI2 peripheral with the pins you defined.
If it is what you want by default, it is OK.

zfields:
Oh, no! I must have felt handcuffed by your PR template. :smile: No worries though! I actually labelled it as a "chore" in the real commit message, because I didn't really see it as a bug either. :+1: 

Long story short, our out of the box configuration exposes the STM32L4xx pins I specified via the Adafruit Feather pinout. I was working from the board schematic and an oscilloscope in order to make the change.

fpistm (approved):
LGTM

PR #1490, opened 2021-09-08
feat: `BUTTON_BUILTIN` alias for `USER_BTN`

zfields:
Create an alias for those new to STM32duino environment.

`USER_BTN` is commonplace in STM32duino, however there are several instances of `BUTTON_BUILTIN` across other Arduino BSPs. The naming convention of `BUTTON_BUILTIN` is more closely aligned to `LED_BUILTIN`, and there appears to be some precedent.

Examples:
- https://github.com/espressif/arduino-esp32/blob/master/variants/micro_s2/pins_arduino.h#L75
- https://github.com/esp8266/Arduino/blob/master/variants/espectro/pins_arduino.h#L40
- https://github.com/f32c/arduino/blob/master/hardware/fpga/f32c/variants/FleaFPGA-Uno/variant.h#L17
_... as well as several other examples that can be discovered by Googling, `"BUTTON_BUILTIN" arduino`._

Since ST is the clear leader in boards having a built-in button, it has a unique opportunity to help reinforce the semi-standardized precedent `BUTTON_BUILTIN` by providing an alias for all boards that support a `USER_BTN`.

### DIFF
--- a/cores/arduino/pins_arduino.h
+++ b/cores/arduino/pins_arduino.h
@@ -40,6 +40,11 @@ _Static_assert(LastPort <= 0x0F, "PortName must be less than 16");
 _Static_assert(NUM_ANALOG_INPUTS <= MAX_ANALOG_INPUTS,
                "Core NUM_ANALOG_INPUTS limited to MAX_ANALOG_INPUTS");
 
+/* Alias USER_BTN when available */
+#if defined(USER_BTN) && !defined(BUTTON_BUILTIN)
+  #define BUTTON_BUILTIN            USER_BTN
+#endif
+
 /* Default for Arduino connector compatibility */
 /* SPI Definitions */
 #ifndef PIN_SPI_SS


fpistm:
Hi @zfields 
6 years I'm working in Arduino Ecosystem and honestly, I never seen this definition.
Searching with google give me only 2 pages. It seems mainly defined by ESP.
All standard Arduino examples use the `USER_BTN` definition.

For Ardunio, no reference:
https://github.com/search?q=org%3Aarduino+%22BUTTON_BUILTIN%22
https://github.com/search?q=org%3Aarduino-libraries+BUTTON_BUILTIN

Same for Adafruit:
https://github.com/search?q=org%3Aadafruit+BUTTON_BUILTIN

And on all GitHub, only 327 occurrences in code:
https://github.com/search?p=32&q=%22BUTTON_BUILTIN%22&type=Code

I think it is really not a good idea to add it, the Arduino standard is `USER_BTN`. Using `BUTTON_BUILTIN` in a code would break Arduino compatibility. If you really want have it as a standard it should be submit first to Arduino then it will benefit to all other project following Arduino standard.

ABOSTM:
I am in line with @fpistm comments.
I prefer to reinforce the Arduino standard USER_BTN (alone),
instead of having a second semi standard definition for the same thing, introducing confusion.
Maybe, library using BUTTON_BUILTIN should stick to Arduino standard (USER_BTN ), or create the alias on theyr own ...

PR #1494, opened 2021-09-15
Configure Swan R5 initial behaviors

zfields:
After putting the Swan through it's paces, we decided it would make a better experience to modify the following behaviors.

chore: Enable Swan I2C pull-ups by default

feat: Improve Swan 3V3 pin control/behavior
- Define constant to reference `DISCHARGE_3V3` pin.
- Define constants `ENABLE_DISCHARGING` and `DISABLE_DISCHARGING` to manipulate `DISCHARGE_3V3` pin.
- Initialize `ENABLE_3V3` pin to be on, and `DISCHARGE_3V3` to be disabled.


fpistm (approved):
LGTM

PR #1506, opened 2021-09-23
spi: Add delay before disabling SPI to avoid truncated clock/data signals

### LINKED ISSUE #1294: NUCLEO-H743ZI2 SPI.transfer() fails with SCLK less than 500 kHz in 1.9.0
This sketch illustrates a failure of SPI.transfer with a NUCLEO-H743ZI2.  The sketch use a SCLK rate of 400kHz.
```
 #include "SPI.h"
const uint8_t CS_PIN = 10;
void setup() {
  SPI.begin();
  pinMode(CS_PIN, OUTPUT);
}
void loop() {
  SPI.beginTransaction(SPISettings(400000, MSBFIRST, SPI_MODE0));
  digitalWrite(CS_PIN, LOW);
  SPI.transfer(0x55);
  digitalWrite(CS_PIN, HIGH);  
  SPI.endTransaction();
  delay(1);
}
```
The problem is shown in this trace of SCLK in yellow and MOSI in green.  The eighth clock pulse  and data are chopped. 

nodelay

I looked at the SPI library code and found where the problem occurs,  SPI is disabled before the last bit is complete.

The file is: STM32/hardware/stm32/1.9.0/libraries/SPI/src/utility/spi_com.c
If you add a delay here at about line 443:
```
#if defined(STM32H7xx) || defined(STM32MP1xx)
  /* Close transfer */
  /* Clear flags */
  HAL_Delay(1);  //  <<-------------- added by WHG to delay call LL_SPI_Disable()
  LL_SPI_ClearFlag_EOT(_SPI);
  LL_SPI_ClearFlag_TXTF(_SPI);
  /* Disable SPI peripheral */
  LL_SPI_Disable(_SPI);
#endif
```
The SPI frame is correct.

delay

I tried to find a SPI status check to replace the delay but was not successful when I looked at the STM32H743/753 Reference Manual.

ABOSTM:
**Summary**
spi: Add delay before disabling SPI to avoid truncated clock/data signals

Add a delay before disabling SPI otherwise last-bit/last-clock
may be truncated.
See https://github.com/stm32duino/Arduino_Core_STM32/issues/1294
Computed delay is half SPI clock.

Fixes #1294

### DIFF
--- a/libraries/SPI/src/utility/spi_com.c
+++ b/libraries/SPI/src/utility/spi_com.c
@@ -35,6 +35,7 @@
   *
   ******************************************************************************
   */
+#include "wiring_time.h"
 #include "core_debug.h"
 #include "stm32_def.h"
 #include "utility/spi_com.h"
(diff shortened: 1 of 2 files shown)


ABOSTM:
cc/ @greiman @mzhboy

libraries/SPI/src/utility/spi_com.c
```diff
@@ -35,6 +35,7 @@
   *
   ******************************************************************************
   */
+#include "Arduino.h"
```
fpistm, line 38:
```suggestion
#include "wiring_time.h"
```

fpistm (changes requested):

libraries/SPI/src/utility/spi_com.c
ABOSTM: Done

fpistm (approved):

PR #1562, opened 2021-11-26
Wire: Max TX buffer length

### LINKED ISSUE #1539: Wire: limit buffer size allocation
Currently, built-in Wire library dynamically allocate Rx and Tx Buffer and raised an error when no more space are available.
(`TwoWire::allocateRxBuffer` and `TwoWire::allocateTxBuffer`)
Goal of this issue is to add a maximum buffer length and return 0 if it is reached to allow application to know the buffer is full and need to be flushed (`endTransmission`).
Ex: https://github.com/simondlevy/VL53L5/blob/b81b3d1bef74bdecbe7615e3a498a8648a2fdd0d/src/st/platform.cpp#L117-L140

ABOSTM:
**Summary**
 Wire: limit TX buffer size to avoid consuming too much stack
    
If data size to be sent is greater than the max defined,
application can split the transfert
Note: no limit implemented in RX buffer, because when receiving,
data are received on I2C, it is not possible to split.

Fixes #1539


libraries/Wire/src/Wire.h
```diff
@@ -28,23 +28,27 @@ extern "C" {
 #include "utility/twi.h"
 }
 
+// Minimal buffer length. Buffers length will be increased when needed,
+// but TX buffer is limited to a maximum to avoid too much stack consumption
 #define BUFFER_LENGTH 32
+#if !defined(WIRE_MAX_TX_BUFF_LENGTH)
```
fpistm, line 34: It should probably be fine to specify that it is limited to uint16_t size ?

libraries/Wire/src/Wire.h
```diff
@@ -28,23 +28,27 @@ extern "C" {
...
+  #define WIRE_MAX_TX_BUFF_LENGTH       1000U
```
fpistm, line 35:
```suggestion
  #define WIRE_MAX_TX_BUFF_LENGTH       1024U
```

libraries/Wire/src/Wire.cpp
```diff
@@ -327,17 +324,18 @@ size_t TwoWire::write(const uint8_t *data, size_t quantity)
 
   if (transmitting) {
     // in master transmitter mode
-    allocateTxBuffer(txBufferLength + quantity);
-    // error if no memory block available to allocate the buffer
-    if (txBuffer == nullptr) {
+    if (allocateTxBuffer(txDataSize + quantity) == 0) {
+      ret = 0;
+    } else if (txBuffer == nullptr) {
```
fpistm, line 329: I wonder if this test is needed as if allocateTxBuffer failed it reach Error_Handler ?

fpistm (changes requested):

libraries/Wire/src/Wire.h
ABOSTM: Done

libraries/Wire/src/Wire.h
ABOSTM: Done

libraries/Wire/src/Wire.cpp
ABOSTM: Right, I remove it for both TX and RX

PR #1602, opened 2021-12-17
interrupts: Disable EXTI when disabling interrupt

ABOSTM:
**Summary**

interrupts: Disable EXTI when disabling interrupt

When disabling interrupt, it should be done at both NVIC and EXTI level.
This avoid to have pending interrupt at EXTI level after call to
stm32_interrupt_disable(), which could fire the callback as soon as
new call to attachInterrupt(), even without new gpio change.

This PR aims to replace/simplify #1258 (which also replace #1168)

### DIFF
--- a/libraries/SrcWrapper/src/stm32/interrupt.cpp
+++ b/libraries/SrcWrapper/src/stm32/interrupt.cpp
@@ -37,6 +37,7 @@
   */
 #include "interrupt.h"
 #include "lock_resource.h"
+#include "stm32yyxx_ll_exti.h"
 #if !defined(HAL_EXTI_MODULE_DISABLED)
 
 /* Private Types */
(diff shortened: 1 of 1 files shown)


ABOSTM:
I made a sketch on my side to check that gpio toggle while interrupt is detached doesn't fire callback when interrupts is re-attached.
@cparata,
I know it is an old story, but would be great if you could test on your side the original issue you faced :smiley:

cparata:
Hi @ABOSTM ,
I will try next week with motor drivers.
Best Regards,
Carlo

fpistm (approved):

cparata:
Hi @ABOSTM ,
I did the test with motor driver and the issue, that I saw, is solved with this fix.
So, I approve it! ;-)
Thank you and Best Regards,
Carlo

Issue #1600, opened 2021-12-15
Add SPI slave example


brainelectronics:
The usage of the SPI library in slave mode is not straight forward, maybe even not implemented.

All examples show how to transmit data FROM a STM32 to some Arduino board (which is not a STM32). In some usecases the STM32 should also be able to be the controlled device to provide data to the SPI master or perform low level task.

brainelectronics:
_Also already requested by @AnHardt in https://github.com/stm32duino/Arduino_Core_STM32/issues/257#issuecomment-553774770_

ABOSTM:
Hi @brainelectronics ,
In fact the point is not to have an SPI slave example:
the problem is that SPI library from this core, supports only master and doesn't support slave,
and it is the same for Arduino SPI library.

Note that you can still make use of STM32Cube HAL/LL API within Arduino sketch.
Thus you can look at STM32Cube HAL/LL SPI examples.

brainelectronics:
@ABOSTM thanks for clarification.
Meanwhile implemented it with the HAL SPI examples for the project I'm currently working on, just thought as some extension of the SPI lib is currently ongoing it might fit in there as well.

Big thank you for this great package and your contributuion, that is maybe not often enough said 👍

PR #1612, opened 2021-12-29
Add SPI slave support and basic example

brainelectronics:
## Summary

This PR implements SPI slave mode.

## References

This is a follow up for the ticket #1600.

## Motivation

Most high level possible functions to implement this feature were used. In
order to be as close as possible to other Arduino examples.

`SPIClass` functions have not been touched as much as possible.

## Validation

Example sketches for SPI master and slave are provided with this PR.

### DIFF
--- a/libraries/SPI/examples/SPIMaster/SPIMaster.ino
+++ b/libraries/SPI/examples/SPIMaster/SPIMaster.ino
@@ -1,0 +1,205 @@
+/*
+  SPI Master
+
+  This example implements a SPI Master device for the SPI Slave example.
+  The data is transmitted to the SPI Slave and its response received back.
+
+  Every 1000ms the LED state toggles and the latest received data is printed
+  to the Serial port
+
+  The circuit:
+   * A LED to show a heartbeat signal
+   * CS - to digital pin PA4 (SS pin)
+   * SCK - to digital pin PA5 (SCK pin)
+   * MISO - to digital pin PA6 (MISO pin)
+   * MOSI - to digital pin PA7 (MOSI pin)
+
+  Default SPI pin definitions can be found at variants/STM32XXxx/<VARIANT>/variant_generic.h
+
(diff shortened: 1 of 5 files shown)


ABOSTM (changes requested):
@brainelectronics,
Thanks for this Pull Request.

This is a good starting point, but if we go for an implementation of SPI slave in this core, your proposal doesn't match with Arduino API spirit:
* user application should access to `SPIClass `API 
but it should not access to `spi_init()` for example (which is more for an internal core usage)
* user application (like the proposed example) should not use STM32 HAL API like HAL_SPI_Receive(), ...
* ... on the contrary, all this HAL_SPI_xxx() stuff should be part of the driver itself:
   For example by reusing/extending SPIClass like `begin()`  with an optional parameter to specify the mode (with default master).
   Also maybe adding new function to this class like a function to register slave receive call back, and another function to provide the slave response ...
* Driver, as well as example, should be compatible with any stm32 series supported. 
This means that :
  - driver implementation should use STM32 HAL/LL API common to STM32 series (or should be implement with compilation switch depending on series). 
  - Example should use the default and predefined variable `SPI` (default SPI instance)
     Instead of defining DEBUG_LED, you can use predefined LED_BUILTIN
     and remove reference to specific MCU pin like PA4

ric96:
Hi, is this actively being worked on?
I wanted to use this functionality and am willing to test and co-auth if required.

brainelectronics:
Hi @ric96 
Unfortunately I don't have much time to work on it. Another project is taking all my free time right now. Of course I'm happy with any contributions here

mirko:
I'd also really like to see this going upstream and I think it'd be worth putting some effort into once for a peace of mind later on, when using it in a project and not having to continuously level local patchsets.
How can I help?

fpistm:
Superseded by #2050.

PR #1616, opened 2022-01-06
gpio: Enable PWR CLK when enabling VddIO2 for GPIOG

ABOSTM:
**Summary**
gpio: Enable PWR CLK when enabling VddIO2 for GPIOG

### DIFF
--- a/libraries/SrcWrapper/src/stm32/PortNames.c
+++ b/libraries/SrcWrapper/src/stm32/PortNames.c
@@ -105,6 +105,7 @@ GPIO_TypeDef *set_GPIO_Port_Clock(uint32_t port_idx)
     case PortG:
 #if defined(PWR_CR2_IOSV) || defined(PWR_SVMCR_IO2VMEN)
       // Enable VDDIO2 supply for 14 I/Os (Port G[15:2])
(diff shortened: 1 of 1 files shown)


ABOSTM:
This PR is abandoned because issue is in STM32U5 HAL,
nevertheless this could be used as a workaround while HAL fix is not released. 

(internal issue 112307).
Following functions
* HAL_RCC_OscConfig()
* HAL_RCC_ClockConfig()
* RCC_SetFlashLatencyFromMSIRange()

are disabling PWR CLK unconditionaly
ex: https://github.com/STMicroelectronics/STM32CubeU5/blob/0eedae3d2a7997a2b5fbfdb57f54c7201e0827c8/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_hal_rcc.c#L1227 

whereas HAL should restore the PWR CLK state, like it is done at some other places like:
https://github.com/STMicroelectronics/STM32CubeU5/blob/0eedae3d2a7997a2b5fbfdb57f54c7201e0827c8/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_hal_rcc.c#L1016

PR #1617, opened 2022-01-06
Interrupts: remove C prototype and keep only C++ prototypes

### LINKED ISSUE #1601: Call of overloaded attachInterrupt is ambiguous with a lambda callback
When calling attachInterrupt with a lambda callback instead of a function I got the following error:
```c++
error: call of overloaded 'attachInterrupt(const uint8_t&, setup()::<lambda()>, int)' is ambiguous
   32 |     attachInterrupt(pin_trackpad_clock[0], [](){trackpad[0].int_on_clock();}, FALLING);
...
C:\Users\<user>\.platformio\packages\framework-arduinoststm32\cores\arduino/WInterrupts.h:28:8: note: candidate: 'void attachInterrupt(uint32_t, callback_function_t, uint32_t)'
   28 |   void attachInterrupt(uint32_t pin, callback_function_t callback, uint32_t mode);
      |        ^~~~~~~~~~~~~~~
C:\Users\<user>\.platformio\packages\framework-arduinoststm32\cores\arduino/WInterrupts.h:32:6: note: candidate: 'void attachInterrupt(uint32_t, void (*)(), uint32_t)'
   32 | void attachInterrupt(uint32_t pin, void (*callback)(void), uint32_t mode);
      |      ^~~~~~~~~~~~~~~
```
Header file WInterrupts.h indeed has 2 prototypes of this function and compiler can't decide which one to use with a lambda.
When the prototype with a function pointer (line 32) is removed, project compiles and works no matter attachInterrupt is called with a lambda or a function pointer.

Please note that Arduino version of attachInterrupt only has one prototype with a function pointer
and esp32 version has one prototype with std::function<void(void)>.

(Edit)
I am using default installation of PlatformIO in VS Code.
I have also tried the same concept with MSVC and MinGW with the same outcome.

ABOSTM:
**Summary**
Interrupts: remove C prototype and keep only C++ prototypes

attachInterrupt() C prototypes was not usable from C source file
(inheritance of former pure C implementation), and causes ambiguity
when used with lambda function as callback
Fixes #1601

### DIFF
--- a/cores/arduino/wiring.h
+++ b/cores/arduino/wiring.h
@@ -36,7 +36,6 @@
 #include "wiring_pulse.h"
 #include "wiring_shift.h"
 #include "wiring_time.h"
-#include "WInterrupts.h"
 
 #include <board.h>
 
@@ -44,8 +43,9 @@
   #include "HardwareTimer.h"
   #include "Tone.h"
   #include "WCharacter.h"
-  #include "WSerial.h"
+  #include "WInterrupts.h"
   #include "WMath.h"
+  #include "WSerial.h"
   #include "WString.h"
 #endif // __cplusplus
 
--- a/cores/arduino/WInterrupts.h
+++ b/cores/arduino/WInterrupts.h
@@ -20,17 +20,10 @@
 #define _WIRING_INTERRUPTS_
 
 #include <stdint.h>
+#include <functional>
 
-#ifdef __cplusplus
-  #include <functional>
-
-  typedef std::function<void(void)> callback_function_t;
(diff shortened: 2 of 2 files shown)


cores/arduino/wiring.h
```diff
@@ -47,6 +46,7 @@
   #include "WSerial.h"
   #include "WMath.h"
   #include "WString.h"
+  #include "WInterrupts.h"
```
fpistm, line 49: Include are sorted by name

fpistm (changes requested):

fpistm (approved):

PR #1609, opened 2021-12-25
Add variant G070RB

MrZloHex:
## Pull Request template

**Summary**

Adding new variant for Generic `G070RBT`

This PR fixes/implements the following **bugs/features**

* [ ] New variant for NUCLEO-G070RB
* [X] Add variant for bare STM32G070RB

New variant will be supported by stm32duino

PR #1638, opened 2022-01-31
Adding new variant for Generic G070xxT

techstudio-design:
**Summary**

* [x] Adding new variant for Generic G070CBT/KBT/RBT

I run the AStyle check and Action locally before commit, all passed.


ABOSTM (changes requested):
Hi @e-tinkers
Thanks for this PR, there was already a PR #1609 but I prefer to keep yours as it supersedes #1609
One minor comment otherwise Looks Good To Me.
By the way you could squash the whole in a single commit.
Generally we don't create commits for a fix (like Astyle), but directly amend the faulty commit.
Thanks

ABOSTM (approved):

PR #1654, opened 2022-02-10
fix: arduino-cli upload failed

fpistm:
`Error during Upload: Property 'upload.tool.serial' is undefined`

Issue present for arduino-cli version started from 0.19.0 to 0.21.0.
See https://github.com/arduino/arduino-cli/issues/1444

/cc @matthijskooijman

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -20,6 +20,7 @@ Nucleo_144.build.core=arduino
 Nucleo_144.build.board=Nucleo_144
 Nucleo_144.build.variant_h=variant_{build.board}.h
 Nucleo_144.build.extra_flags=-D{build.product_line} {build.enable_usb} {build.xSerial}
+Nucleo_144.upload.tool.default=massStorageCopy
 
 # NUCLEO_F207ZG board
 Nucleo_144.menu.pnum.NUCLEO_F207ZG=Nucleo F207ZG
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I'm not entirely sure this fix is correct. I have not tested, but AFAUI this will no longer allow changing the upload method using the "menu" entries using older arduino-cli, since you now set a new-style upload tool for the board as a whole, but no (new-style) tool for the menu entries (only old-style tool lines, but those are not converted in 0.21).

IOW, I think you need lines like below (instead of the ones you added now):
```
Nucleo_144.menu.upload_method.MassStorage.upload.tool.default=massStorageCopy
Nucleo_144.menu.upload_method.swdMethod.upload.tool.default=stm32CubeProg
```

Also, might be good to mention the full range of affected arduino-cli versions, looks like this problem was introduced in 0.19 with [pluggable discovery](https://github.com/arduino/arduino-cli/commit/ec027a7f435d63d34f0dfd21a8274ece655712e9), it was already working before that.

fpistm:
> I'm not entirely sure this fix is correct. I have not tested, but AFAUI this will no longer allow changing the upload method using the "menu" entries using older arduino-cli, since you now set a new-style upload tool for the board as a whole, but no (new-style) tool for the menu entries (only old-style tool lines, but those are not converted in 0.21).
> 

In fact with IDE it works  but not with the cli.

> IOW, I think you need lines like below (instead of the ones you added now):
> 
> ```
> Nucleo_144.menu.upload_method.MassStorage.upload.tool.default=massStorageCopy
> Nucleo_144.menu.upload_method.swdMethod.upload.tool.default=stm32CubeProg
> ```
> 
Unfortunately this is not recognized properly. I guess no fix can be provided for this.
So upload will simply not be compatible with arduino-cli version started from 0.19.0 to 0.21.0.

Now I wondering if we could use the pluggable discovery feature....

PR #1655, opened 2022-02-10
add STM32G031FxPx (and STM32G031Y8Yx, STM32G041F6Px, STM32G041F8Px STM32G041Y8Yx)

olikraus:
This PR adds STM32G031FxPx devices to the boards.txt file.

The STM32G031FxPx TSSOP20 variant was missing. I think this is a nice and powerful variant for small DIY projects. More pins than the SO8, but easier to solder compared to the LQFP variants.


olikraus:
Oh, nice, I didn't know that wiki page. I will go through the wiki and try to provide missing information.

olikraus:
ok, I fixed boards.txt and added ldscript.ld.

> instead of adding only 3 new generic, (G031F4Px, G031F6Px and G031F8Px) you can copy/paste boards_entry.txt (which contains 7 new generic variant

Done for G031FxPx

> you need to add the new variants to README.md

Done

> for the new variant, we need a linker script : ldscript.ld

Done

> Generic System Clock configuration

I think the default clock configuration was already there and is fine from my point of view. At least no changes required.

I also did a successful test with my own STM32G031F8Px board.

Note: cores_config_ci.json and cores_config.json not yet updated.

ABOSTM (changes requested):
@olikraus 
Thanks for this update.
To be complete, you could add the missing board sharing the same directory

ABOSTM:
Sorry, part of my comment did come up.
I mean add all the variants sharing the same directory. So yes  G031Y8Yx,  G041F6Px, G041F8Px G041Y8Yx
 you can copy/paste `boards_entry.txt ` (which contains 7 new generic variant)
Don't forget to update README.md accordingly.
Thanks

olikraus:
> you can copy/paste boards_entry.txt (which contains 7 new generic variant)

Done

> Don't forget to update README.md accordingly.

Done

ABOSTM (approved):
@olikraus thanks.

PR #1663, opened 2022-02-22
Wire: implement STOP sequence to recover I2C bus.

### LINKED ISSUE #1661: How to avoid I2C bus stuck after a board reset
I noticed that sometimes if you reset the board during an I2C transaction the I2C bus remains stuck.
A possible solution that I found is to send a bit-banged STOP sequence before the I2C begin. Maybe it would be better to include this procedure in the Wire begin implementation or in a dedicated Wire API to improve the stability of the library.
You can find below the bit-banged STOP sequence that normally I send to unlock the I2C bus before the I2C begin:

```
pinMode(SCL_PIN, OUTPUT);
pinMode(SDA_PIN, OUTPUT);

for (int i = 0; i<10; i++) {
  digitalWrite(SDA_PIN, LOW);
  delay(3);
  digitalWrite(SCL_PIN, HIGH);
  delay(3);
  digitalWrite(SDA_PIN, HIGH);
  delay(3);
  digitalWrite(SCL_PIN, LOW);
  delay(3);
}

pinMode(SDA_PIN, INPUT);
pinMode(SCL_PIN, INPUT);
``` 

Best Regards,
Carlo

ABOSTM:
**Summary**
Wire: implement STOP sequence to recover I2C bus.

Useful in case of bus stuck after a reset for example
Fixes #1661

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -83,6 +83,8 @@ void TwoWire::begin(uint8_t address, bool generalCall)
 
   _i2c.generalCall = (generalCall == true) ? 1 : 0;
 
+  recoverBus(); // in case I2C bus (device) is stuck after a reset for example
+
   i2c_custom_init(&_i2c, 100000, I2C_ADDRESSINGMODE_7BIT, ownAddress);
 
   if (_i2c.isMaster == 0) {
@@ -498,6 +500,28 @@ inline void TwoWire::resetTxBuffer(void)
 {
   if (txBuffer != nullptr) {
     memset(txBuffer, 0, txBufferAllocated);
+  }
+}
+
+// Send clear bus (clock pulse) sequence to recover bus.
+// Useful in case of bus stuck after a reset for example
+// a mix implementation of Clear Bus from
+// https://www.nxp.com/docs/en/user-guide/UM10204.pdf
+// https://bits4device.wordpress.com/2017/07/28/i2c-bus-recovery/
+void TwoWire::recoverBus(void)
+{
+  pinMode(pinNametoDigitalPin(_i2c.sda), INPUT);
+
+  if (digitalReadFast(_i2c.sda) == LOW) {
+    pinMode(pinNametoDigitalPin(_i2c.scl), OUTPUT);
+
+    for (int i = 0; i < 20; i++) {
+      digitalWriteFast(_i2c.scl, LOW);
+      delayMicroseconds(10);
+      digitalWriteFast(_i2c.scl, HIGH);
+      delayMicroseconds(10);
+    }
+    pinMode(pinNametoDigitalPin(_i2c.scl), INPUT);
   }
 }
 
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -66,6 +66,7 @@ class TwoWire : public Stream {
 
     void resetRxBuffer(void);
     void resetTxBuffer(void);
+    void recoverBus(void);
 
   public:
     TwoWire();


libraries/Wire/src/Wire.cpp
```diff
@@ -501,6 +503,27 @@ inline void TwoWire::resetTxBuffer(void)
...
+  pinMode(pinNametoDigitalPin(_i2c.scl), OUTPUT);
+
+  for (int i = 0; i < 10; i++) {
+    digitalWrite(pinNametoDigitalPin(_i2c.sda), LOW);
+    delay(3);
+    digitalWrite(pinNametoDigitalPin(_i2c.scl), HIGH);
+    delay(3);
+    digitalWrite(pinNametoDigitalPin(_i2c.sda), HIGH);
+    delay(3);
+    digitalWrite(pinNametoDigitalPin(_i2c.scl), LOW);
```
fpistm, line 520:
```suggestion
    digitalWriteFast(_i2c.sda, LOW);
    delay(3);
    digitalWriteFast(_i2c.scl, HIGH);
    delay(3);
    digitalWriteFast(_i2c.sda, HIGH);
    delay(3);
    digitalWriteFast(_i2c.scl, LOW);
```

fpistm (changes requested):

cparata (approved):

libraries/Wire/src/Wire.cpp
ABOSTM: Yes, it it is better. Done

uzi18: @ABOSTM  maybe add also isStuck() method ?

fpistm (approved):

libraries/Wire/src/Wire.cpp
fpistm: > 

What did you mean?
Check if SDA held low ?

uzi18: @fpistm exactly 👍

fpistm: Here the goal is to force bus recovery before Wire instance init. The main use case is when the board is reset during an I2C transfer. The method is not public.

ABOSTM:
Reworked implementation:
Recover only when SDA is low.
Clear bus (CLK pulses) inspired from
https://www.nxp.com/docs/en/user-guide/UM10204.pdf
https://bits4device.wordpress.com/2017/07/28/i2c-bus-recovery/

fpistm (approved):
LGTM

cparata (approved):

cparata:
Hi @ABOSTM ,
I tried the fix. It improves the performances, but sometimes I still get the bus stuck. My setup is with F401RE, IKS02A1 and X_NUCLEO_IKS02A1_ISM330DHCX_FIFO example. Putting a cycle with 20 loops and a delay of 10 us seems improving the stability on my side.
Best Regards,
Carlo

```
void TwoWire::recoverBus(void)
{
  pinMode(pinNametoDigitalPin(_i2c.sda), INPUT);

  if (digitalReadFast(_i2c.sda) == LOW) {
    pinMode(pinNametoDigitalPin(_i2c.scl), OUTPUT);

    for (int i = 0; i < 20; i++) {
      digitalWriteFast(_i2c.scl, LOW);
      delayMicroseconds(10);
      digitalWriteFast(_i2c.scl, HIGH);
      delayMicroseconds(10);
    }
    pinMode(pinNametoDigitalPin(_i2c.scl), INPUT);
  }
}
```

ABOSTM:
Hi @cparata,
This is strange as I used the same example, ISM330DHCX_FIFO, but on a B-U585I-IOT02A.
And could not reproduce this issue after more than 100 reset.
But I trust you and I am fine with your proposal.

PR #1674, opened 2022-03-09
feat: Add STM32Cube HAL getter to SPI, Wire and HardwareSerial

### LINKED ISSUE #1035: request: Implement getters for internal HAL handle structures or make it public
For some hardware advanced features (like DMA) it's impossible to combine core code and pure HAL/LL calls cuz internal handles are hidden inside core code and used only internally.
For example, for my current project i need to redefine serial rx handler to do some logic without polling, so i have to edit core to access serial_t.
Or, for example, for SPI DMA implementing i need to fetch internal spi init handle to link it with DMA, ive just moved spi handle to public section, but it's a rough way and also requires modifying core code and make core updating task more complex.

What do you think?

ABOSTM:
**Summary**
feat: Add STM32Cube HAL getter to SPI, Wire and HardwareSerial

Could be used for example to mix Arduino API and STM32Cube HAL API.
For example to configure DMA.
Warning: Use at your own risk

Fixes #1035

### DIFF
--- a/cores/arduino/HardwareSerial.h
+++ b/cores/arduino/HardwareSerial.h
@@ -174,6 +174,13 @@ class HardwareSerial : public Stream {
     // Interrupt handlers
     static void _rx_complete_irq(serial_t *obj);
     static int _tx_complete_irq(serial_t *obj);
+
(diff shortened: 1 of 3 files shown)


uzi18:
Thx

fpistm (approved):

PR #1687, opened 2022-04-08
system(F4) update STM32F4xx HAL Drivers to v1.8.0 and CMSIS Drivers to v2.6.8

ABOSTM:
system(F4) update STM32F4xx HAL Drivers to v1.8.0
system(F4): update STM32F4xx CMSIS Drivers to v2.6.8


fpistm (approved):
LGTM

HAL ETH has been fully reworked and introduce compatibility break. A new issue should be raised on STM32Ethernet library to manage this. With this PR F4 Ethernet is no more functional with the lib.

PR #1688, opened 2022-04-08
system(G0) update STM32G0xx HAL Drivers to v1.4.4 and CMSIS Drivers to v1.4.2

ABOSTM:
system(G0) update STM32G0xx HAL Drivers to v1.4.4
system(G0): update STM32G0xx CMSIS Drivers to v1.4.2


fpistm (approved):
LGTM

PR #1691, opened 2022-04-15
wire: add fix for toggling NoStretch

ric96:
On a few devboards running linux, dragonboard410c, pi4 etc, I am
seeing issues with i2cdetect and other i2c bash utilities where
the stm32 holds the SCL line low and hangs the bus. This seems
to fix the issue.

Signed-off-by: Sahaj Sarup <sahaj.sarup@linaro.org>


ABOSTM (approved):
@ric96 ,
Thanks for this PR.
LGTM

PR #1697, opened 2022-04-21
core: fix HardwareSerial compilation failed when USART disabled

### LINKED ISSUE #1695: Unable to build with Serial disabled.
Since https://github.com/stm32duino/Arduino_Core_STM32/pull/1674 a regression has been introduce.
Disabling Serial raised a build error, getter API have to be protected to avoid this.

Other getter have to be tested without HAL module support or with HAL module only.

ABOSTM:
core: fix HardwareSerial compilation failed when USART disabled

Fixes #1695


fpistm (approved):
LGTM. Thanks

PR #1708, opened 2022-05-04
Cleaning the build process

massonal:
**Summary**

This PR implements the following improvements:
* Cleaning the build flags: some flags in the compile step act on linking, and are therefore useless.
* Reducing dependencies: a couple of files in the core depend on symbols defined in SrcWrapper. Moving  them there makes for a simpler dependency graph (breaks a cycle of dependencies).

None of these changes fixes a bug, nor do they introduce new features. They just make for a simpler, cleaner, (faster?) build process.


fpistm (commented):
LGTM, tests will be done with external libraries (STM32Ethernet,...)
Waiting @ABOSTM

ABOSTM (approved):
LGTM

fpistm:
Tested with some libraries. All seems correct.

PR #1715, opened 2022-05-12
Add support to CRC peripheral by default

cparata:
Hello,
I would like to add the CRC peripheral support by default in the Core. It could be useful to run some binary libraries like MotionFX.
Best Regards,
Carlo


libraries/SrcWrapper/src/stm32/hw_config.c
```diff
@@ -19,6 +19,8 @@
 extern "C" {
 #endif
 
+CRC_HandleTypeDef hcrc = {.Instance = CRC};
```
fpistm, line 22:
```suggestion
CRC_HandleTypeDef hcrc = {
#ifdef STM32MP1xx
  .Instance = CRC2
#else
  .Instance = CRC
#endif
};
```

fpistm (changes requested):

libraries/SrcWrapper/src/stm32/hw_config.c
fpistm:
```suggestion
#ifdef CRC2_BASE
  #define CRC_INSTANCE CRC2
#elif defined(CRC_BASE)
  #define CRC_INSTANCE CRC
#else
  #error "No CRC instance available"
#endif
CRC_HandleTypeDef hcrc = {.Instance = CRC_INSTANCE};
```

libraries/SrcWrapper/src/stm32/hw_config.c
```diff
@@ -53,6 +55,9 @@ void hw_config_init(void)
   /* Configure the system clock */
   SystemClock_Config();
 
+  /* Initialize the CRC */
+  HAL_CRC_Init(&hcrc);
```
fpistm, line 59:
```suggestion
#if defined(CRC_BASE)
  HAL_CRC_Init(&hcrc);
#endif
```

libraries/SrcWrapper/src/stm32/hw_config.c
fpistm:
```suggestion
#if defined(CRC_BASE)
CRC_HandleTypeDef hcrc = {.Instance = CRC};
#endif
```

libraries/SrcWrapper/src/stm32/clock.c
```diff
@@ -173,6 +173,11 @@ void configIPClock(void)
   /* Enable SYSCFG clock, needed for example: Pin remap or Analog switch ... */
   __HAL_RCC_SYSCFG_CLK_ENABLE();
 #endif
+
+#if defined(__HAL_RCC_CRC_CLK_ENABLE)
```
fpistm, line 177:
```suggestion
#if defined(CRC_BASE) && defined(__HAL_RCC_CRC_CLK_ENABLE)
```

fpistm (changes requested):

PR #1721, opened 2022-05-18
HardwareTimer: call refresh() after parameter update when timer not running

ABOSTM:
**Summary**
HardwareTimer: call refresh() after parameter update when timer not running

In case timer is not running,
after an update of parameters (setOverflow(), setCaptureCompare(), setPrescaleFactor()),
call refresh() to force register update independently of Preload setting.

Fixes https://www.stm32duino.com/viewtopic.php?f=7&t=1563


fpistm (approved):
LGTM

PR #1729, opened 2022-05-28
Improve support for hand-written asm source files

ivankravets:
Subj.


fpistm (approved):
Assuming it is correct.

ivankravets:
Thanks 🙏

PR #1743, opened 2022-06-20
Arduino lint

fpistm:
This PR:
- fixes several warnings and errors raised by [Arduino-lint](https://github.com/arduino/arduino-lint) tool.
- deploy a new CI workflow to check no Arduino rules are broken.
- properly set fpu and float-abi for precompiled binaries usage:    
    https://arduino.github.io/arduino-cli/0.21/library-specification/#precompiled-binaries


fpistm:
Hi @valeros 

Could you help me on this issue with pio build:
https://github.com/stm32duino/Arduino_Core_STM32/runs/6964033012?check_suite_focus=true#step:3:8213

This PR adds some examples. One required to enable USB HID options.
How can we define it or simply skip it if not applicable?

valeros:
Hi @fpistm ! I see here several options:
- Update `CI/build/platformio-builder.py` in a way that it will dynamically detect whether an example requires additional USB flags/macros (e.g. based on example name).
- Introduce a special file that will indicate that this example shouldn't be compiled. Something like it's done in the Arduino core for ESP32 [here](https://github.com/espressif/arduino-esp32/blob/master/.github/scripts/install-platformio-esp32.sh#L86).
- Simply maintain a list of skipped examples or keywords in `CI/build/platformio-builder.py`. 

Does anything from the above make sense?

PR #1746, opened 2022-06-20
Add functionality to ignore library examples in CI for PlatformIO

valeros:
By default `Keyboard` and `Mouse` libraries are skipped as they may require additional configuration steps

Related #1743


fpistm (approved):
LGTM. Thank you a lot @valeros for your quick feedback and solution.

PR #1750, opened 2022-06-24
Print and debug improvements

matthijskooijman:
**Summary**

This PR makes some changes to the printing/printf code, both in the Print class and in `core_debug`.

The main change that I needed is to add a "v" version of the `core_debug()` function. Like `vprintf` vs `printf`, the `vcore_debug` function does not accept individual arguments to interpolate into format string, but a `va_list` instance. This allows it to be called from another function that is itself a varargs function and forwards its variable arguments.

I originally thought I needed this for `Print::printf`, so I implemented `Print::vprintf` too. I ended up using `core_debug()` instead, but the `vprintf()` implementation might be useful for others (and adds zero code space if unused), so I just added it.

Finally, when looking at the `core_debug()` code, I noticed the function was `static` and declared in a header file, which causes it to potentially take up a lot more flash space than needed, so I fixed that as well. See the commit message for details.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -277,6 +277,17 @@ int Print::printf(const __FlashStringHelper *format, ...)
   return retval;
 }
 
+int Print::vprintf(const char *format, va_list ap)
+{
+  return vdprintf((int)this, format, ap);
+}
+
+int Print::vprintf(const __FlashStringHelper *format, va_list ap)
+{
+  return vdprintf((int)this, (const char *)format, ap);
+}
+
+
 // Private Methods /////////////////////////////////////////////////////////////
 
 size_t Print::printNumber(unsigned long n, uint8_t base)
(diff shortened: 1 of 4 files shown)


fpistm (changes requested):
Hi @matthijskooijman 

Thanks for the PR.
You have to move this include outside the `#if !defined(NDEBUG)`
https://github.com/stm32duino/Arduino_Core_STM32/blob/e3f4ee44344866dc5442f653e383a7da14e4db10/cores/arduino/core_debug.h#L5

else it does not built properly when Core logs are not enabled 😉 

And please fix astyle issue.

matthijskooijman:
Good call, should have tested that. Fixed (and tested ;-p) now.

fpistm (approved):
Thanks @matthijskooijman for this PR, it LGTM.
About the failed CI, it is not linked to this PR but to an update of the arduino-lint which raised new error/warning. This is fixed thanks #1754.

matthijskooijman:
Thanks!

PR #1764, opened 2022-07-15
System(g0) update stm32 G0xx HAL Drivers & CMSIS

TDhaouST:
- system(G0): update STM32G0xx CMSIS Drivers to v1.4.3
- system(G0) update STM32G0xx HAL Drivers to v1.4.5


fpistm (approved):
LGTM

PR #1776, opened 2022-07-28
Implementing an alternative build framework with CMake

massonal:
## Implementing an alternative build framework with CMake

This PR implements an alternative build engine to Arduino IDE or arduino-cli.
Although widely compatible with these tools, a CMake implementation is much faster (especially on Windows, especially behind a proxy/firewall): in the order of 10sec, instead of up to a minute with the IDE on my machine. Other benefits include more flexibility in describing the build, and enhanced support of incremental compilation, including when making changes to the core.

Most features of Arduino IDE are reimplemented here; the most obvious missing feature is the lack of automatic dependency management (libraries). Indeed, the way Arduino implements can not be mirrored cleanly with CMake.

## How to use:
Examples can be found at [https://github.com/massonal/STM32CMake_workspace](https://github.com/massonal/STM32CMake_workspace).
Briefly, there needs to be a CMakeLists.txt present in the sketch, along the sources. The recommended template makes heavy use of wrapper functions designed to mirror the Arduino automation; these are implemented in the `/cmake` folder here.
Please read [the wiki I started writing on my fork](https://github.com/massonal/Arduino_Core_STM32/wiki) for more details.

## How to maintain
* Arduino doesn't "see" CMake files, so this PR overall should have no effect on the Arduino tools.
* Most CMake files are autogenerated from a template; [this script](https://github.com/massonal/Arduino_Core_STM32/blob/cmake_dev/cmake/scripts/cmake_updater_hook.py) updates them all. I advise running it automatically as a git hook after this PR gets merged.


README_CMAKE.md
```diff
@@ -0,0 +1,53 @@
+You can now use CMake to build Arduino sketches with this core.
```
fpistm, line 1: CMake can now be used....

Passive form 😉

fpistm (changes requested):
Please use passive form as far as possible

README_CMAKE.md
massonal: Fixed!

Thanks for the feedback.

cmake/scripts/cmake_easy_setup.py
```diff
@@ -0,0 +1,159 @@
...
+    please review the generated CMakeLists.txt to remove the placeholder value before calling `cmake`.
+    """)
+
+if shargs.cli is None :
+    print("""
+    Error: `arduino-cli` not found in $PATH.
+    Please install arduino-cli and make it available from your system $PATH,
+    or give its location through the `--cli` argument.
+    """)
+    exit(-1)
```
fpistm, line 38: Will be never called as argparse will raised automatically an error

cmake/scripts/cmake_easy_setup.py
```diff
@@ -0,0 +1,159 @@
...
+        boardname=shargs.board,
+    ))
+
+
+print("Generated", shargs.output or shargs.sketch/"CMakeLists.txt")
+print("""
+Unless you are building a very simple sketch with no library (e.g., Blink),
+you are advised to edit this file to fill in any missing dependency relationship.
+For details, please refer to
+https://github.com/massonal/Arduino_Core_STM32/wiki/Arduino-(in)compatibility#library-management
```
fpistm, line 151: Not updated

cmake/scripts/cmake_easy_setup.py
massonal: * about `shargs.sketch and not shargs.board` : both are optional arguments, so the condition _can_ fire.
Try running `python3 ./cmake/scripts/cmake_easy_setup.py -s ~/1_Blink/`...

* about `shargs.cli is None` : the default value is `shutil.which("arduino-cli")`, and that can be None: https://docs.python.org/3/library/shutil.html#shutil.which.

cmake/scripts/cmake_easy_setup.py
massonal: You're right!
fixed in 5a60ce5

fpistm (changes requested):
It misses .elf extension to the build file.

fpistm (approved):
Thanks @massonal  for this PR.
As a first implementation it is a good jobs.
Ready to merge and deploy.

PR #1781, opened 2022-08-04
Added some more STM32G0*** generic boards

alextrical:
IMPORTANT: Please review the [CONTRIBUTING.md](../CONTRIBUTING.md) file for detailed contributing guidelines.

**Summary**
Added extra board definitions for the following board families:
STM32G0B1CB
STM32G0B1CC
STM32G0B1CE
STM32G0B1KB
STM32G0B1KC
STM32G0B1KE
STM32G0B1MB
STM32G0B1MC
STM32G0B1ME
STM32G0C1MC
STM32G0C1ME
STM32G0B1NE
STM32G0C1NE
STM32G0B1VB
STM32G0B1VC
STM32G0B1VE
STM32G0C1VB
STM32G0C1VC
STM32G0C1VE
STM32G030JM
STM32G031C4
STM32G031C6
STM32G031C8
STM32G031G4
STM32G031G6
STM32G031G8
STM32G041C4
STM32G041C6
STM32G041C8
STM32G041G4
STM32G041G6
STM32G041C8
STM32G050C6
STM32G050C8
STM32G050F6
STM32G050K6
STM32G050K8
STM32G051C6
STM32G051C8
STM32G061C6
STM32G061C8
STM32G051F6
STM32G051F8
STM32G061F6
STM32G061F8
STM32G051G6
STM32G051G8
STM32G061G6
STM32G061G8


ABOSTM (changes requested):
@alextrical,
Thanks for this Pull Request that is is lot of new supported generic MCU.
I have some concerns:
* It looks like you rename some directories, like for example
from         G051F6P_G051F8(P-Y)_G061F6P_G061F8(P-Y)
to              G051F(6-8)(P-Y)_G061F(6-8)(P-Y)
This is not possible (at least not this way), because it is automatically done by script https://github.com/stm32duino/Arduino_Core_STM32/blob/main/CI/update/stm32variant.py
Also, your proposed name suggests that chip  G051F6Y exist, but it is not the case.

* Generally when adding new generic variant, we also update SystemClock_Config()  in generic_clock.c: https://github.com/stm32duino/wiki/wiki/Add-a-new-variant-%28board%29#3---generic-system-clock-configuration
so that instead of using reset clock configuration (which is slow), we use PLL with the max frequency supported, and with USB clock 48MHz (if USB supported, so that clock is ready to support USB)

PR #1787, opened 2022-08-08
fix: variant Nucleo F031K6: switch PLL source to HSI

ABOSTM:
fix: variant Nucleo F031K6: switch PLL source to HSI

On Nucleo F031K6, the HSE bypass through MCO from STLink
is not configured by default
(It would need SolderBridge configuration)
Thus use HSI by default.


fpistm (approved):

massonal (commented):
This matches what we discussed on the side: looks all good to me!

PR #1835, opened 2022-09-27
I2C slave: allow function wrapped callbacks

### LINKED ISSUE #1833: Wire add support for callback_function_t callbacks

The Wire methods onReceive and onRequest dont allow to pass a function via std::bind.
This would result to have to use a singleton to have access the hanlders if the I2C client code is an object like a library.

A solution similar to the attachInterrupt would solve the issue.

Something like this (similar to what we do with `attachInterrupt()`)

```cpp
   _i2bus->begin(address + baseAddress);
   _i2bus->onReceive(std::bind(&I2CClient::onReceive, this));
   _i2bus->onRequest(std::bind(&I2CClient::onRequest, this));
```
My current workaround is to have multiple definitions of the static handling methods
each pair call the right object handling methods.

This is not the best solution with `callback_function_t` I could just pass the Wire object to class instanceand it will self manage.

Another solution would be to pass the object pointer to the registration method.
This approach is used by PacketSerial:
https://github.com/bakercp/PacketSerial/blob/master/src/PacketSerial.h#L372

hitech95:
**Summary**

* [x] #1833

This allow to have TwoWire callbacks methods inside classes. You can use std::bind to wrap the callback functions.

This is an example of a Class that completely handle the I2C comunication without having the user to manually implement the callbacks:

```cpp
/*
 * Lib Functions
 */

I2Client::I2Client(TwoWire *bus)
{
    this->_bus = bus;
}

bool I2Client::begin(uint8_t baseAddress)
{
    _bus->begin(baseAddress);
    _bus->onRequest(std::bind(&I2Client::onHandleRequest, this));

    return true;
}

void I2Client::onHandleRequest()
{
    _bus->write("hello\n");
}

/*
 * Library Usage (main.cpp)
 */

#include <Arduino.h>
#include <Wire.h>

TwoWire bus(PB9, PB8);
I2Client client(&bus);

 void setup()
{
    client.begin(0x4E); // join i2c bus on address 0x4E
}

void loop()
{
}
```

**Validation**

* Ensure CI build is passed.

**Code formatting**

**Closing issues**

Closes #1833

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -438,13 +438,13 @@ void TwoWire::onRequestService(i2c_t *obj)
 }
 
 // sets function called on slave write
-void TwoWire::onReceive(void (*function)(int))
+void TwoWire::onReceive(cb_function_receive_t function)
 {
   user_onReceive = function;
 }
 
 // sets function called on slave read
-void TwoWire::onRequest(void (*function)(void))
+void TwoWire::onRequest(cb_function_request_t function)
 {
   user_onRequest = function;
 }
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -22,6 +22,8 @@
 #ifndef TwoWire_h
 #define TwoWire_h
 
+#include <functional>
+
 #include "Stream.h"
 #include "Arduino.h"
 extern "C" {
@@ -40,6 +42,10 @@ extern "C" {
 #define WIRE_HAS_END 1
 
 class TwoWire : public Stream {
+  public:
+    typedef std::function<void(int)> cb_function_receive_t;
+    typedef std::function<void(void)> cb_function_request_t;
+
   private:
     uint8_t *rxBuffer;
     uint16_t rxBufferAllocated;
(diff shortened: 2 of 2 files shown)


hitech95:
@fpistm I was thinking the same, I've used the `attachInterrupt` as a reference.
If you think that the failover to void pointers is not necessary I can drop them.

Today I'll do more tests.

fpistm:
> @fpistm I was thinking the same, I've used the `attachInterrupt` as a reference. If you think that the failover to void pointers is not necessary I can drop them.
> 
> Today I'll do more tests.

Yes it is probably the case but requires to be be tested 😉

hitech95:
I've tested with a STM32G030C8T and a MNUCLEO F401 and both works.
(One as a slave and the other as a master and the other way round)

fpistm:
> I've tested with a STM32G030C8T and a NUCLEO F401 and both works. (One as a slave and the other as a master and the other way round)

So you have tested with the c declaration removed ? if it works could you update the PR.

hitech95:
Not yet, I'm gonna do it tomorrow.
I've been quite busy with work.

fpistm:
Hi @hitech95 
any update on this?

hitech95:
Hi
sorry I had lot of issues at work.
I'm gonna do some tests this weekend!

fpistm:
No worry. I understand perfectly  😉

hitech95:
I've tested with the `slave_receive` and `slave_sender` examples from the basic arduino enviroment.
They works also by removing the C failbacks.

So I'm pushing a simplified patch for that.

Tested on Nucleo STM32 F401RE

fpistm (approved):
LGTM.
Tested on several series.

Thank you @hitech95

PR #1836, opened 2022-09-28
fix: U575Z(G-I)TxQ_U585ZITxQ ram issue

### LINKED ISSUE #1827: NUCLEO-U575ZI-Q memory size incorrect in Arduino IDE showing 256 kbytes should be 786 kbytes
Incorrect dynamic memory size being used by Arduino IDE for board Using NUCLEO-U575ZI-Q
IDE is showing - Maximum as 256 kbytes where device actually has **786 Kbyte**s of SRAM

Using NUCLEO-U575ZI-Q 
Arduino IDE 1.8.19
Core 2.3.0

example report from 'blink' sketch

Sketch uses 19820 bytes (0%) of program storage space. Maximum is 2097152 bytes.
Global variables use 1016 bytes (0%) of dynamic memory, leaving 261128 bytes for local variables. Maximum is 262144 bytes.

thank you
Joe

fpistm:
The xml describing those mcu give 256Kb ram.
Issue raised internally to fix this then it will be corrected in generated files.

Based on RM0456:
image

SRAM 1/2/3 are consecutive but SRAM4 not.

Fixes #1827


ABOSTM (approved):
LGTM

PR #1839, opened 2022-09-29
Add SUBGHZSPI support to SPI library.

fpistm:
This PR supersedes #1803.

It adds `DEBUG_SUBGHZSPI_*` pins support.
To ease usage some alias have been added:
```C
// Alias
#ifndef DEBUG_SUBGHZSPI_MOSI
  #define DEBUG_SUBGHZSPI_MOSI        PA7_ALT1
#endif
#ifndef DEBUG_SUBGHZSPI_MISO
  #define DEBUG_SUBGHZSPI_MISO        PA6_ALT1
#endif
#ifndef DEBUG_SUBGHZSPI_SCLK
  #define DEBUG_SUBGHZSPI_SCLK        PA5_ALT1
#endif
#ifndef DEBUG_SUBGHZSPI_SS
  #define DEBUG_SUBGHZSPI_SS          PA4_ALT1
#endif
```

New class created:
`class SUBGHZSPIClass : public SPIClass `
By default debug pins are not configured, only internal SPI are configured.
To enable the debug pins simply call:
`SubGHZ_SPI.enableDebugPins();`

Then you can connect an analyzer to see signals. Example:
image

Send "Hello World!" in ascii (48 65 6c 6c 6f 20 57 6f 72 6c 64 21).

/cc @matthijskooijman @LMESTM


ABOSTM (approved):
LGTM

matthijskooijman:
Looks ok to me. I had not realized that actual pins were involved (before I thought that some virtual dummy pins were added just to signal to the SPI code that SUBGHZ_SPI was to be used), but now I see that there are actual pins attached to the SUBGHZ_SPI module. IICU these are not regular SPI pins (in particular MISO seems to be output-only), so I do not think SUBGHZ_SPI can be used to drive an external SPI slave, right?

I do wonder if the API offered now (passing these debug pins to select SUBGHZ_SPI, but by default not actually using those pins but only use them when calling `debug_SubGHzSPI_pins`) is a bit unexpected, but it does seem pragmatic. One alternative I could imagine is to add an SPI constructor that just accepts the `SUBGHZ_SPI` pointer instead of pin numbers (and passes that on to `spi_init` by presetting `_spi->spi` and keeping the pin members of `_spi` set to `NC`), then it is a bit clearer that no pins are involved (and then to configure the debug out pins, you could pass the pins as normal maybe, even also support passing a subset). Not sure if that would be much better or worth the work, though.

fpistm:
Hi @matthijskooijman 

This PR is not ready as I want made some tests.
You are right about the pins they are for debug purpose only so should be only output.

lyusupov:
@matthijskooijman 

Below is my 'gist' on how to make the SUBGHZSPI radio working with [_basicmac_](https://github.com/LacunaSpace/basicmac) library:

```
#ifdef HAL_SUBGHZ_MODULE_ENABLED

#define SUBGHZ_DEFAULT_TIMEOUT     100U    /* HAL Timeout in ms               */
/* SystemCoreClock dividers. Corresponding to time execution of while loop.   */
#define SUBGHZ_DEFAULT_LOOP_TIME   ((SystemCoreClock*28U)>>19U)

static SUBGHZ_HandleTypeDef hsubghz = {.Init = {.BaudratePrescaler =
                                               SUBGHZSPI_BAUDRATEPRESCALER_8 } };
#endif /* HAL_SUBGHZ_MODULE_ENABLED */

static void hal_spi_init () {
#ifdef HAL_SUBGHZ_MODULE_ENABLED
    HAL_SUBGHZ_Init(&hsubghz);
#endif /* HAL_SUBGHZ_MODULE_ENABLED */
}

void hal_spi_select (int on) {
    if (on)
        LL_PWR_SelectSUBGHZSPI_NSS();
    else
        LL_PWR_UnselectSUBGHZSPI_NSS();
}

// perform SPI transaction with radio
u1_t hal_spi (u1_t out) {
#ifdef HAL_SUBGHZ_MODULE_ENABLED
  HAL_StatusTypeDef status = HAL_OK;
  __IO uint32_t count;

  count = SUBGHZ_DEFAULT_TIMEOUT * SUBGHZ_DEFAULT_LOOP_TIME;

  /* Wait until TXE flag is set */
  do
  {
    if (count == 0U)
    {
      status = HAL_ERROR;
      hsubghz.ErrorCode = HAL_SUBGHZ_ERROR_TIMEOUT;
      break;
    }
    count--;
  } while (READ_BIT(SUBGHZSPI->SR, SPI_SR_TXE) != (SPI_SR_TXE));

  __IO uint8_t *spidr = ((__IO uint8_t *)&SUBGHZSPI->DR);
  *spidr = out;

  count = SUBGHZ_DEFAULT_TIMEOUT * SUBGHZ_DEFAULT_LOOP_TIME;

  /* Wait until RXNE flag is set */
  do
  {
    if (count == 0U)
    {
      status = HAL_ERROR;
      hsubghz.ErrorCode = HAL_SUBGHZ_ERROR_TIMEOUT;
      break;
    }
    count--;
  } while (READ_BIT(SUBGHZSPI->SR, SPI_SR_RXNE) != (SPI_SR_RXNE));

  return (uint8_t)(READ_REG(SUBGHZSPI->DR));
#else
  return 0;
#endif /* HAL_SUBGHZ_MODULE_ENABLED */
}
```

See [this reference](https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/libraries/arduino-basicmac/src/hal/hal.cpp) for more details.

matthijskooijman:
> Below is my 'gist' on how to make the SUBGHZSPI radio working with [basicmac](https://github.com/LacunaSpace/basicmac) library:

Hey, cool! Once this PR is merged, I'd be interested in merging your work into basicmac (where I happen to also have write access ;-P).

libraries/SPI/src/SPI.h
```diff
@@ -304,4 +310,18 @@ class SPIClass {
 
 extern SPIClass SPI;
 
+#if defined(SUBGHZSPI_BASE)
+class SUBGHZSPIClass : public SPIClass {
+  public:
+    SUBGHZSPIClass(): SPIClass{SUBGHZSPI_MOSI, SUBGHZSPI_MISO, SUBGHZSPI_SCLK, SUBGHZSPI_SS} {}
+
+    void begin(uint8_t _pin = CS_PIN_CONTROLLED_BY_USER);
+    byte transfer(uint8_t _pin, uint8_t _data, SPITransferMode _mode = SPI_LAST);
```
matthijskooijman, line 319: These functions are not virtual, so if a sketch passes the `SUBGHZSPIClass` object through `SPIClass` pointer or reference (which is a common pattern with libraries), the original methods will be called instead of these. I guess we need to make them virtual in `SPIClass` to make this work in all cases?

libraries/SPI/src/SPI.cpp
```diff
@@ -421,3 +421,41 @@ void SPIClass::detachInterrupt(void)
...
   // Should be disableInterrupt()
 }
+
+#if defined(SUBGHZSPI_BASE)
+SUBGHZSPIClass SubGHZ_SPI;
+
+void SUBGHZSPIClass::begin(uint8_t _pin)
+{
+  if (_pin != CS_PIN_CONTROLLED_BY_USER) {
+    LL_PWR_SelectSUBGHZSPI_NSS();
```
matthijskooijman, line 431: Why select this pin here? The original begin(pin) implementation just configures the pin to output high (unselected), so I guess only an unselect is needed here?

matthijskooijman: Also, the superclass actually does a `pinMode` on the passed `_pin`, I'm not sure if this is problematic? Or actully, this is probably handled because `_spi.pin_ssel` is always set by the `SUBGHZSPIClass` constructor.

Also, shouldn't this code (and also below) actually check *which* pin is passed? Now it just modifies the internal NSS pin regardless of what was passed, which might be surprising (which could show up in the theoretical usecase of using this SPI bus in debug mode to do write-only SPI transactions to an externally connected device).

libraries/SPI/src/SPI.cpp
```diff
@@ -421,3 +421,41 @@ void SPIClass::detachInterrupt(void)
...
+  }
+  SPIClass::begin(_pin);
+  if (_pin != CS_PIN_CONTROLLED_BY_USER) {
+    LL_PWR_UnselectSUBGHZSPI_NSS();
+  }
+}
+
+byte SUBGHZSPIClass::transfer(uint8_t _pin, uint8_t _data, SPITransferMode _mode)
```
matthijskooijman, line 439: This only overrides a single `transfer()` overload, but there are three more (`transfer16` plus two versions that take buffer pointers) and a `beginTransaction()` method that all control the passed pin.

I wonder if, instead of overriding these methods, it would make sense to introduce a single (virtual) `controlNss()` method or so, that is called in all these places (maybe also from `begin()` for the initial `pinMode`) and can be overridden by `SUBGHZSPIClass`?

matthijskooijman (commented):
I left some comments inline.

I'm also wondering what pin number is a sketch supposed to pass to various begin/transfer functions? Currently it can be any non-zero number, but that seems a bit broad (but if so, it should probably be at least documented in a comment on the class)?

fpistm:
> I'm also wondering what pin number is a sketch supposed to pass to various begin/transfer functions? Currently it can be any non-zero number, but that seems a bit broad (but if so, it should probably be at least documented in a comment on the class)?

It is documented here:
https://github.com/stm32duino/wiki/wiki/API#spi

Mainly:

> We give to the user 3 possibilities about the management of the CS pin:
> 
>     the CS pin is managed directly by the user code before to transfer the data (like the Arduino SPI library)
>     or the user gives the CS pin number to the library API and the library manages itself the CS pin (see example below)
>     or the user uses a hardware CS pin linked to the SPI peripheral
> 

With SubGHZSPI, 2 last cases are the same as there is no hardware CS.

libraries/SPI/src/SPI.h
fpistm: done

libraries/SPI/src/SPI.cpp
fpistm: Right for the first point.
About the second, now we call all SPIClass with CS_PIN_CONTROLLED_BY_USER .

libraries/SPI/src/SPI.cpp
fpistm: Should be ok now.

Oliv4945:
Hello @fpistm any advice on using it to communicate with the SubGHZ IP?
The result of the command `0x1D 07 40 00 00 00` should be chip status and the default LoRa sync word, ie: `0x00 00 00 00 14 24`. But I get `0xFF FF FF FF 00 00`

```cpp
#include <SPI.h>

#define SubGHZ_NOP 0x00
#define SubGHZ_READ_REGISTER 0x1D
#define SubGHZ_SIZE_READ_REGISTER 4
#define SubGHZ_REG_LR_SYNCWORD_MSB 0x07
#define SubGHZ_REG_LR_SYNCWORD_LSB 0x40
#define SubGHZ_SIZE_REG_LR_SYNCWORD 2


void setup() {
  Serial.begin(921600);
  delay(1000);
  
  SubGHZ_SPI.begin(!CS_PIN_CONTROLLED_BY_USER);
  uint8_t buffer[SubGHZ_SIZE_READ_REGISTER + SubGHZ_SIZE_REG_LR_SYNCWORD] = {SubGHZ_READ_REGISTER, SubGHZ_REG_LR_SYNCWORD_MSB, SubGHZ_REG_LR_SYNCWORD_LSB, SubGHZ_NOP, SubGHZ_NOP, SubGHZ_NOP};

  Serial.print("Command: 0x");
  for (uint8_t i; i<SubGHZ_SIZE_READ_REGISTER + SubGHZ_SIZE_REG_LR_SYNCWORD; i++) {
    Serial.printf("%.2X ", buffer[i]);
  }
  Serial.println();

  Serial.print("Result: 0x");
  SubGHZ_SPI.transfer(!CS_PIN_CONTROLLED_BY_USER, buffer, 4);
  for (int i; i<SubGHZ_SIZE_READ_REGISTER + SubGHZ_SIZE_REG_LR_SYNCWORD; i++) {
    Serial.printf("%.2X ", buffer[i]);
  }
  Serial.println();
}

void loop() {}
```
Sorry if it is a basic question, last time I used Arduino SPI was a while ago...


@matthijskooijman long time no see :) Glad that you are still playing with LoRa!

fpistm:
Hi @Oliv4945 
This PR allows to manage the SUBGHZSPI not the 'chip'. It requires to be enabled properly but it is not the goal of this PR.

matthijskooijman:
> It is documented here:
https://github.com/stm32duino/wiki/wiki/API#spi

Hm, I'm not sure I understand entirely yet, so this post is me trying to make sense of things. Note that I personally do not really care about letting the SPI library control the CS pin, since it offers less control and clarity about when the CS pin is asserted exactly (and it seems this is also inconsistent or bugged currently, see below). However, *if* we offer this feature, at least it should be consistent.

> 1. the CS pin is managed directly by the user code before to transfer the data (like the Arduino SPI library)

In regular SPI library, this is done by *not* passing a CS pin to the constructor, and then not passing any pin numbers to the begin/begintransaction/transfer functions (or passing `CS_CONTROLLED_BY_USER`).

> 2. or the user gives the CS pin number to the library API and the library manages itself the CS pin (see example below)

In regular SPI library, this is done by *not* passing a CS pin to the constructor, and *passing* the pin to be controlled to begin/begintransaction/transfer functions.

> 3. or the user uses a hardware CS pin linked to the SPI peripheral

In regular SPI library, this is done by *passing* a CS pin to the constructor. The pin value passed to begin/begintransaction/transfer is ignored.


For SubGHZ_SPI, case 1 (from the user perspective) ends up passing `DEBUG_SUBGHZSPI_NSS` to the regular SPI library, which makes it look like case 3, except that some subghz-specific exceptions in `spi_init` prevent pinmuxing the pins and disable SPI hardware NSS control.

For SubGHZ_SPI, case 2 (from the user perspective), this ends up controlling the NSS virtual pin, regardless of which pin is passed. This seems a bit confusing to me, it would make sense if this only happened when passing `DEBUG_SUBGHZSPI_NSS` or maybe some other value?

For SubGHZ_SPI, case 3 (from the user perspective) ends up behaving almost exactly like case 1 (since there `DEBUG_SUBGHZSPI_NSS` is also passed to the plain SPI library, and the same exceptions in `spi_init` will trigger), except that the pin value passed to begin/begintransaction/transfer is not ignored but instead causes the internal NSS signal to be controlled.


It seems this interface is a bit confusing, though I'm not sure what would be better. From the user perspective, I guess there's basically two cases: Where the use controls the internal NSS signal manually (which works fine as case 1 now), and where the SPI library automatically controls it (which I guess feels more like case 3 than case 2, except that I think the timing of NSS switches is more like case 2 than case 3). Maybe the API to trigger automatic NSS control  should be neither case 2 or 3, but be some SubGHZ_SPI-specific API?

I also wonder if it makes sense to have a SubGHZ_SPI constructor that accepts pins at all, though. Normally, the pins passed are passed through the pin maps to be checked, so you can only pass pins that are actually connected to the SPI block. However, for SubGHZ_SPI, this pin map is skipped (because of the `spi_init` exception), so any pins passed will work (in the sense that the SPI block controls the internal SPI signals, not the pins passed). Only when calling `enableDebugPins()` will the pins be applied to the pinmap (but any failures by incorrect pins are, I think, ignored). I guess this constructor could be useful when using `enableDebugPins()` to select to which pins the debug signals should be mapped exactly, except that they can only be mapped in exactly one place, so this does not add anything.

So maybe the SUBGHZSPIClass constructor with arguments can just be removed? This also makes (maybe) it a little more clear that case 3 (hardware control of NSS) is not supported (though I guess directly controlling the internal NSS signal might be more similar to case 3 than case 2, maybe...).

### NSS hardware control bugged?

This is probably off-topic, but I noticed this in the code and want to at least mention this. When using NSS hardware control (case 3), according to the reference manual the NSS line is asserted when the SPI unit is enabled (`SPE` is set). Looking at the code, this is done when `SPI.begin()` is called, and it remains set until `SPI.end()` is called, leaving the NSS pin asserted all the time. I would expect that NSS is only asserted between `beginTransaction()` and `endTransaction()`, but maybe this code and/or API predates the presence of `beginTransaction()` and `endTransaction()` (which were added later mostly to allow switching between settings and to allow auto-disabling interrupts during transactions, I'm not sure if we considered hardware NSS control when adding these since this happened for AVR initially) and calling `end()` between transactions is how this is intended to be used?

There also seems to be a discrepancy between case 2 and 3 in terms of timing: For case 2, the NSS pin is asserted for each transfer separately (which can be each byte, every 2 bytes for `transfer16` or complete transactions if a complete buffer is passed to `transfer`), while for case 3, the NSS pin is asserted for longer.

Finally, case 3 is now selected by the constructor, so you cannot mix that with other cases (e.g. have one slave where the NSS is controlled by hardware and another where the NSS is controlled manually), but I guess that is fine, since then you can just as well control everything manually (except when mixing libraries using the same SPI bus with different expectations, but I suspect most if not all libraries implement case 1 anyway).

matthijskooijman:
@fpistm As we discussed privately before, I've taken a stab at another class to control the SubGhz internal signals (NSS, Busy, Reset) and manage the radio interrupt. See https://github.com/matthijskooijman/Arduino_Core_STM32/commit/f661252d976ab84330675070d2e310254fe4556f (single commit on top of this PR).

That class now also includes the SPI object for the radio that was still a global in your PR.

I'm not sure where this code should live exactly - it seems largely unrelated to SPI, so the SPI library is probably not the right place. So maybe in a library of its own, then?


@Oliv4945 I suspect the radio might be in reset by default, so you might need to clear the reset first. See the `writeReset()` method in the commit linked above for this.

fpistm:
Hi @matthijskooijman 
I've updated my PR based on your comments/discussion.

Note: Default instance: `SubGHZ_SPI` removed

matthijskooijman:
@fpistm I've updated this PR, with the following changes to your code:
 - I fixed spi_init check for SUBGHZSPI, it checked the wrong variable (`handle->Instance` wasn't set until after the check). This wasn't a problem before when SUBGHZSPIClass still passed actual pin numbers to SPIClass (which spi_init could resolve to the right SPI peripheral), but now that the pin numbers are replaced with `NC` this would trigger the "ERROR: at least one SPI pin has no peripheral" message. Fixing the check resolves this. I've also added a check that all pins must be NC for SUBGHZ and slightly refactored the code around there to reuse the `handle->Init.NSS` assignment code (which results in the same behavior since `pin_ssel` is `NC`).
  - I added some `using` directives, since if you override a method in a subclass, other overloads (with different argument types) are not automatically inherited anymore. These directives make this inheritance explicit.
  
 I've put these in a fixup commit for easy review, this commit must be squashed before merging!
 
 Additionally, I've added a commit that adds my SubGhz object, now in a separate library as discussed, adding a README, keywords.txt, library.properties and an example sketch.

I've changed the pin API to use bools instead of LOW/HIGH, I think that's the (marginally) better approach after all.

I've also added an `spi_settings` object to be used with `beginTransaction()` (since that method always requires a settings object to be passed). Using this also ensures the SPI speed is as fast as possible (up to 16Mhz is supported by the radio). It would be good if the speed was this high by default for the SubGHZ SPI object, but I'm not sure what the cleanest way for this is (we could call `setClockDivider`, but then we would have to know the input clock speed and back-calculate from Mhz to a divider, or we could preset the speed in `spiSettings`, but that is currently private).

I've updated our (private) RadioLib prototype branch to use this new SubGhz library, that works well for me now. Next I'll clean up that branch and produce a proper public PR from that.

fpistm:
Thanks @matthijskooijman 

I've pushed a change to pio build to avoid issue raised by the ci. (required to rebase on top of main)
Still some typo to fix and applying astyle.

I agree for the fixup you can merge it. Pay attention to astyle for this commit.

matthijskooijman:
I've pushed again to fix astyle and codespell.

Continuous integration still fails, since this now tries to compile the SubGhz library example for non-wl boards, which fails. I guess it would be useful to include this example in the CI build, but I'm not entirely sure what the best approach is (we could run `platformio-builder.py` twice [here](https://github.com/stm32duino/Arduino_Core_STM32/blob/618610d1ebc575c7b772b94ecc9b25d4c6e3e35d/.github/actions/pio-build/entrypoint.sh#L28), once ignoring the SubGhz library and once for the STM32WL board, but maybe that makes the generated output a bit more confusing (since there will be a summary of build results half way then, I think. I haven't looked at the core_build workflow at all yet).

matthijskooijman:
Seems our comments crossed. Ignoring the SubGhz library seems ok, maybe it can be added to the test build later. :-)

I've rebased on main and merged the fixup (it was 490afe0c6965ef19c40dd249f30022b07a61d0df if you want to review it again).

fpistm:
It can be ignored now. Even if it is not existing it should not be an issue. Thanks for the quick update.

libraries/SubGhz/README.md
```diff
@@ -0,0 +1,134 @@
...
+#### "GPIO" signals
+Some of the signals which are normally (on an actual external SX126x
+radio chip) available externally and controlled by GPIO are now
+internally connected to registers in the RCC and PWR modules. These
+signals (Nss, Reset and Busy) can be controlled and/or read through this
+library.
+
+The Nss signal can be written (and read back) using
+`SubGhz.setNssActive(bool)` and `SubGhz.isNssActive()`, the Reset signal
+using `SubGhz.setResetActive(bool) and `SubGhz.isResetActive()` and the
```
Oliv4945, line 44:
```suggestion
using `SubGhz.setResetActive(bool)` and `SubGhz.isResetActive()` and the
```
Missing character "`"

libraries/SubGhz/README.md
```diff
@@ -0,0 +1,134 @@
...
+can be used to temporarily disable the interrupt. If the interrupt is
+triggered while it is disabled, it will become pending (indicated by the
+`SubGhz.isInterruptPending()` method) and the interrupt handler will run
+directly when the interrupt is enabled again (unless cleared with
+`SubGhz.clearPendingInterrupt()`). Note that there is no method to query
+whether the interrupt is currently enabled, as the interrupt controller
+hardware does allow reading back this value.
+
+Note that since these DIO signals are always connected to the interrupt
+controller, drive a single interrupt and are not available on any
```
Oliv4945, line 82: `interrupt and are not available` missing words?

libraries/SubGhz/examples/ReadRegister/ReadRegister.ino
```diff
@@ -0,0 +1,56 @@
...
+  SubGhz.SPI.transfer(0x07); // Write MSB of register address: SUBGHZ_LSYNCRH
+  SubGhz.SPI.transfer(0x40); // Write LSB of register address: SUBGHZ_LSYNCRH
+  uint8_t status = SubGhz.SPI.transfer(0x0);  // Read status
+  uint8_t value = SubGhz.SPI.transfer(0x0);  // Read register value
+
+  // End transaction
+  SubGhz.setNssActive(false);
+  SubGhz.SPI.endTransaction();
+
+  // This should print a register value 0x14 (reset value of the sync word MSB)
```
Oliv4945, line 47:
```suggestion
  // This should print a register value 0x14 (reset value of the LoRa sync word MSB)
```

Oliv4945 (commented):
Thanks @matthijskooijman the reset line did hte trick for me :)

I wrote minor changes suggestions

libraries/SubGhz/README.md
matthijskooijman: It's actually a list of three things: The signals are always connected, the signals drive a single interrupt (not three separate ones) and the signals are not available anywhere else, but I can see how this is confusing. I'll reword to  be more clear.

matthijskooijman:
@Oliv4945 Thanks for your review and suggestions, I've processed them in a new version.

@fpistm I'm done with this PR, so it can be merged as far as I'm concerned.

fpistm:
Thanks @matthijskooijman 

LGTM even if PIO is CI is broken.

PR #1848, opened 2022-10-19
Update generic_clock.c

pawkon:
Proper clock definition 96MHz and RCC_HSE_ON based on STM32CubeIDE

As discussed here: https://www.stm32duino.com/viewtopic.php?p=10988#p10988


fpistm:
Hi @pawkon 
Thanks for the PR. Anyway as it is the generic clock config, it should use the internal clock instead of HSE.
Don't know why it is not the case before, probably a wrong copy paste. All generic provide a default clock config with HSI or MSI.
I guess the Nucleo F413ZH should be added with HSE usage.

pawkon:
It is also possible to use :

`RCC_OscInitStruct.HSEState = RCC_HSE_BYPASS;`

But with other changes that I've made. It seams that mainly PLLM, PLLN, PLLP, PLLQ, PLLR were wrong

fpistm:
Hi @pawkon 
Again thanks for the PR. @gautierg-st will fix the generic clock and add the support of the Nucleo F413ZH. Fortunately we have one boards and will be able to test.

Issue #722, opened 2019-10-23
[New Variant] request list


fpistm:
Hereafter the list of all new variants requested

**Board support requested:**
 - [ ] [STM32F469iDiscovery](https://www.st.com/en/evaluation-tools/32f469idiscovery.html) (#571)
 - [ ] [STM32F407ZGT6 Mini](https://stm32-base.org/boards/STM32F407ZGT6-VCC-GND-Large) (#635)
 - [ ] [STM32F103ZET6 Mini](https://item.taobao.com/item.htm?&id=41867070778) (#635)
 - [ ] [NUCLEO-L010RB](https://www.st.com/en/evaluation-tools/nucleo-l010rb.html) (#662)
 - [ ] [NUCLEO-F746ZG](https://www.st.com/en/evaluation-tools/nucleo-f746zg.html) (#715)

**Note that user can add a STM32 based board following this [wiki](https://github.com/stm32duino/wiki/wiki/Add-a-new-variant-(board))**

Any contribution is welcome, so do not hesitate to submit a PR.

r-pv:
Hi, could you expand a bit the **Add new board** wiki with info about how to set properly the analogs?, because I'm having problems with my variant for the [F030C8](https://github.com/r-pv/rSTM32F030C8T6_ProMiniPlus/tree/master/RBOARD_F030C8T6), when I tried to read PA0 ( analogRead(PA0) ) it really reads the PA3, for the moment I solve it by using analogRead(PA_0). So I was under the impression that only has to match the:
NUM_ANALOG_FIRST on the variant.h file with the duplicates on the digitalPin array on variant.cpp, but it seems that may be more to it, right?. Thanks.

Hedda:
Request support for STM32MP1 (STM32 MP1) that comes onboard STM32MP157A-DK1 development board

https://www.st.com/en/evaluation-tools/stm32mp157a-dk1.html

STM32MP157A-DK1 developmnet board also comes in STM32MP157C-DK2 discovery kit

https://www.st.com/en/evaluation-tools/stm32mp157c-dk2.html

STM32MP157A-DK1 and STM32MP157C-DK2 Discovery kits are $99 with LCD or $69 without
https://www.st.com/en/evaluation-tools/stm32mp157a-dk1.html
https://www.st.com/en/evaluation-tools/stm32mp157c-dk2.html

STM32MP157A-EV1 Evaluation kit for PCB engineers does have a steeper price at $399
[...]

PR #1913, opened 2023-01-02
Update boards.txt for stm32f0discovery (MB1034B)

devweblab:
- Added support for STM32 Discovery board MB1034 B-00
- Added support for BMP (Generic F0)

Instructions for flashing the BMP (incl. VCP) firmware is here: https://github.com/blackmagic-debug/blackmagic/tree/main/src/platforms/stlink

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1428,6 +1428,14 @@ GenF0.menu.pnum.GENERIC_F051R4TX.build.board=GENERIC_F051R4TX
 GenF0.menu.pnum.GENERIC_F051R4TX.build.product_line=STM32F051x8
 GenF0.menu.pnum.GENERIC_F051R4TX.build.variant=STM32F0xx/F051R4T
 
+# Generic F051R8Tx
+GenF0.menu.pnum.GENERIC_F051R8TX=Generic F051R8Tx
+GenF0.menu.pnum.GENERIC_F051R8TX.upload.maximum_size=65536
+GenF0.menu.pnum.GENERIC_F051R8TX.upload.maximum_data_size=8192
+GenF0.menu.pnum.GENERIC_F051R8TX.build.board=GENERIC_F051R8TX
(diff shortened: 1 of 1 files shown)


fpistm (commented):
Hi @devweblab 
Thanks for the PR.
It seems not valid as not linker script provided.
You declare only the generic no link with the discovery which would require dedicated entry in Discovery menu and dedicated variant files which allows to defined pins ordering and also dedicated LED, user button, system clock....
For the geric it should also be fine to get a generic clock config.
You can also add in the boards.txt the generic F051R8Hx.

devweblab:
Hi @fpistm 
Ah ok, don't know why it is missing, sorry. It is working with a copy of the linker file from STM32F051R4Tx
https://github.com/stm32duino/Arduino_Core_STM32/blob/main/variants/STM32F0xx/F051R4T/ldscript.ld
Thank you.

fpistm:
PR is not valid and does not correspond to the description. No update since 2 weeks so close it.
Note: BMP upload method should not be added.

devweblab:
Closing this PR is ok, thank you.
Can you please explain me why BMP is not added for F0 boards?
The variant STM32F0xx/F051R8(H-T) is already in you repository. If possible please add the genereic F051R8Tx cpu.

fpistm:
Request added to #722
Yes the folder exists for the variant as we generate all files for all MCU? It still misses a linker script, a generic clock config and to declare it to the boards.txt to declare it.
Generic entries provide 3 default way to flash BMP is specific one, in this case if you want it you should define it on your side using boards.local.txt feature of arduino to override default one: https://arduino.github.io/arduino-cli/0.21/platform-specification/#boardslocaltxt

devweblab:
Thank you!
Note: For boards >F0 there is the BMP upload. Example for F1:
https://github.com/stm32duino/Arduino_Core_STM32/blob/main/boards.txt#L2629

PR #1941, opened 2023-02-03
fix(CRC): explicitly set InputDataFormat to avoid assert

fpistm:
when assert feature is enabled.
Default value (0) is not correct.

Issue raised on the forum:
https://www.stm32duino.com/viewtopic.php?t=1864

/CC @cparata


cparata (approved):

Bambofy:
Hi, shouldn't the #if check that the macro is not defined instead of is defined?

```
#if !defined(CRC_INPUTDATA_FORMAT_BYTES)
                          .InputDataFormat = CRC_INPUTDATA_FORMAT_BYTES
#endif
```

PR #1944, opened 2023-02-07
platform: moved from arm-none-eabi-gcc 10.3.1-2.3 to 12.2.1-1.2

fpistm:
* Release [xpack-dev-tools/arm-none-eabi-gcc-xpack v12.2.1-1.2](https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/releases/v12.2.1-1.2) from @ilg-ul (thanks to him)
* Default language standards set to gnu C/C++17 (previously gnu C/C++11)

`dev` branch of  [stm32duino/BoardManagerFiles](https://github.com/stm32duino/BoardManagerFiles) updated to references the new toolchain used by CI:
https://raw.githubusercontent.com/stm32duino/BoardManagerFiles/dev/package_stmicroelectronics_index.json

#### 32-bit support
Support for 32-bit Intel Linux and Intel Windows was dropped in 2022. Support for 32-bit Arm Linux (armv7l) will be preserved for a while, due to the large user base of 32-bit Raspberry Pi systems.

### DIFF
--- a/CI/update/templates/stm32yyxx_ll.h
+++ b/CI/update/templates/stm32yyxx_ll.h
@@ -4,6 +4,9 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-parameter"
 #pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#ifdef __cplusplus
+  #pragma GCC diagnostic ignored "-Wregister"
+#endif
 
 /* Include Low Layers drivers */
 {% for ll_header_name in ll_header_list %}
--- a/CI/update/templates/stm32yyxx_ll_ppp.h
+++ b/CI/update/templates/stm32yyxx_ll_ppp.h
@@ -4,6 +4,9 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-parameter"
 #pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#ifdef __cplusplus
+  #pragma GCC diagnostic ignored "-Wregister"
+#endif
 
 {% for serie in serieslist %}
   {% if loop.first %}
--- a/cores/arduino/avr/dtostrf.c
+++ b/cores/arduino/avr/dtostrf.c
@@ -19,6 +19,7 @@
 */
 
 #include <stdio.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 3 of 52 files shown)


ilg-ul:
If I understand your index.json file correctly, you provide support for multiple toolchain releases, including the latest 12.x.

Any reasons for not also including the 11.x? At least the latest [v11.3.1-1.1.2](https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/releases/tag/v11.3.1-1.1.2)?

ilg-ul:
Your index.json lists the following:

- "9.3.1-1.3"
- "10.2.1-1.1"
- "10.3.1-2.3"
- "12.2.1-1.2"

I confess that I don't know how this file is used, but if users can select any of the above releases, I think they should also be allowed to chose "11.3.1-1.1", even if the latest one is "12.2.1-1.2".

fpistm:
No user could not select the toolchain. The release give their tools dependencies.

ilg-ul:
> No user could not select the toolchain. 

I see. But then why are those old definitions still present in the JSON file?

fpistm:
> > No user could not select the toolchain.
> 
> I see. But then why are those old definitions still present in the JSON file?

Because all release are available. User select the core version and then tools dependencies are selected.

fpistm:
To go further, the .json provides all information to install a specifc core version.
Ex here the 2.0.0:
https://github.com/stm32duino/BoardManagerFiles/blob/b815bc0f5ae37ec571dd9f7829d060dbebdb0382/package_stmicroelectronics_index.json#L15
which provides the tools dependencies:
https://github.com/stm32duino/BoardManagerFiles/blob/b815bc0f5ae37ec571dd9f7829d060dbebdb0382/package_stmicroelectronics_index.json#L49-L65

So 2.0.0 request to have gcc version 9.3.1-1.3

For the next 2.5.0, the 12.2.1-1.2 will be required.

ilg-ul:
I'm not familiar with the Duino versioning scheme, and I don't know how much freedom you have with your releases, but, from the toolchain point of view, it is useful to have the choice to build a project with a specific version. Can you go back in time and insert an intermediate release, like 2.4.x, configured with GCC 11.3? Or make 2.5.0 with 11.3, and 2.5.1 with 12.2?

fpistm:
Honestly, I see no reason for doing this nor the interest.

PR #2044, opened 2023-06-21
update ram size for stm32f72x and f73x

gautierg-st:
On STM32F72x/F73x, RAM size is 256k and not 192k.

See comment in https://github.com/stm32duino/Arduino_Core_STM32/commit/a855f1291c5e54daf7858e1b39a171677aed04ff


fpistm (approved):
LGTM. Thanks @gautierg-st

PR #1995, opened 2023-04-05
Add STM32C0116-DK support.

honnet:
**Summary**

This PR implements the following **features**:

* [ ] Feature 1: add support for [stm32c0116-dk](https://www.st.com/en/evaluation-tools/stm32c0116-dk.html) dev board.
* [ ] Feature 2: add support for the MCU in this dev board (STM32C011x).

**Motivation**: these are new architectures / dev boards that ST seems to push for.

Let me know if there's anything that can be improved :)

PR #2048, opened 2023-06-29
variants(C0): add generic C011D6, C011F(4-6)(P-U) and C031F(4-6)P

fpistm:
Base for #1995.

/CC @honnet


honnet:
Very cool!
Now I understand better what was wrong in my [original pull request](https://github.com/stm32duino/Arduino_Core_STM32/pull/1995#pullrequestreview-1372241244), hopefully it will help me improve my future PRs, thanks a lot!

---

PS: There's a last question if I may:

I had to [remove the "WEAK" attribute](https://github.com/honnet/Arduino_Core_STM32/commit/52ff62e0edfe580c7d2db3174e2eec2f16297c09#diff-ad8a1cf5ea2623c4abaf4f9c9f74d971e223f99c1041ff5a093afa53495c832cR71-R72) for SystemClock_Config(), otherwise some of these parameters* were overwritten.

=> Could it be because it was (incorrectly) in variant_generic.cpp instead of [generic_clock.c](https://github.com/stm32duino/Arduino_Core_STM32/pull/2048/commits/a1c0debd39202afc92cf83e17f2aca0aeecbc890#diff-529455f37e6bb65bb194f54f572fab2b7a28d9cf64137749e94c30e18a0f62bfR24)?


*I think it was the Clock source or pre-scaler.

fpistm:
The function have to be WEAK to be able to overwrite it at sketch level that's all.

honnet:
OK! Thanks!

PR #2049, opened 2023-06-29
variants(C0): add STM32C0116-DK support

fpistm:
Supersede  #1995.


fpistm:
@honnet if you can give your feedback then I can merge and release. 😉

honnet:
It works!

I was a bit worried about the WEAK attribute with SystemClock_Config() but I just tested the clock timings (UART + logic analyser) on both the STM32C0116-DK and a custom STM32C011D6 board, they're good!

Thanks again!!

---

PS: It seems good but I will still think about other possible tests later, this was a [5 am hotel hack](https://photos.app.goo.gl/ooruk4FUoKXfkJTK7) :)

PR #2050, opened 2023-06-29
Add SPI slave mode support

patricklaf:
**Summary**

This PR implements SPI slave mode for the SPI library.

There is an existing PR, #1612, but it is not actively developped and it uses HAL and LL functions from user code.

**How to use**

There is no change to the API for master mode. Existing code should just work without any change.

This is an exemple of a slave device:

```cpp
volatile bool rx = false;

void SPI_ISR() {
  rx = true;
}

void setup() {
   // Add SPI_SLAVE to begin function
  SPI.begin(SS, SPI_SLAVE);
  // Add an ISR to the SS pin to detect device selection by SPI master
  SPI.attachSlaveInterrupt(SS, SPI_ISR); 
}

void loop() {
  if (rx) {
    // Use the various SPI.transfer() functions to read and write data
    rx = false;
  }
}
```

**Limitation**

- The read and write operations of the slave must be as fast as possible or data will be lost.

- An SPI device is only slave or master. `begin()` and `end()` can be called with different device mode to allow role switch.


fpistm:
Hi @patricklaf 
I've rebased and reworked your PR to match latest SPI changes.
I guess, It would be fine to add examples.

AlejandroHuerta:
Has been PR been forgotten? It would be helpful to have this feature and not need to modify the local source code.

AlejandroHuerta:
It's a bit over my head atm but I have looked into it a bit. Just need more free time.

philrittenhouse:
@patricklaf Thank you for writing this!  I've been testing it out and it seems to work well if the slave knows how many bytes to expect from the master.  In my use case it can vary, but AFAICT the spi.transfer() functions need a fixed length.  

Am I missing something?  If not, I'm thinking of adding a helper function spi_com.c that reads a variable amount of data until the CS goes high.  Let me know if you see any issue with this approach

fpistm:
> is a good first step and provides a good

OK, let's see.

PR #2072, opened 2023-07-16
fix #2071 add support for max overflow value for 32 bits timer

### LINKED ISSUE #2070: Use CDC for USB HS
The USB header files contain define USE_USB_HS. But it will not work to initialize CDC for HS, since Id = 0 is written in the CDC_Init function.

void CDC_init(void)
{
  if (!CDC_initialized) {
    /* Init Device Library */
    if (USBD_Init(&hUSBD_Device_CDC, &USBD_Desc, 0) == USBD_OK) ...

### LINKED ISSUE #2071: Setting 32 bit timers overflow value to maximum is not possible
**Describe the bug**
The function `HardwareTimer::setOverflow()` allow to set the overflow value as a `uint32_t` an then set the value of the STM32 ARR Register to: $overflow - 1$ since the function is meant to recieve as argument the overflow value and not the **terminal count**.
has shown in the function implementation below.

https://github.com/stm32duino/Arduino_Core_STM32/blob/8ff274affbbce42f1d452e74365b66ba03e66c6c/libraries/SrcWrapper/src/HardwareTimer.cpp#L521-L557

this works fine with 16 bit counters and 32 bit counters when not used in the full range.
but to set the max value of a 32 bit counter the **ARR** register value should be set to `0xFFFFFFFF` to do so the function argument should be set to `0x100000000` but since the argument is defined as `uint32_t` the value will not fit and overflow as `0`

I want to point out that setting the value to the maximum is an important thing, since not only will increase marginally the maximum counter value but more importantly allows to calculate values of thresholds and time elapsed between input capture event easly because
`elapsedTicks = capture - oldCapture`  if all variables are defined as `uint32_t` and the elapsedTicks guaranteed to be under `0x100000000` will work even if an overflow occurs.
and in the same way when generating a square wave updating the **ARR** value with the next event value   can be done as `nextVal = lastVal + delta` under the same conditions.
this methods are common practice on timer usage on any $\mu Controller$ 

**workaround**
In the actual state the problem can be solved by either
- compensating for the missing value in the period computation
  ```C++
  if(capture < captureOld) elapsedTicks = capture - oldCapture - 1;
  else elapsedTicks = capture - oldCapture;
  ```
- initializing the timer overflow by accessing directly to the STM32 HAL 
  ```C++
  timerObj_t _timerObj;
  _timerObj.handle.Instance = TIM2;
  __HAL_TIM_SET_AUTORELOAD(&_timerObj.handle, 0xFFFFFFFF);
  ```
neither of these solution are efficient nor elegant.

**Suggested Fixes**
to me the easiest way to fix this issue is changing the function definition from 
```C++
void HardwareTimer::setOverflow(uint32_t overflow, TimerFormat_t format)
``` 
to 
```C++
void HardwareTimer::setOverflow(uint64_t overflow, TimerFormat_t format)
```
by using this approach the fix should not cause any compatibility issues with existing code

**Constatation**
here you can see the serial output of the STM32 when measuring a 1hz pulse over an 80Mhz timebase
both generated by the same oscilltor so they are perfectly coherent.
all the readings are perfect except for when the overflow occurs in that case you sistematically get one count more that is due to the missing code.
```
captureOld capture    difference
4059789198 4139789198 80000000
4139789198 4219789198 80000000
4219789198 4821903    80000001
4821903    84821903   80000000
84821903   164821903  80000000
164821903  244821903  80000000
```

after applying one of the above workaround the error **never** appears  again
```
captureOld capture    difference
4106049847 4186049847 80000000
4186049847 4266049847 80000000
4266049847 51082551   80000000
51082551   131082551  80000000
131082551  211082551  80000000
211082551  291082551  80000000
```

here there is a the code on witch i've found the bug:
```C++
#include <Arduino.h>

HardwareTimer *FCT; //FCT Frequency Counter Timer
volatile uint32_t captureOld = 0, capture;
volatile uint32_t period;
volatile bool newMeas = false;

void IC_Callback(){
    capture = FCT->getCaptureCompare(LL_TIM_CHANNEL_CH1);
    Serial.print(captureOld);    //print inside an interrupt, bad practice but accettable for debugging
    Serial.print(" ");
    Serial.print(capture);
    Serial.print(" ");
    period = capture - captureOld;
    Serial.println(period);

    captureOld = capture;
    newMeas = true;
}

void setup()
{
    Serial.begin(115200);

    FCT = new HardwareTimer(TIM2);

    FCT->setMode(LL_TIM_CHANNEL_CH1,TIMER_INPUT_CAPTURE_RISING,PA_0);
    FCT->setPrescaleFactor(1);

    FCT->setOverflow(0xFFFFFFFF);//max value on 32 bit          // invert the comment on these line to passs 
    //timerObj_t _timerObj;                                     //from the bug to the workaround
    //_timerObj.handle.Instance = TIM2;                         //
    //__HAL_TIM_SET_AUTORELOAD(&_timerObj.handle, 0xFFFFFFFF);  //

    FCT->updateRegistersIfNotRunning(_timerObj.handle.Instance);
    FCT->attachInterrupt(LL_TIM_CHANNEL_CH1,IC_Callback);
    FCT->resume();

}


void loop()
{
    if (newMeas){
        //Serial.println(period);
        newMeas = false;
    }
}


// unrelated to the issue
// this was the clock configuration to use a 10Mhz extenal oscillator
// that was important to have the 1pps signal exactly coherent to the clock
// otherwise the problem can be hard to spot
// this may be replicable using a wave generated by another timer
extern "C" void SystemClock_Config(void){...}
```

**Desktop :**
 - OS: Windows 10
 - Platformio version:  core 6.1.7 on clion
 - STM32 core version: 2.6.0
 - Upload method: Default

**Board:**
 - Name:  Nucleo L476RG
 - Extra hardware used if any: squarewave generator

GianfrancoIU1JSU:
**Summary**
the following fixes #2071  in the way proposed in the issue itself.

this will now allow to set the 32bit HW timers overflow to the full value.

the `getOverflow()` function was also modified to accomodate the new value.
this will not cause problems with existing code since they are using values castable in `uint32_t`

**This PR fixes the following bug**
#2071 

**Validation**
the modification was tested and work as expected
testing code present in #2071

### DIFF
--- a/cores/arduino/HardwareTimer.h
+++ b/cores/arduino/HardwareTimer.h
@@ -113,7 +113,7 @@ class HardwareTimer {
     uint32_t getPrescaleFactor();
 
     void setOverflow(uint64_t val, TimerFormat_t format = TICK_FORMAT); // set AutoReload register depending on format provided
-    uint32_t getOverflow(TimerFormat_t format = TICK_FORMAT); // return overflow depending on format provided
+    uint64_t getOverflow(TimerFormat_t format = TICK_FORMAT); // return overflow depending on format provided
 
     void setPWM(uint32_t channel, PinName pin, uint32_t frequency, uint32_t dutycycle, callback_function_t PeriodCallback = nullptr, callback_function_t CompareCallback = nullptr); // Set all in one command freq in HZ, Duty in percentage. Including both interrupt.
     void setPWM(uint32_t channel, uint32_t pin, uint32_t frequency, uint32_t dutycycle, callback_function_t PeriodCallback = nullptr, callback_function_t CompareCallback = nullptr);
--- a/cores/arduino/HardwareTimer.h
+++ b/cores/arduino/HardwareTimer.h
@@ -112,7 +112,7 @@ class HardwareTimer {
     void setPrescaleFactor(uint32_t prescaler); // set prescaler register (which is factor value - 1)
     uint32_t getPrescaleFactor();
 
-    void setOverflow(uint32_t val, TimerFormat_t format = TICK_FORMAT); // set AutoReload register depending on format provided
+    void setOverflow(uint64_t val, TimerFormat_t format = TICK_FORMAT); // set AutoReload register depending on format provided
     uint32_t getOverflow(TimerFormat_t format = TICK_FORMAT); // return overflow depending on format provided
 
     void setPWM(uint32_t channel, PinName pin, uint32_t frequency, uint32_t dutycycle, callback_function_t PeriodCallback = nullptr, callback_function_t CompareCallback = nullptr); // Set all in one command freq in HZ, Duty in percentage. Including both interrupt.
--- a/libraries/SrcWrapper/src/HardwareTimer.cpp
+++ b/libraries/SrcWrapper/src/HardwareTimer.cpp
@@ -483,18 +483,18 @@ void HardwareTimer::setPrescaleFactor(uint32_t prescaler)
   *           MICROSEC_FORMAT: return number of microsecondes for overflow
   *           HERTZ_FORMAT:    return frequency in hertz for overflow
   */
-uint32_t HardwareTimer::getOverflow(TimerFormat_t format)
+uint64_t HardwareTimer::getOverflow(TimerFormat_t format)
 {
   // Hardware register correspond to period count-1. Example ARR register value 9 means period of 10 timer cycle
   uint32_t ARR_RegisterValue = LL_TIM_GetAutoReload(_timerObj.handle.Instance);
   uint32_t Prescalerfactor = LL_TIM_GetPrescaler(_timerObj.handle.Instance) + 1;
-  uint32_t return_value;
+  uint64_t return_value;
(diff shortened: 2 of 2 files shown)


fpistm:
Hi @GianfrancoIU1JSU 
Thanks for the PR, we will review it, as a first comment, it seems not complete and probably generate warnings.
Check should be done if the value is higher than an UINT32_MAX.
GetOverflow should also be updated to return a uint64_t.
Examples and codes using them should be also updated.

fpistm:
Hi @GianfrancoIU1JSU 
Unfortunately, we remembered that the HardwareTimer library uses all timer like 16 bits timer. See

https://github.com/stm32duino/Arduino_Core_STM32/wiki/HardwareTimer-library#1-introduction

>  For genericity purpose, HardwareTimer library uses all timers like a 16bits timer (even if some may be wider).

So even if this PR allow you to use max 32 bits values, it is not enough to support properly 32 bits timers as some side effects can occurs.

PR #2082, opened 2023-07-22
[chore] Speedup for simple SPI.transfer

RudolphRiedel:
**Summary**

This PR gives a speed boost to simple SPI.transfer(data) and SPI.transfer(buffer, count) operations
in which the user controls the chip-select by directly using only the necessary LL-HAL functions and not going thru the CS controlling functions and the spi_transfer() function.

With no DMA support in the SPI class the sending of a buffer takes a long time and with this modification
it takes at least less time.

Measured with STM32C031C6, SPI clock 8MHz and a buffer of 231 bytes:
digitalWrite(CS, LOW);
SPI.transfer(&buffer, 231);
digitalWrite(CS, HIGH);

original code: 587,26µs
this pr: 474.92µs

### DIFF
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -407,18 +407,18 @@ void SPIClass::transfer(byte _pin, void *_bufout, void *_bufin, size_t _count, S
 uint8_t SPIClass::transfer(uint8_t data)
 {
 #if defined(SPI_SR_TXP)
-    while (!LL_SPI_IsActiveFlag_TXP(_spi.handle.Instance));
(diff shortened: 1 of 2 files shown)


fpistm:
Hi @RudolphRiedel 
I will not be able to review it soon but will get back to you when review it.

fpistm:
Hi @RudolphRiedel 
Thanks for this PR. When reviewed it, I though we can go further and remove some unused and over complicated internal CS management. That's why I've made this PR: #2171 .
Feel free to test ang give your feedback.
So I close this one as the new PR include your proposal.

PR #2099, opened 2023-08-09
chore: update to CMSIS 5.9.0

fpistm:
Package available here:
https://github.com/stm32duino/ArduinoModule-CMSIS/releases/tag/5.9.0

Can be tested using dev package index json:
https://github.com/stm32duino/BoardManagerFiles/raw/dev/package_stmicroelectronics_index.json

### DIFF
--- a/.github/actions/pio-build/README.md
+++ b/.github/actions/pio-build/README.md
@@ -6,7 +6,7 @@ This action build thanks PIO.
 
 ### `cmsis-version`
 
-The CMSIS version to use. Default `"5.7.0"`.
+The CMSIS version to use. Default `"5.9.0"`.
 
 ## Example usage
 
--- a/.github/actions/pio-build/action.yml
+++ b/.github/actions/pio-build/action.yml
@@ -4,7 +4,7 @@ description: 'Compile using PlatformIO'
 inputs:
   cmsis-version:
     description: 'CMSIS package version to use'
-    default: '5.7.0'
+    default: '5.9.0'
 runs:
   using: 'docker'
   image: 'Dockerfile'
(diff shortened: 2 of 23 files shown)


fpistm:
Hi @valeros 
Hope you are fine. I've updated the cmsis version but it seems pio does not like it. Any thoughts on this?

valeros:
Hi @fpistm, thanks for pinging. It seems that the CMSIS package is not updated properly. I'm curious how you handle such updates. I've found the `entrypoint.sh` script that does some manipulations like downloading a CMSIS package, but I don't see any other actions besides extracting the CMSIS package to the home folder of the CI runner. Am I missing something?

fpistm:
It seems it is properly downloaded. With arduino, no issue. The file is properly built.  Maybe GH action cache issue, but seems not the case.

valeros:
> It seems it is properly downloaded. With arduino, no issue. The file is properly built. Maybe GH action cache issue, but seems not the case.

I'm afraid it's not enough to simply extract the CMSIS package in the core folder, PlatformIO won't see this update as the CMSIS is a standalone package in the PlatformIO ecosystem. I see here two ways how to handle this:
1. The easiest approach is to publish the new CMSIS 5.9.0 to the PlatformIO Registry and then dynamically replace the version field in the `entrypoint.sh`. A single drawback is that we'll lose the ability of replacing the CMSIS package with arbitrary versions directly from the https://github.com/stm32duino/ArduinoModule-CMSIS
2. We keep the current workflow with downloading the CMSIS package from https://github.com/stm32duino/ArduinoModule-CMSIS, but introduce some additional logic that will properly substitute the already downloaded CMSIS with a new one. 

Let me know, what's more suitable for you.

fpistm:
Hi @valeros 
OK understood. I thought it already uses the downloaded one. Seems this download was not useful 😉 
So I guess option 2 is the best way else we will got this issue again in the next update.

fpistm:
PIO build will be managed later.

PR #2100, opened 2023-08-10
chore(pio): disable build

fpistm:
Since #2099, pio build failed as it does not used the specified CMSIS version.

/cc @valeros

### DIFF
--- a/.github/workflows/PIO-build.yml
+++ b/.github/workflows/PIO-build.yml
@@ -39,6 +39,11 @@ jobs:
     - name: Checkout
       uses: actions/checkout@main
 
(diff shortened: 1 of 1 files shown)


fpistm:
> into our development queue, but I want to be upfront, right now I don't have an exact timeline on when we'll have this sorted out. Considering the potential impact of the issue on your CI scripts and the importance of maintaining a smooth workflow, I would recommend keeping PlatformIO integration disabled within your CI pipelines.

Thanks @valeros (sorry for the delay, I was on vacation 2 weeks).
No worry, let me know when you could do it.

Issue #5, opened 2017-06-02
ADC Libraries


fpistm:
_From @palmerr23 on April 26, 2017 5:23_

Frederic,

I've been testing the advanced ADC\DAC modes (beyond single conversion, polling) using the templates from STM32Cube_FW_F4_V1.15.0\Projects\STM324x9I_EVAL\Examples\ADC\ and \DAC\

When using these other analog routines, HAL_ADC_MspInit() etc need to be customised for each configuration.

Now I'm trying out the dual-interleaved mode, it has become clear that the code in analog.c is just an initial test.

As HAL_ADC_MspInit() needs to be customised for each configuration, I'm looking at hiding the MSP_Init functions in analog.c at the system level and moving the code up to a library. The Init functions can then be customised for each mode.

When I tried this, other functions break - like dac_stop() and pwm_stop().  I think I'll be able to leave them in analog.c

Give me a few days and I'll have something workable.

Single-Polling and Single-interrupt are working already, using pretty much standard code from the Cube libraries, next up is dual-interleaved

Richard

_Copied from original issue: stm32duino/Arduino_Core_STM32F4#11_

fpistm:
_From @palmerr23 on May 6, 2017 7:3_

Frederic,

I've tested a number of ADC modes - polling, interrupt, DMA. 

Example files are in https://github.com/palmerr23/Black-F407VET6-cube/tree/master/examples

Currently debugging Dual Mode Interleaved  (edit: now working).

fpistm:
Proposed to move on the new repo  [Arduino_Core_STM32](https://github.com/stm32duino/Arduino_Core_STM32) as this one is deprecated. (same code).

fpistm:
_From @palmerr23 on May 24, 2017 8:36_

Frederic,  Fine by me. I'm shifting across to the new core also.

tve:
Q: instead of adding stuff to the arduino-compatible analogRead have you considered using a different ADC API altogether? Leave analogRead for compatibility purposes but make something better available for advanced uses. The main benefit I see is that a new API allows set-up/tear-down to be separated from the actual reading function. For example, https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/master/STM32F1/libraries/STM32ADC looks pretty darn nice to me and should not be a ton of work to port to LL (if that has all the functionality necessary).

fpistm:
Yes, this is what I would like.
Like for timer: HardwareTimer...

[...]

PR #2133, opened 2023-09-18
Initialize each ADC only once in Analogread

Candas1:
**Summary**
This PR improves the speed of analogread by initializing each ADC only once(mentioned in [this](https://github.com/stm32duino/Arduino_Core_STM32/issues/5) issue)

Currently when using analogread, the following steps are executed:
- initialize adc
- calibrate adc
- configure channel
- start adc
- poll for conversion
- stop adc
- deinitialize adc

With this PR, if the ADC wasn't initialized yet by analogread, the following steps are executed:
- initialize adc
- calibrate adc
- configure channel
- start adc
- poll for conversion
- ~~stop adc~~
- ~~deinitialize adc~~

If the ADC was initialized by analogread, only the following steps are executed:
- ~~initialize adc~~
- ~~calibrate adc~~
- configure channel
- start adc
- poll for conversion
- ~~stop adc~~
- ~~deinitialize adc~~

I tested my changes on a STM32F103RCT6.
I ran analogread in a loop:
- before the change, I was getting a frequency of up to 11khz
- after the change, I was getting a frequency of up to 25khz

With this change, I was also able to run injected adc in parallel with analogread without interfering, the only prerequisite is that analogread should run first.

[EDIT] It was late so I forgot some details.
I tried reading the same pin from 3 different ADCs (PC2_ALT0, PC2_ALT1, PC2_ALT2) consecutively in a loop, it was working well, the init happened only once for each ADC. The same example wasn't working before my change, I was getting readings for one of the analogread calls only.

If you think users were using the previous behavior as a feature, I could also #ifdef the new behavior to make it optional.


fpistm:
Hi @Candas1 
Thanks for this PR.
I will review it, IIWR, for H7 ADC is specific as some of them could be directly connected to the pin (dual pad)

Candas1:
Bonjour Frederic,

Thanks for considering this.
I already see Astyle failed, I think this wasn't used when analog.cpp was first committed.
Can I run Astyle on vscode and just let it prettify the code as you expect it ? are you using a particular set of rules ?

Thanks in advance.

fpistm:
> I already see Astyle failed, I think this wasn't used when analog.cpp was first committed.

It was used but as you change the code it do not match anymore the format.

We provide a script to apply change. in CI/astyle and the rules are defined in the  astylerc file: https://github.com/stm32duino/Arduino_Core_STM32/tree/main/CI/astyle

Candas1:
But it's even reporting code that I haven't changed.
Thank you, just found out the astylerc. No worries I will check that.

Candas1:
I got the style fixed locally by astyle. Do you trust this tool or it requires testing again ?
I can push the new changes but I see once check is still running.

fpistm:
> I got the style fixed locally by astyle. Do you trust this tool or it requires testing again ? I can push the new changes but I see one check is still running.

You can force push.

fpistm:
> But it's even reporting code that I haven't changed.

This is normal as you changed code block, so some indent are required :wink:

Candas1:
No worries, I pushed the fixes.
I will use astyle from now on.

fpistm:
@Candas1 
please, could you share the sketch you used to measure ADC performance?

Candas1:
@fpistm I created [this](https://github.com/Candas1/analogread_test) simplified example for you, I hope it's ok

fpistm:
Thanks.

I've made a test with a Nucleo-L476RG with your PR and it breaks  the analogRead.

Candas1:
> Thanks.
> 
> I've made a test with a Nucleo-L476RG with your PR and it breaks the analogRead.

Any details what is happening ?

fpistm:
> Any details what is happening ?

I guess the ADC channel conf moved not help and probably more.
I've suspected this PR would not be so simple as we have to support 20 series and several ADC IP version. STM32F1 is probably the most easiest to manage.
Moreover, looking at your code, the ADC index seems not correct for me as you assumes they are contiguous. Anyway, IIWR on some series you can have ADC1 and ADC3 without ADC2.

Candas1:
> Moreover, looking at your code, the ADC index seems not correct for me as you assumes they are contiguous. Anyway, IIWR on some series you can have ADC1 and ADC3 without ADC2.
Do you have a table of all models and the ADCs it has ?
That seemed obvious to me. But knowing this, I can propose a fix, that's not a problem.

But do you think it's not worth trying ?

Candas1:
Initializing the ADC for each adc sampling is also not the standard way to proceed to be honest.
But if you have some hints what goes wrong I can have a look.

Maybe some families don't like changing the channel configuration while the ADC is on, that could be fixed as well.

Candas1:
I have some B-G431B-ESC1 laying around, I can maybe try my implementation on those.
But if I cannot reproduce it I cannot fix it.

Do you mind keeping this open and we decide later if it's too complicated ?
If some people are interested they can also try on their hardware, I saw other issues that are connected.

CJT1111:
Hello, can you upload your optimized code for a look?

Candas1:
Hi,

I did an obvious mistake that I didn't upload yet in this PR.
Still it wasn't working well on Families other than F1. 
F1 seems less picky as Guillaume said.

Now I got F3/F4/L4/G4/H5 nucleos I can test with.
I need to get back at it and try the low level functions.
I will share it.

CJT1111:
Thank you very much for your reply. I am looking forward to your sharing. Please call me when you share.

Candas1:
> Thank you very much for your reply. I am looking forward to your sharing. Please call me when you share.

Sorry for the late reply.
I updated the code, it's simpler now but it might be slower.
I tested it on G4, need to do more tests.

The only one doubt I have now that it's not possible to change the resolution between reads of the adc is not reinitialized, need to check how to deal about it.

CJT1111:
That's all right, I'm looking forward to your effect. Finally, I have a suggestion: I have done a long experiment to improve the speed of ADC in Arduino development environment, for example DMA, but the Arduinol development environment can not call interrupt, I also consulted fpistm, he replied me that Arduino' official does not have DMA API, so he did not do it Method, so I hope the final optimized code can be a library like stevstrongl's [Arduino_.STM32]
(https:/github.com/rogerclarkmelbourne/Arduino_.STM32), he has an extra myADC library inside
(https:/github.com/rogerclarkmelbourne/Arduino._STM32/tree/master/STM32F1/libraries/STM32ADc), if we can do that, it's really a good move.

Candas1:
> so I hope the final optimized code can be a library like stevstrongl's [Arduino_.STM32]

The scope of this PR is only to optimize analogRead without impacting it's behavior and the projects using it.
I also wanted analogread to not impact other use case (injected adc).
So far I achieved it by skipping the Mspinit/deinit, but I need to test much mure. Man changes are visibles in this PR.
I shared some additional tips [here](https://community.simplefoc.com/t/stm32-analogread/4476/5?u=candas1).

I have idea about an ADC library as part of SimpleFOC but it would require a much more complicated API, it's out of scope here.

CJT1111:
Ok, thank you for your reply.

PR #2135, opened 2023-09-20
Update PlatformIO build script

valeros:
This PR updates PlatformIO build script according to the latest changes in the core:
- Updated build flags
- Better formatting


fpistm:
Hi @valeros 
Thanks for the PR.
Could you rebase on top of the main then the pio CI will be launched.

fpistm:
Hi @valeros 
I've rebased and revert commit which disabled the pio build.
I've added the cmsis_dsp to the list of ignored library, as it always use the CMSIS 5.7.0 eve if the 5.9.0 is defined as default version.

PR #2155, opened 2023-10-29
Wrap cmake policies in if statements

TrentHouliston:
**Summary**

The policy statements for CMake policy CMP0135 do not have if statements around them.
As a result, if you have a CMake version less than 3.24 CMake will fail

The policies should have an if statement so that versions between 3.21 and 3.24 can be built.


fpistm (approved):
LGTM
Thanks @TrentHouliston for this PR.

massonal:
Belatedly, LGTM too 👍

PR #2163, opened 2023-11-02
Optimize flash accesses to reduce wear and increase speed

me21:
**Summary**

This PR introduces two bool flags to avoid unnecessary flash rereads to buffer and flash page erases when writing big objects in EEPROMClass::put.

This will improve the flash memory lifetime and access speed by skipping flash reads or erases when unnecessary.

This PR is based on the results of the discussion https://github.com/orgs/stm32duino/discussions/2161.

### DIFF
--- a/libraries/EEPROM/src/utility/stm32_eeprom.c
+++ b/libraries/EEPROM/src/utility/stm32_eeprom.c
@@ -100,7 +100,7 @@ extern "C" {
 #endif /* FLASH_BASE_ADDRESS */
 
 #if !defined(DATA_EEPROM_BASE)
-static bool is_buffer_filled = false, is_buffer_dirty = false;
+static bool is_buffer_filled = false, do_auto_flush = true;
 static uint8_t eeprom_buffer[E2END + 1] __attribute__((aligned(8))) = {0};
 #endif
 
@@ -142,11 +142,33 @@ void eeprom_write_byte(uint32_t pos, uint8_t value)
   }
 #else
   eeprom_buffered_write_byte(pos, value);
-  eeprom_buffer_flush();
+  if(do_auto_flush)
+    eeprom_buffer_flush();
 #endif /* _EEPROM_BASE */
 }
 
 #if !defined(DATA_EEPROM_BASE)
+
+/**
+  * @brief  This function disables the auto flush in eeprom_write_byte - user will need to call eeprom_buffer_flush() explicitly
+  *         It affects write operations through eeprom_write_byte() explicit call, EEPtr, EERef and EEPROMClass classes
+  * @param  none
+  * @retval none
+  */
+void disable_auto_flush(void)
+{
+  do_auto_flush = false;
+}
+
+/**
+  * @brief  This function enables the auto flush in eeprom_write_byte - it is enabled by default
+  * @param  none
+  * @retval none
+  */
+void enable_auto_flush(void)
+{
+  do_auto_flush = true;
+}
 
 /**
   * @brief  Function reads a byte from the eeprom buffer
@@ -222,8 +244,6 @@ void eeprom_buffer_flush(void)
   */
 void eeprom_buffer_flush(void)
 {
-  if(!is_buffer_dirty)
-    return;
 #if defined(ICACHE) && defined (HAL_ICACHE_MODULE_ENABLED) && !defined(HAL_ICACHE_MODULE_DISABLED)
   bool icache_enabled = false;
   if (HAL_ICACHE_IsEnabled() == 1) {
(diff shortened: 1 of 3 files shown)


libraries/EEPROM/src/utility/stm32_eeprom.c
```diff
@@ -141,12 +142,34 @@ void eeprom_write_byte(uint32_t pos, uint8_t value)
...
+/**
+  * @brief  This function enables the auto flush in eeprom_write_byte - it is enabled by default
+  * @param  none
+  * @retval none
+  */
+void enable_auto_flush(void)
+{
+  do_auto_flush = true;
+}
+
```
fpistm, line 172: Those functions are not defined in .h.
And can be grouped:
```
void eeprom_auto_flush(bool enable)
{
  do_auto_flush = enable;
}
```

libraries/EEPROM/src/utility/stm32_eeprom.c
```diff
@@ -194,6 +219,7 @@ void eeprom_buffer_fill(void)
     }
   }
 #endif /* ICACHE && HAL_ICACHE_MODULE_ENABLED && !HAL_ICACHE_MODULE_DISABLED */
+  is_buffer_filled = true;
```
fpistm, line 222:
```suggestion
```

libraries/EEPROM/src/utility/stm32_eeprom.c
```diff
@@ -175,6 +198,8 @@ void eeprom_buffered_write_byte(uint32_t pos, uint8_t value)
   */
 void eeprom_buffer_fill(void)
 {
+  if(is_buffer_filled)
+    return;
```
fpistm, line 202: Don't think it is a good idea. User should be able to perfom a `eeprom_buffer_fill` when he wants. See the `BufferedEEPROM` example. Moreover it It guarantees the buffer is well aligned with the flash.
```suggestion
```

libraries/EEPROM/src/utility/stm32_eeprom.c
```diff
@@ -141,12 +142,34 @@ void eeprom_write_byte(uint32_t pos, uint8_t value)
   }
 #else
   eeprom_buffered_write_byte(pos, value);
-  eeprom_buffer_flush();
+  if(do_auto_flush)
+    eeprom_buffer_flush();
```
fpistm, line 146:
```suggestion
  if(do_auto_flush) {
    eeprom_buffer_flush();
  }
```

libraries/EEPROM/src/utility/stm32_eeprom.c
```diff
@@ -100,6 +100,7 @@ extern "C" {
 #endif /* FLASH_BASE_ADDRESS */
 
 #if !defined(DATA_EEPROM_BASE)
+static bool is_buffer_filled = false, do_auto_flush = true;
```
fpistm, line 103: See my comment below about `is_buffer_filled` usage
```suggestion
static bool do_auto_flush = true;
```

libraries/EEPROM/src/utility/stm32_eeprom.h
```diff
@@ -116,6 +116,8 @@ void eeprom_write_byte(uint32_t pos, uint8_t value);
 #if !defined(DATA_EEPROM_BASE)
 void eeprom_buffer_fill();
 void eeprom_buffer_flush();
+void eeprom_disable_auto_flush();
+void eeprom_enable_auto_flush();
```
fpistm, line 120: Not aligned with .c definition

fpistm (changes requested):
This PR does not build by default due to issue within definition/declaration mismatch.
Did you test it ?

libraries/EEPROM/src/utility/stm32_eeprom.c
me21: I did a test build, it built just fine for me. But I agree it's better to group them into one function.

me21:
> This PR does not build by default due to issue within definition/declaration mismatch. Did you test it ?

I did a test build. It built just fine for me. But I think I better close this PR and create a new one since there are too many small commits resulting from editing the files directly in the Github web UI. What do you think?

libraries/EEPROM/src/utility/stm32_eeprom.c
me21: I agree, filling the buffer again is a way to discard the changes before they are written to the flash.
But at the same time I feel that `EEPROMClass::read` doing that on each byte read is a pessimization. It's true that user can read data using `eeprom_buffered_read_byte` if he doesn't want that, but shouldn't the library offer the same capability through `EEPROMClass`?

fpistm:
As you want. Commits could be squash when PR merged.

PR #2174, opened 2023-11-10
Add Generic F723-F730-F733 Variants

shreyask21:
Added generic F723I(C-E)(K-T)_F730I8K_F733IE(K-T) variant
Added HSI clock config and ldscript from STM32CubeIDE


fpistm (approved):
Thanks @shreyask21 for this PR.
I've made some changes and now it is OK.

PR #2254, opened 2024-01-19
HardwareSerial's flush() method to override the method of the base class Print

ilolis:
This PR addresses the following **bugs/features**

* [x] Bug: Fixing issue with base class Print's `void flush()` method not being overridden by derivative HardwareSerial's `void flush(uint32_t timeout = 0)` method.

The base class Print [defines `void flush()`](https://github.com/ilolis/Arduino_Core_STM32/blob/dce241d8ebb6affce38adbeebf34eb6372cfbe9b/cores/arduino/Print.h#L103), while the derivative HardwareSerial[ defines `void flush(uint32_t timeout = 0)`](https://github.com/stm32duino/Arduino_Core_STM32/blob/1863c25025441c8c6dbcea5b8ed496db21c12785/cores/arduino/HardwareSerial.h#L128). The issue was that when using the base class instance pointer, the virtual function of the derivative was not being called. This PR resolves this. 

I might be missing something since I am new to this project, so please feel free to correct me :)

### DIFF
--- a/cores/arduino/HardwareSerial.cpp
+++ b/cores/arduino/HardwareSerial.cpp
@@ -487,6 +487,11 @@ int HardwareSerial::availableForWrite(void)
   return tail - head - 1;
 }
 
+void HardwareSerial::flush()
+{
+  flush(0);
+}
+
 void HardwareSerial::flush(uint32_t timeout)
 {
   // If we have never written a byte, no need to flush. This special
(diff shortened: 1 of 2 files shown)


fpistm (approved):
LGTM
Thanls @ilolis

fpistm:
Tested with:
```
Stream &myStream = Serial; 

void setup() {
  Serial.begin(9600);
}

void loop() {
  // read the input on analog pin 0:
  int sensorValue = analogRead(A0);
  // print out the value you read:
  myStream.println(sensorValue);
  myStream.flush();
  delay(1);  // delay in between reads for stability
}
```

PR #2283, opened 2024-02-21
Multiple sketch support

### LINKED ISSUE #2282: cmake build fails if more than one build_sketch definition is used
I am trying to build multiple sketches using a project wide CMakeLists.txt file. If I call the build_sketch() function with each of the targets then cmake reports an error about duplicated targets:
```
...
Make Error at /home/dpd/.arduino15/packages/STMicroelectronics/hardware/stm32/2.7.1/variants/STM32F1xx/F103C8T_F103CB(T-U)/CMakeLists.txt:5 (add_library):
  add_library cannot create target "variant" because another target with the
  same name already exists.  The existing target is an interface library
  created in source directory
  "/home/dpd/.arduino15/packages/STMicroelectronics/hardware/stm32/2.7.1/variants/STM32F1xx/F103C8T_F103CB(T-U)".
  See documentation for policy CMP0002 for more details.
...
```
I have tried various cmake property settings but the error persists. I was able to get the builds to work by putting an `if (NOT TARGET variant)` guard around the add_subdirectory calls in build_sketch.cmake.
```patch
*** build_sketch.cmake.orig	2024-02-21 17:32:23.290460959 +1100
--- build_sketch.cmake	2024-02-21 17:33:09.851221766 +1100
***************
*** 6,14 ****
  include(set_base_arduino_config)
  
  function(build_sketch)
!   add_subdirectory(${BUILD_VARIANT_PATH} ./variant)
!   add_subdirectory(${BUILD_CORE_PATH} ./cores/arduino)
!   add_subdirectory(${BUILD_LIB_PATH} ./libraries)
  
    cmake_parse_arguments(PARSE_ARGV 0 SKBD "" "TARGET" "SOURCES;DEPENDS")
  
--- 6,17 ----
  include(set_base_arduino_config)
  
  function(build_sketch)
!   if (NOT TARGET variant)
!     add_subdirectory(${BUILD_VARIANT_PATH} ./variant)
!     add_subdirectory(${BUILD_CORE_PATH} ./cores/arduino)
!     add_subdirectory(${BUILD_LIB_PATH} ./libraries)
!   endif()
! 
  
    cmake_parse_arguments(PARSE_ARGV 0 SKBD "" "TARGET" "SOURCES;DEPENDS")
  ```
If there is a better solution then I would love to know otherwise I can submit a pull request.

### PR DESCRIPTION
**Pull request by ddowling**
**Summary**

Allow CMakeLists.txt files to contain multiple build_sketch() targets.

This PR fixes #2282 

A lot of Arduino library projects will contain multiple example applications that demonstrate the library functionality. It is also common practice on larger embedded applications to have a number of small test programs that test specific parts of the systems so these can be tested in isolation from the overall application code. 
This change means all of the examples and tests can be built together from the top level CMakeList.txt file.

There is a dependency error when _.ino_ files are not in the parent directory of the project that is fixed with this pull request.

### DIFF
--- a/cmake/sketch_preprocess_sources.cmake
+++ b/cmake/sketch_preprocess_sources.cmake
@@ -27,7 +27,7 @@ function(sketch_preprocess_sources)
         COMPILE_OPTIONS "-include;Arduino.h;-include;${SRCFILE}.h"
         OBJECT_DEPENDS "${SRCFILE}.h"
       )
-      list(APPEND SRCLIST ${SRCFILE}.cpp)
+      list(APPEND SRCLIST ${CMAKE_CURRENT_BINARY_DIR}/${SRC_BASE_NAME}.cpp)
     else()
       list(APPEND SRCLIST ${SRCFILE})
     endif()
--- a/cmake/sketch_preprocess_sources.cmake
+++ b/cmake/sketch_preprocess_sources.cmake
@@ -5,29 +5,31 @@ function(sketch_preprocess_sources)
   set(SRCLIST "")
   foreach(SRCFILE IN LISTS SPC_SOURCES)
     if (${SRCFILE} MATCHES "\.ino$")
+      # Convert <file>.ino to ${CMAKE_CURRENT_BINARY_DIR}/<file>.ino.{cpp,h}
       cmake_path(GET SRCFILE FILENAME SRC_BASE_NAME)
+      set(SRC_BINARY_BASE_NAME ${CMAKE_CURRENT_BINARY_DIR}/${SRC_BASE_NAME})
 
       configure_file(
         ${SRCFILE}
-        ${CMAKE_CURRENT_BINARY_DIR}/${SRC_BASE_NAME}.cpp
+        ${SRC_BINARY_BASE_NAME}.cpp
         COPYONLY
       )
 
-      add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${SRC_BASE_NAME}.h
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmake/sketch_preprocess_sources.cmake [hunk 1]
```diff
@@ -27,7 +27,7 @@ function(sketch_preprocess_sources)
         COMPILE_OPTIONS "-include;Arduino.h;-include;${SRCFILE}.h"
         OBJECT_DEPENDS "${SRCFILE}.h"
       )
-      list(APPEND SRCLIST ${SRCFILE}.cpp)
+      list(APPEND SRCLIST ${CMAKE_CURRENT_BINARY_DIR}/${SRC_BASE_NAME}.cpp)
```
massonal, line 30: Can you apply the same change on line 25?
```py
set_source_files_properties(${SRCFILE}.cpp
```

ddowling: Seems I missed some other places where ${CMAKE_CURRENT_BINARY_DIR} was not prepended to the source file. I have pushed another commit. This has uncovered another issue with my code however as the generated include file has definitions like `bool help_cmd(CommandParser *)` which trip the compiler up as the generated `sketch.ino.h` gets added  included in `sketch.ino.cpp` before the actual includes for the library. I will see how currently Arduino build manages the includes and get back with a fix.

ddowling: Actually this appears to be a "known" issue with how Arduino ino file pre-processing works and it not specific to the cmake support. 
See [https://www.gammon.com.au/forum/?id=12625]. I will find a solution to this in my own code.

massonal (approved):

fpistm (approved):

PR #2399, opened 2024-06-12
feat: Blues Cygnet

zfields:
This migrates the Ray's original Cygnet work to the STM32L433CCT6, and incorporates many suggestions from @fpistm's code review.


zfields:
This migrates the Ray's original Cygnet work to the STM32L433CCT6, and incorporates many suggestions from @fpistm's code review.

@fpistm Can you please sanity check this work. This is my first time making significant modifications to the board definition.

tools/platformio/boards_remap.json
```diff
@@ -31,6 +31,8 @@
     "vccgnd_f103zet6": "VCCGND_F103ZET6_MINI",
     "waveshare_open103z": "GENERIC_F103ZEHX",
     "bw_swan_r5": "SWAN_R5",
```
zfields, line 33: left for backward compatibility

fpistm (changes requested):
Hi @zfields 
Thanks for the updated PR.
Just on remarks about the link even if reserved on your side I would prefer add it when really available to avoid further support because one link does not exist and not related to other PR.

zfields:
@fpistm I've adopted your change. :+1: Thank you for your help! :heart:

fpistm (approved):
Thanks @zfields
I've squash some commits to be more clean.
LGTM

Issue #2162, opened 2023-11-01
[New Variant] Support for RAK3271T


HelgeSeidel:
I kindly request support for the RAK3271T module. 

Existing variant RAK3271 worked fine for the module RAK3271E and RAK3271S. The RAK3172T module has an TXCO installed for a lower operation temperature and need a special configuration. Changing the LoRaWAN definition LORAWAN_BOARD_HAS_TCXO in [variant_RAK3172_MODULE.h](https://github.com/Oliv4945/Arduino_Core_STM32/blob/3874daca7e48a15801441a6a88ba471a7a092fdd/variants/STM32WLxx/WL54CCU_WL55CCU_WLE4C(8-B-C)U_WLE5C(8-B-C)U/variant_RAK3172_MODULE.h) from 0U to 1U was successful.

For details please see [RAK3172T TXCO configuration Issue](https://github.com/stm32duino/STM32LoRaWAN/issues/34).

Unfortunately I was not able to do the needed changes.

fpistm:
> Unfortunately I was not able to do the needed changes.

Hi, added to #722 . Anyway I don't understand why you are not able to do the change if it works with other variant by simply change TCXO value.

PR #2431, opened 2024-07-10
Added RAK3172T Variant

nmaas87:
**Summary**

Adding Support for RAK3217T, should close https://github.com/stm32duino/Arduino_Core_STM32/issues/2162 and https://github.com/stm32duino/STM32LoRaWAN/issues/34 - RAK3217 was already available, RAK3217T uses a TCXO which needed to be enabled in the variant files so that examples with STM32LoRaWAN works.

* https://github.com/stm32duino/Arduino_Core_STM32/issues/2162
* https://github.com/stm32duino/STM32LoRaWAN/issues/34 

Needed to use the RAK3172T myself and wanted the support to be available to all users.

**Validation**

* LoRaWAN example from STM32LoRaWAN works and I can successfully connect to my gateway. Without the change, the TCXO is not used and the radio does not work correctly / cannot receive anything.


fpistm:
Hi @nmaas87 
Thanks for the PR, I've simplified it using the same variant_*.* files as only TCXO value is different.

fpistm (approved):
LGTM

PR #2435, opened 2024-07-12
Add single buffer mode option to USB CDC to fix off-by-one corruption

gigaj0ule:

[Bug]

When using the debug mode, halting the CPU at a breakpoint can create a condition where the HAL USB state machine encounters an off-by-one error in the packet handling. This error is un-recoverable. completely breaks USB communications, and requires you to reset the STM32.

This is caused by an issue with the double-buffering option available in the endpoint hardware.

[Fix] 

Adding the option to disable double buffer mode, and to allow single buffer mode, fixes this issue, and, in my tests, also improves reliability in USB communication with noisy systems, such as motor controllers.

Single buffer mode can now be enabled with -DUSBD_CDC_USE_SINGLE_BUFFER at compile time. 

[Breaking: None] 

This commit does not break any functions of STM32DUINO and is a completely optional command-line option.


fpistm (approved):
No issue foreseen.
LGTM

Issue #2438, opened 2024-07-14
Error uploading with stm32CubeProgrammer (DFU) since v2.8.0


brickZA:
**Describe the bug**
Get the following error uploading to a F401CCx blackpill board when upgrading to v2.8.0:

```
Sketch uses 39208 bytes (14%) of program storage space. Maximum is 262144 bytes.
Global variables use 5416 bytes (8%) of dynamic memory, leaving 60120 bytes for local variables. Maximum is 65536 bytes.
java.io.IOException: Cannot run program "{busybox}": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at java.lang.Runtime.exec(Runtime.java:620)
	at java.lang.Runtime.exec(Runtime.java:485)
	at processing.app.helpers.ProcessUtils.exec(ProcessUtils.java:11)
	at cc.arduino.packages.Uploader.executeUploadCommand(Uploader.java:129)
	at cc.arduino.packages.uploaders.SerialUploader.runCommand(SerialUploader.java:383)
	at cc.arduino.packages.uploaders.SerialUploader.uploadUsingPreferences(SerialUploader.java:197)
	at cc.arduino.UploaderUtils.upload(UploaderUtils.java:77)
	at processing.app.SketchController.upload(SketchController.java:732)
	at processing.app.SketchController.exportApplet(SketchController.java:703)
	at processing.app.Editor$UploadHandler.run(Editor.java:2055)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 11 more
An error occurred while uploading the sketch
```

**To Reproduce**
Complete source code which can be used to reproduce the issue. Please try to be as generic as possible (no extra code, extra hardware,...)

1. Upgrade core to v2.8.0
2. Upload as per usual
3. See error above

**Expected behavior**
It uploads succesfully

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04 x86-64
 -  - Arduino IDE version: 1.8.13
 - STM32 core version: 2.8.0
 - Tools menu settings if not the default:
[...]

PR #2439, opened 2024-07-14
Fix for issue 2438: Error uploading with stm32CubeProgrammer (DFU)

brickZA:
## Pull Request template

**Summary**

This PR fixes/implements the following  bug:

* [Issue 2438: Error uploading with stm32CubeProgrammer (DFU) since v2.8.0](https://github.com/stm32duino/Arduino_Core_STM32/issues/2438) 

Fix v2.8.0 regression that breaks uploads by reverting two lines that changed in `platform.txt` back to the 2.7.1 version.

**Validation**

Validated uploading works again.


fpistm:
Arduino 1.8.x no more supported. 
This PR simply bring back old definition but only for one target and mixing old and new definition.

PR #2442, opened 2024-07-16
Fix Wire memory leak

### LINKED ISSUE #2441: Wire memory leak on multiple begin() call
**Describe the bug**
When calling begin() after the receive and transmit buffers have been allocated, the begin function orphans the allocations by resetting the pointer to nullptr, on the next call of allocateTxBuffer(), we get a new memory allocation while the previous one has never been freed, this in turns causes heap segmentation and will lead to lack of memory in the long run.

Proposed fix in #2442 

**To Reproduce**
Using this simplified code, we can see that the Wire receive buffer is reallocated every requestFrom call
```
#include <Arduino.h>
#include <Wire.h>

void setup() {
}

void loop() {
  Wire.begin();
  Wire.requestFrom(1,1);
}
```

1. Run demo code
2. Place a breakpoint in allocateRxBuffer right after the realloc line
3. Notice the tmp pointer increase location every call
4. Wait long enough and you'll run out of memory

**Expected behavior**
The wire library should not cause a memory leak, even if calling begin() multiple times is bad practice.

 - Name: BlackPill STM32F411CEU6
 - Hardware Revision: V3.1

  Applicable to any stm32

benmalenfant:
**Summary**

This PR fixes the memory leak bug caused by multiple Wire.begin() calls, mentioned in #2441

The strategy used here is to initialise the buffer variables on the Wire object initialisation instead of the begin.

**Validation**

* Ensure CI build is passed.
* Ran the example code in #2441, now allocateRxBuffer() is only called once at the first call of recieveFrom() and the memory leak has been fixed

**Code formatting**

**Closing issues**

Fixes #2441

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -38,6 +38,11 @@ TwoWire::TwoWire()
   memset((void *)&_i2c, 0, sizeof(_i2c));
   _i2c.sda = digitalPinToPinName(SDA);
   _i2c.scl = digitalPinToPinName(SCL);
+
+  txBuffer = nullptr;
+  txBufferAllocated = 0;
+  rxBuffer = nullptr;
+  rxBufferAllocated = 0;
 }
 
 TwoWire::TwoWire(uint32_t sda, uint32_t scl)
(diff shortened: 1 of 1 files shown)


fpistm (approved):
LGTM
Moreover it makes sense to not init pointer and size at `begin`() as the `reset*Buffer` are called after and do nothing if the pointer is null which was the case before this PR.

Issue #2445, opened 2024-07-16
2.8.0 breaks core reference


mtboswell-aeroseal:
**Describe the bug**
When compiling an Arduino sketch for a custom variant defined in the sketchbook's /hardware/ folder, the SrcWrapper library is not found because the path to the SrcWrapper library is specified with the {runtime.platform.path} variable which points to the hardware folder in the sketchbook instead of the hardware folder in the ST package.

**To Reproduce**

1. Create a variant based on the STM32 Arduino platform and place it in the sketchbook/hardware/org/stm32/ folder. 
2. Attempt to compile an empty sketch targeting this custom variant.

**Expected behavior**
Libraries are found and sketch compiles.

**Screenshots**
None

**Desktop (please complete the following information):**
 - OS: Windows
 - Arduino IDE version: 2.3.2
 - STM32 core version: 2.8.0
 - Tools menu settings if not the default: N/A
 - Upload method: N/A

**Board (please complete the following information):**
 - Name: Custom
 - Hardware Revision: Custom
 - Extra hardware used if any: N/A

**Additional context**
Line 32 of platform.txt changed the path of several includes to point to a path including the {runtime.platform.path} variable. This variable does not always point to the ST board package and thus breaks the include path for custom variants specified in the sketchbook folder.

fpistm:
Have you an zip to reproduce this? I don't understand what you called a custom variant? Or any Arduino documentation ?

mtboswell-aeroseal:
Unfortunately, we do not have a zip that we can share publicly. 

Here is the part of the platform specification that describes referencing another platform: https://arduino.github.io/arduino-cli/1.0/platform-specification/#referencing-another-core-variant-or-tool .

In our %SKETCHBOOK%\hardware\MyOrg\stm32\boards.txt file we have a line similar to:

myboard.build.core=STMicroelectronics:arduino

Which allows us to use the ST core even though we have our own separate boards.txt.

This works fine on 2.7.1, but on 2.8.0 the compile fails because the ST include path now has the {runtime.platform.path} variable in it, and the {runtime.platform.path} variable changes from "%LOCALAPPDATA%\Arduino15\packages\STMicroelectronics\hardware\stm32\2.8.0" to "%SKETCHBOOK%\hardware\MyOrg\stm32" when a board from the local platform is selected.

mtboswell-aeroseal:
[...]

PR #2454, opened 2024-07-18
fix(platform.txt): add default vid/pid definition

fpistm:
this avoids to add them to `boards.txt` if not needed.

cmake updated to get the default value from `platform.txt`.

/cc @massonal 

This fixes this:
https://github.com/stm32duino/Arduino_Core_STM32/issues/2445#issuecomment-2231956589

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -25,8 +25,8 @@ Nucleo_144.upload.maximum_size=0
 Nucleo_144.upload.maximum_data_size=0
 Nucleo_144.vid.0=0x0483
 Nucleo_144.pid.0=0x5740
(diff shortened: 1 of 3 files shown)


massonal (commented):
The comment at the top of `platformtxt_filter` is now outdated. It needs to mention this new policy wrt vid/pid.

fpistm:
> The comment at the top of `platformtxt_filter` is now outdated. It needs to mention this new policy wrt vid/pid.

https://github.com/stm32duino/Arduino_Core_STM32/commit/ecdf6af8ac4f25f2bce828bd0aaf88fc3f262959

PR #2490, opened 2024-08-20
chore: mark some segments as READONLY

### LINKED ISSUE #2475: LOAD segment with RWX permissions during stm32f411ceu6 flashing on arduinoststm32 @ 4.20801.240802 (2.8.1)
The error is not critical, because the program code works without any comments, but I don't like the concept itself, when I see warnings during compilation:

> Executing task in folder Example: C:\Users\test\.platformio\penv\Scripts\platformio.exe run 
> 
> Processing genericSTM32F411CE (platform: ststm32; board: blackpill_f411ce; framework: arduino)
> -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Verbose mode can be enabled via `-v, --verbose` option
> CONFIGURATION: https://docs.platformio.org/page/boards/ststm32/blackpill_f411ce.html
> PLATFORM: ST STM32 (17.4.0) > WeAct Studio BlackPill V2.0 (STM32F411CE)    
> HARDWARE: STM32F411CEU6 100MHz, 128KB RAM, 512KB Flash
> DEBUG: Current (blackmagic) External (blackmagic, cmsis-dap, jlink, stlink)
> PACKAGES:
>  - framework-arduinoststm32 @ 4.20801.240802 (2.8.1)
>  - framework-cmsis @ 2.50900.0 (5.9.0)
>  - toolchain-gccarmnoneeabi @ 1.120301.0 (12.3.1)
> LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf

> Compiling .pio\build\genericSTM32F411CE\src\main.cpp.o
> Linking .pio\build\genericSTM32F411CE\firmware.elf
> c:/users/test/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/12.3.1/../../../../arm-none-eabi/bin/ld.exe: warning: .pio/build/genericSTM32F411CE/firmware.elf has a LOAD segment with RWX permissions
> Checking size .pio\build\genericSTM32F411CE\firmware.elf
> Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
> RAM:   [          ]   1.2% (used 1604 bytes from 131072 bytes)
> Flash: [=         ]   7.9% (used 41468 bytes from 524288 bytes)
> Building .pio\build\genericSTM32F411CE\firmware.bin

To get rid of it, I had to use the search for a solution.
There were several mentions [(here](https://stackoverflow.com/questions/73429929/gnu-linker-elf-has-a-load-segment-with-rwx-permissions-embedded-arm-project) and [here)](https://wiki.st.com/stm32mcu/wiki/STM32CubeIDE:STM32CubeIDE_errata_1.15.x#General_issues) of similar problems, but no clear solution was found.
Through some experiments, I managed to eliminate the warning during compilation, the solution is below:
1. It is necessary to find the linker file of the microcontroller for which the firmware is compiled, for my MCU `stm32f411ceu6`  the path is as follows

> .platformio\packages\framework-arduinoststm32\variants\STM32F4xx\F411C(C-E)(U-Y)\ldscript.ld
2. Change the lines like this
```
  .preinit_array     :
  {
    . = ALIGN(4);
    PROVIDE_HIDDEN (__preinit_array_start = .);
    KEEP (*(.preinit_array*))
    PROVIDE_HIDDEN (__preinit_array_end = .);
    . = ALIGN(4);
  } >FLASH

  .init_array (READONLY) :
  {
    . = ALIGN(4);
    PROVIDE_HIDDEN (__init_array_start = .);
    KEEP (*(SORT(.init_array.*)))
    KEEP (*(.init_array*))
    PROVIDE_HIDDEN (__init_array_end = .);
    . = ALIGN(4);
  } >FLASH

  .fini_array (READONLY) :
  {
    . = ALIGN(4);
    PROVIDE_HIDDEN (__fini_array_start = .);
    KEEP (*(SORT(.fini_array.*)))
    KEEP (*(.fini_array*))
    PROVIDE_HIDDEN (__fini_array_end = .);
    . = ALIGN(4);
  } >FLASH
```
Add `(READONLY)` for two functions except .preinit_array, although the example above in the link said the opposite in the `STM32CubeIDE errata 1.15.x` document.
After this, the firmware will be installed without warnings.

fpistm:
This PR reverts 794945c10abb2dc6a875c471bb9aa4cbb131e5a2 then update all linker scripts to mark some segments as `READONLY`.    

Prevent LOAD segment with RWX permissions warning.
The `READONLY` keyword is only supported in GCC11 and later, remove it if using GCC10 or earlier.

Fixes #2475.


fpistm:
This PR fixed all warnings for standard application.
Anyway, if a Flash operation (write) is performedclosed to the code segment warning can be raised.
Main example is when the EEPROM emulation is used as it uses last flash page, mainly for mcu with small flash size.

fpistm:
To go further, warning is reached due to the usage of at least one RAM function:
`#define __RAM_FUNC __attribute__((section(".RamFunc")))`

Example with HAL FLASH API: 
https://github.com/stm32duino/Arduino_Core_STM32/blob/2aa4e44bc5f114b2b6eae053d8ebc0eb45a4e93a/system/Drivers/STM32C0xx_HAL_Driver/Src/stm32c0xx_hal_flash.c#L653

So, I kept the warning disabled, but new linker script should use the `READONLY` attribute to prevent warning.

PR #2502, opened 2024-08-28
Remove atomicity warning that does not apply to STM32

nocnokneo:
**Summary**

Removes an atomicity warning in `HardwareSerial.h` that does not apply to (32-bit) STM32. Based on a review on [this response](https://stackoverflow.com/a/52785864/471839) and my own testing, I'm convinced that the atomicity warning and link to an ArduinoCore-avr issue is not relevant.


fpistm (changes requested):
Hi @nocnokneo 
Thanks for thePR.
You're right this warning could be removed it is an artifact when the class was created using original Arduino Class.

fpistm (approved):

PR #2526, opened 2024-09-25
variant(wb55): add STeaMi board

nedseb:
Add support to the STeaMi board developed for the project I-NOVMICRO with help from ST Rousset.

I tried to follow all the steps of the Wiki. Let me know if something is missing.

Issue #2437, opened 2024-07-13
Add OpenOCD + CMSIS-DAP upload and debug


maxgerhardt:
**Is your feature request/improvement related to a problem? Please describe.**

Some people don't have a ST-Link or a regular USB-to-UART adapter, but have a e.g. Raspberry Pi Pico, which can be [easily programmed](https://github.com/raspberrypi/debugprobe/releases) (drag and drop firmware file onto USB drive) to be a CMSIS-DAP compliant debugger and UART probe.

However, this core fails to support upload via CMSIS-DAP.

**Describe the solution you'd like**

Include the xPack OpenOCD package and add the upload action for every board with `-f interfaces/cmsis-dap.cfg`. Using OpenOCD also greatly expands the number of other usable uploaders and debugging methods (e.g., JLink).

PlatformIO can already do this perfectly fine.

**Describe alternatives you've considered**

1. Use the UART upload with the Pico: Slower, plus you have to constantly switch the bootloader jumpers between 0 and 1 for bootloader or firmware run mode
2. Make people buy an actual ST-Link: Why, when they have a perfectly capable CMSIS-DAP 

**Additional context**

Example OpenOCD upload to STM32F1

```
openocd -d2 -s C:\Users\Max\.platformio\packages\tool-openocd/openocd/scripts -f interface/cmsis-dap.cfg -c "transport select swd" -f target/stm32f1x.cfg -c "program {.pio\build\genericSTM32F103C8\firmware.elf}  verify reset; shutdown;"
xPack Open On-Chip Debugger 0.12.0-01004-g9ea7f3d64-dirty (2023-01-30-15:04)
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
debug_level: 2

swd
Info : Using CMSIS-DAPv2 interface with VID:PID=0x2e8a:0x000c, serial=E66038B713193E30
Info : CMSIS-DAP: SWD supported
Info : CMSIS-DAP: Atomic commands supported
Info : CMSIS-DAP: Test domain timer supported
Info : CMSIS-DAP: FW Version = 2.0.0
Info : CMSIS-DAP: Interface Initialised (SWD)
Info : SWCLK/TCK = 0 SWDIO/TMS = 0 TDI = 0 TDO = 0 nTRST = 0 nRESET = 0
Info : CMSIS-DAP: Interface ready
```

fpistm:
Hi @maxgerhardt 

[...]

PR #2528, opened 2024-09-27
feat(platform): Add OpenOCD programming support (STlink and DAP)

nedseb:
- Modified `platform.txt` to enable programming via OpenOCD, which was already used for debugging.
- Added an example demonstrating how to program all Nucleo-64 boards, whether they use STLink or DAPLink.

To avoid conflict, the support for STeaMi will be added after the merge of the PR #2526

*Benefits:*

- Enables programming of STeaMi boards.
- Supports P_NUCLEO boards that use DAPLink instead of STLink to leverage WebUSB functionality.


fpistm:
Hi @nedseb 

Thanks again for the PR.

> Seems related to issue #2437
Right. 
/CC @maxgerhardt 

About the PR I've pushed 2 commits. 1 to fix issue with the menu and 1 to enhance the openOCD tagret definition.
I've tested the debug and it always works.

Ivoz:
Nice feature @nedseb , hopefully will be fun to test using a daplink dongle using this

PR #2542, opened 2024-11-02
cdc_queue.cpp change to uint32_t length, print(0ULL) bugfix

warmonkey:
## Pull Request template
**Summary**
raise cdc buffer size 32KB limit due to uint16_t length variables.

This PR fixes bug:
* cdc queue limited
now CDC_TRANSMIT_QUEUE_BUFFER_PACKET_NUMBER  and/or CDC_RECEIVE_QUEUE_BUFFER_PACKET_NUMBER both can be greater than 512. which every buffer size is 64B, 64*512=32768

* print(0ULL) no output 
https://github.com/arduino/ArduinoCore-API/issues/178

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -375,6 +375,12 @@ size_t Print::printULLNumber(unsigned long long n64, uint8_t base)
   uint8_t i = 0;
   uint8_t innerLoops = 0;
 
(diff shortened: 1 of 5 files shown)


fpistm:
Hi @warmonkey thanks for this PR. 
Could you split in 2.PR  one for cdc and one fir the print, please?

PR #2547, opened 2024-11-04
chore(usb): cdc_queue now use uint32_t length

fpistm:
### Summary
remove cdc buffer size 32KB limit due to uint16_t length variables.

This PR fixes bug:
cdc queue limit.
now CDC_TRANSMIT_QUEUE_BUFFER_PACKET_NUMBER and/or CDC_RECEIVE_QUEUE_BUFFER_PACKET_NUMBER both can be greater than 512. which every buffer size is 64B, 64*512=32768


fpistm:
Closed in favor of #2550

Issue #2538, opened 2024-10-28
UART Tx/Rx Pin Swap


fronders:
I'm working on implementing Rx/Tx pin swap functionality (mentioned in #1418) for a project at the moment, thought I'd open a PR when I'm done.

I have a question about which option you guys prefer:
- have a dedicated pin swap enable method for user to explicitly select i.e. smth like `void setPinSwap(bool enabled);`. This means user would have to call it before `begin()`, with `Rx` and `Tx` pins specified _unswapped_ in constructor or `setRx()` and `setTx()`
- automatically enable pin swap if Tx pin is found in the Rx pin map and vice versa

I personally prefer the second one that would just let the user select needed Rx Tx pins wanted and if MCU supports it things would just work. 
What do you think?

fpistm:
Hi @fronders 
I agree second one seems good option.

Issue #1418, opened 2021-06-11
[U(S)ART] request list


fpistm:
This issue aims to list all U(S)ART requests:

- [x] Hardware Flow control for serial communication (#394) --> #1634
- [x] Support for UART RX/TX level inversion  (#1160 and #2669) --> #2701
- [x] HardwareSerial Error Callback( #1162) --> wont fix as only one request and no work on this.
- [x] Support 9 Bit Mode (https://github.com/stm32duino/Arduino_Core_STM32/issues/1418#issuecomment-874684382) --> Only one requestand not supported by Arduino Core API.

## Note

**This is a community projects,  Any contribution is welcome, so feel free submit a PR.** 😉

4ndreas:
I needed the 9 Bit Mode for a project and changed all the buffers to uint16_t and it worked. But I'm not sure if there are some side effects.
Could also solved with a flag for the 9th bit.

fpistm:
Hi @4ndreas 
THis was discussed some years ago:
https://github.com/stm32duino/Arduino_Core_STM32/issues/482#issuecomment-476229801

Main risk is the compatibility issue.

matthijskooijman:
I have one more suggestion for the list: Support serial writes when interrupts are disabled (e.g. from an ISR). This would make debugging a lot easier, and prevent surprise when someone prints from an ISR and the entire board locks up as soon as the buffer is full.

To support this usecase, the busy-wait loops in `flush()` and `write()` must be modified to call the ISR (or other code that moves the buffer into the UART) whenever a bit is written.

I implemented something similar on the Arduino AVR core, where IIRC it checks if interrupts are disabled, and if so, checks the TXE bit inside the busywait loop and calls the ISR if it is set. A similar approach can be taken here, though there are some complications:
 - There is not just a single "interrupts disabled" bit, AFAIU when another interrupt is running, the global interrupt mask is unchanged, but IRQ priorities might still prevent the TX interrupt from firing. To simplify things, I think the busy wait loops should simply disable the TXE/TXC (or whichever is used exactly) before the loop, and then inside the loop *always* call the ISR manually, even when interrupts were enabled already.
 -  There is some additional layering with the HAL, the stm32/uart.c and HardwareSerial that makes it a bit more tricky to get this right.

The second point makes this more complicated than I have time for right now, but I at least wanted to share this request and my thoughts.

fpistm:
Thanks @matthijskooijman for sharing.
[...]

PR #2601, opened 2024-12-11
feat(uart): support UART Tx Rx pin swap function

fronders:
This enables UART Tx and Rx pin swap function on STM32 families that support it (F0, L0, L4, G0, G4, H7 etc.) 

In order to enable pin swap, user just needs to _swap the pins_ either when declaring `HardwareSerial` object:

```cpp
//                      RX    TX
HardwareSerial Serial1(PA10, PA9); // uses normal pins
//                      RX    TX
HardwareSerial Serial1(PA9, PA10); // uses swapped pins
```

or by using `setTx()` and `setRx()` methods:
```cpp
Serial1.setRx(PA9);  // normally it is a Tx pin
Serial1.setTx(PA10); // normally it is a Rx pin
Serial1.begin(9600);
```

If the chip does not support the swap, or the pins are incorrect, the library behaves the same way as before (throws an error for invalid pins).

The cool thing is that it even supports half-duplex mode on the _normally Rx pin_ - the Rx pin is used for transmission and reception then. User can either use `Rx == Tx` trick as before or just provide a single pin definition

Fixes: #2538
See also: #1418


fronders:
Hi @fpistm, I'm finishing testing at the moment with different STM32 chips and will report back as soon as I'm done and things are ready to be merged :)

fpistm:
Hi @fronders,

thanks for the PR, will try to review it this week.

fpistm (approved):
LGTM

thanks @fronders and sorry for the delay.

PR #2455, opened 2024-07-18
J-Link Support for STM32F1

jamesy0ung:
**Summary**

This PR implements the support for using a STM32CubeProgrammer with a J-Link from the Arduino IDE

Explain the **motivation** for making this change. What existing problem does the pull request solve?

I have a J-Link probe and can use it directly in STM32CubeProgrammer, but not in the Arduino IDE

**Validation**

I tested it and it works, doesn't seem to break anything else.

close #2452

Issue #2452, opened 2024-07-18
J-Link support when using STM32CubeProgrammer


jamesy0ung:
STM32CubeProgrammer has support for J-Link probes, however the Arduino core only allows it to use SWD with ST-Link, DFU or UART

I'd like to be able to upload using a J-Link probe from the Arduino IDE. Thank you!

![Screenshot 2024-07-18 170617](https://github.com/user-attachments/assets/3b1e67a7-1c82-4d89-b2e6-38f2483f4571)

fpistm:
Hi @jamesy0ung 
Feel free to contribute and provide a PR.

jamesy0ung:
I don't know much about the internals of Arduino Cores, though I'd be happy to if I can get some pointers as to what needs to be modified

I did a search for STM32CubeProgrammer and saw some references to it in platform.txt, however I'm not sure what exactly creates the option in the Arduino IDE and where/how I can extend it.

I also searched up the docs for the STM32CubeProgrammer CLI and it's just a flag passed to it.

jamesy0ung:
I haven't done any work on Arduino cores before, but I got the CLI working by using the following flag. `-c port=JLINK ap=0`

I tried to modify boards.txt to add support but I can't see it in the Arduino IDE as a programmer
```

GenF1.menu.upload_method.jlinkMethod=STM32CubeProgrammer (J-Link)
GenF1.menu.upload_method.jlinkMethod.upload.protocol=swd
GenF1.menu.upload_method.jlinkMethod.upload.options=-c port=JLINK ap=0
GenF1.menu.upload_method.jlinkMethod.upload.tool=stm32CubeProg
```

Is there anything I'm missing?

fpistm:
I would recommend to check the script:
https://github.com/stm32duino/Arduino_Tools/blob/main/stm32CubeProg.sh

And simply add a new interface (protocol): jlink then in the script simply call it with correct option.
About the fact it is not in the menu, this is an Arduino IDE issue. You have to delete the cache.
#### Path:
* Linux: `$HOME/.config/arduino-ide/`
* Windows: `C:\Users\<login>\AppData\Roaming\arduino-ide`

jamesy0ung:
I've opened 2 PR's

One adds a case to the flashing script, and the other adds support under the STM32F1, I'll look into adding J-Link programmer support for other boards, but I don't have any other hardware to test with

Amir1387aht2:
[...]

PR #2621, opened 2025-01-07
feat(platform): extend Black Magic Debug Probe upload method to more boards

ALTracer:
**Summary**

This PR implements the following **feature**

* [ ] Enable BMP upload method in Programmers menu for more boards

Black Magic Debug Probe supports interactive debugging as well as one-shot uploading, via gdb script or multi-argument command line. This is useful with Generic boards which often don't have an onboard debugger. Method was contributed here earlier and carried on for GenF1, GenF3, GenF4, GenL0, GenFlight.
Let's extend this to 3rd party boards, too. Garatronic Pybstick26 was omitted because it only has LEDs on PA13/14 SWD, no testpoints even.

Of course I am unsure about allowing this on Nucleo/Disco/Eval products from ST, which should have a working STLink/V2 or V3E running proprietary firmware; but also P1x6 SWD 2.54mm or STDC-14 headers. If you deem this okay, then it's +20 more lines in a third commit.
BMDA does not create a COM port, it will need a new upload method for tcp socket (like openocd but simpler) which is how one uses BMDA with STLink backend (or CMSIS-DAP, or FTDI MPSSE, or JLink etc.) And of course you don't ship BMDA builds, there's no xPack for it but only GH CI builds and releases.

GenU0 and GenWBA are not supported by BMP (yet), it will treat them as generic Cortex-M without flash reprogramming.

Similar PRs: PR #2528 closing #2437 (CMSIS-DAP), PR #2455 closing #2452 (JLink). I wonder if something systemwide happened that you now have to set per-family `upload_method`s.

Small feature request: is it possible to also Debug via BMP not OpenOCD in Theia IDE? What is required to configure Cortex-Debug for this automatically? Editing conffiles is less convenient. I may open a separate issue on this.

**Validation**

* CI is likely unaffected. Manual review and testing is welcome.
* I've used BMP-compatible debuggers to upload compiled sketches to small boards, i.e. it's now possible to develop on two `blackpill-f411ce` boards or so (one flashed with BMP, one running user sketches) just like two Raspberry Pico (they use picoprobe/debugprobe on one of them but it's not self-contained and needs OpenOCD/pyocd/probe-rs as gdbserver).

**Code formatting**

* No code touched, only `boards.txt`, local `astyle.py` doesn't complain

**Closing issues**

Not opening an issue myself for this.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1738,6 +1738,10 @@ GenC0.menu.upload_method.serialMethod.upload.protocol=serial
 GenC0.menu.upload_method.serialMethod.upload.options=-c {serial.port.file}
 GenC0.menu.upload_method.serialMethod.upload.tool=stm32CubeProg
 
+GenC0.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
+GenC0.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
+GenC0.menu.upload_method.bmpMethod.upload.tool=bmp_upload
+
 GenC0.menu.upload_method.OpenOCDSTLink=OpenOCD STLink (SWD)
 GenC0.menu.upload_method.OpenOCDSTLink.upload.protocol=stlink
 GenC0.menu.upload_method.OpenOCDSTLink.upload.tool=openocd_upload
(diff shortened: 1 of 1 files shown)


dragonmux:
> What do you mean by `BMDA` ?

👋🏼 lead maintainer of Black Magic Debug here, what he's referring to is Black Magic Debug App, just known as BMDA by the project. It is the BMD firmware but built to run as a host program and able to interface with BMPs and other dongles running the firmware, CMSIS-DAP adaptors, J-Link's, ST-Links, and FTDI devices being used as adaptors.

It provides the same GDB server and all the goodness of automatic part discovery, but over TCP/IP (port 2000 by default) instead of a serial device. No config files, just a single binary - `blackmagic-bmda` if using the nightlies and official releases, `blackmagic` if built yourself.

PR #2647, opened 2025-02-01
Update wrong HSE value on variant_B_G431B_ESC1.h

Moddingear:
HSE was set to 24MHz instead of 8MHz, and was overriding anything else trying to set the HSE speed, causing 1ms to be 3ms. Changing it here fixes everything.

See https://www.st.com/en/evaluation-tools/b-g431b-esc1.html#cad-resources, mb1419-g431cbu6-b01, Crystal is 8MHz in schematic.


fpistm (approved):

fpistm:
Thanks @Moddingear.
Nice catch. We didn't see this  during the review and as we do not have this board we never test it.

PR #2712, opened 2025-04-15
feat (uart): implement proper operator bool()

patricklaf:
This PR implements the _HardwareSerial_ bool() operator.

Currently, the core blocks if there is a write to a closed _HardwareSerial_ instance. The tx buffer will rapidly be full and the core waits forever for free space.

The bool() operator is a good way to check if the _HardwareSerial_ object is opened and ready to transmit data but in the current implementation, it allways returns true.

The PR implements a complete  bool() operator:
- Returns true when the _HardwareSerial_ object is ready.
- Returns false when the _HardwareSerial_ object is not ready.

Sample code:

```cpp
Serial.begin();
if (Serial) {
  Serial.println("Serial is opened");
  Serial.flush();
}
Serial.end();
if (Serial) {
  Serial.println("Serial is closed. Never be printed");
}
```

### DIFF
--- a/cores/arduino/HardwareSerial.cpp
+++ b/cores/arduino/HardwareSerial.cpp
@@ -446,13 +446,17 @@ void HardwareSerial::begin(unsigned long baud, byte config)
       break;
   }
 
-  uart_init(&_serial, (uint32_t)baud, databits, parity, stopbits, _rx_invert, _tx_invert, _data_invert);
-  enableHalfDuplexRx();
-  uart_attach_rx_callback(&_serial, _rx_complete_irq);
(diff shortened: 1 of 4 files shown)


fpistm:
Hi @patricklaf 
Thanks for the PR.
It makes sense.
I've made a commit on top of your change to go further in the logic:
https://github.com/fpistm/Arduino_Core_STM32/commit/e07f5af36c183a26bd30a21780caa6270249926c

Please, could you check it and give your feedback.
If ok for you could you add it to your PR and squash all commits in one?

fpistm (approved):
Thanks @patricklaf 
LGTM

patricklaf:
Thank you @fpistm for your help.

fpistm:
Welcome and thanks to contribute.

PR #2726, opened 2025-05-15
Add generic board variant for STM32C071G(8-B)Ux

honzasp:
This PR adds support for generic STM32C071G(8-B)Ux boards. I followed the [guide](https://github.com/stm32duino/Arduino_Core_STM32/wiki/Add-a-new-variant-(board)), the CheckVariant example passes, and a very small example on my board works.

I also found a likely bug in `boards.txt` that prevented USB CDC Serial from working on generic C0 boards, so I also fixed that.


fpistm (approved):
LGTM.
Thanks for fixing missing USB option.

PR #2751, opened 2025-06-09
Fix LINKFLAGS containing a tuple in platformio build script

Stary2001:
This PR fixes an issue in the platformio build script where a tuple containing the arguments is placed into LINKFLAGS. This broke a custom script in a different project (Bosch BSEC2 sensor library) that looks at LINKFLAGS.
I'm not sure how it worked before..

I've tested this locally and it's a minor change.

### DIFF
--- a/tools/platformio/platformio-build.py
+++ b/tools/platformio/platformio-build.py
@@ -376,13 +376,11 @@ if not board_config.get("build.ldscript", ""):
         print("Warning! Cannot find linker script for the current target!\n")
     env.Append(
         LINKFLAGS=[
-            (
-                "-Wl,--default-script",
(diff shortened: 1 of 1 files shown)


Stary2001:
@valeros I've made a simple reproducer here - https://github.com/Stary2001/pio-project-for-pr-2751
It doesn't need anything, apart from lib_deps = boschsensortec/bsec2 - it fails before compiling any source files.

The extra script they use is here: https://github.com/boschsensortec/Bosch-BSEC2-Library/blob/master/extra_script.py#L18 - it iterates over LINKFLAGS to find the mcpu= argument that's being used.

PR #2752, opened 2025-06-17
variant(WL): add Oceanus-I module and EV

wm-eisos:
## Support for Oceanus-I module and EV

**Summary**

This PR adds support for the Wuerth Elektronik Oceanus-I module and EV board which has the STM32WLE5CC MCU.


wm-eisos:
Hey @fpistm,

I just applied code formatting on one file that was causing the astyle workflow to fail, however there was another workflow that was failing which was due to links im not sure how i can fix that.

Best Regards
Wajdi

wm-eisos:
> 2.11.0 planned in the coming weeks. Any requirements about this ?

i would like to add support for this in platform io but since it uses this repo as a dependency then i will wait for it to be released and updated here https://github.com/platformio/platform-ststm32.

Are you involved in that as well?

PR #2754, opened 2025-06-22
fix(uart): prevent deinitializing another UART

thiagogba:
If called from an uninitialized HardwareSerial, the code unintentionally deinit the UART at uart_handlers[0]


fpistm (approved):
Thanks @thiagogba for the PR and the fix.

PR #2791, opened 2025-08-14
Fix build for STM32C051xx boards

nessotrin:
**Summary**

This PR adds the ability to compile for STM32C051xx based boards. Most of the infrastructure was already there.
The STM32C051xx is a recently upgraded version of the STM32C031xx with 2x more flash.

* [ ] Add ldscript (direct copy from STM32C031xx)
* [ ] Fix #ifdef missing the STM32C051xx

The link script still mentions the STM32C031CxTx. Since it is auto-generated I chose to leave it as is. Someone with STM32CubeIDE could generate a new one, but I suspect there would be no difference whatsoever.


fpistm (approved):
Hi  @nessotrin 
I've rework your PR to not modify the HAL GPIO.
I've also completed the STM32C051C addition.

PR #2826, opened 2025-09-26
Define default pins for Serial2

zfields:
Add default pins for Serial2 instance.

## Pull Request template

Please, Make sure that your PR is not a duplicate.
Search among the [Pull request](https://github.com/stm32duino/Arduino_Core_STM32/pulls) before creating one.

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/stm32duino/Arduino_Core_STM32/blob/main/CONTRIBUTING.md) file for detailed contributing guidelines.

Thanks for submitting a pull request.
Please provide enough information so that others can review your pull request:

**Summary**

* [ ] Bug 1
* [ ] Bug 2
* [ ] Feature 1
* [ ] Feature 2
* [ ] Breaking changes

**Validation**

* Ensure CI build is passed.

**Code formatting**

**Closing issues**

Fixes #xxx


fpistm (changes requested):

zfields:
Thank you for the correction!

I think I went wrong, because I tested this out on hardware, and it appeared to work correctly. :exploding_head: 

Can you explain why it "worked" even though I was wrong, and where should I be looking to know if the standard or `_ALT1` version of pins should be used?

I don't want to continue to make these errors as we evolve our product.

Thank you again!

fpistm:
> I think I went wrong, because I tested this out on hardware, and it appeared to work correctly. 🤯
> 
> Can you explain why it "worked" even though I was wrong, 

Of course it work but with LPUART1 not USART2 as you expected :wink:

> and where should I be looking to know if the standard or `_ALT1` version of pins should be used instead?

To sum up a pin (GPIO) can have several peripherals and also several instance of one type of peripheral.

You can find in the datasheet all alternate functions of one pin:
image

The PeripheralPins.c simply list all possible values per features.
Example for PA2:
https://github.com/stm32duino/Arduino_Core_STM32/blob/f1432b05b1a919ed270b6edd8b61a55e6a369c7b/variants/STM32L4xx/L433C(B-C)(T-U)_L443CC(T-U)/PeripheralPins_CYGNET.c#L129-L130

In this case we found the 2 possibilities for serial Tx and the required alternate function (AF):
`GPIO_AF8_LPUART1` or `GPIO_AF7_USART2`.
To be able to select the one we want for PA2, the suffix ALTx (for ALTernative), pay attention it is different of the alternate function.
It only allows to get the right entry in the table and then configure the GPIO pin accordingly.

variants/STM32L4xx/L433C(B-C)(T-U)_L443CC(T-U)/variant_CYGNET.h
```diff
@@ -153,6 +153,14 @@
   #define PIN_SERIAL_TX         PA9
 #endif
 
+// Default pin used for generic `Serial2` instance
+#ifndef PIN_SERIAL2_RX
+  #define PIN_SERIAL2_RX         PA3
```
zfields, line 158:
```suggestion
  #define PIN_SERIAL2_RX         PA3_ALT1
```

variants/STM32L4xx/L433C(B-C)(T-U)_L443CC(T-U)/variant_CYGNET.h
```diff
@@ -153,6 +153,14 @@
...
+#endif
+#ifndef PIN_SERIAL2_TX
+  #define PIN_SERIAL2_TX         PA2
```
zfields, line 161:
```suggestion
  #define PIN_SERIAL2_TX         PA2_ALT1
```

zfields:
I've updated to use `PA2_ALT1` and `PA3_ALT1`.

Defined here:
https://github.com/zfields/Arduino_Core_STM32/blob/patch-2/variants/STM32L4xx/L433C(B-C)(T-U)_L443CC(T-U)/variant_CYGNET.h#L63-L64

fpistm (approved):

PR #2873, opened 2025-12-08
feat: Blues Heron (STM32U575CIT6)

zfields:


boards.txt
```diff
@@ -16822,13 +16839,13 @@ GenWL.menu.dbg.enable_all.build.flags.debug=-g
...
-Blues.menu.dbg.none=None
+Blues.menu.dbg.enable_all=Core Logs and Symbols Enabled (-g)
+Blues.menu.dbg.enable_all.build.flags.debug=-g
 Blues.menu.dbg.enable_sym=Symbols Enabled (-g)
 Blues.menu.dbg.enable_sym.build.flags.debug=-g -DNDEBUG
 Blues.menu.dbg.enable_log=Core logs Enabled
 Blues.menu.dbg.enable_log.build.flags.debug=
-Blues.menu.dbg.enable_all=Core Logs and Symbols Enabled (-g)
-Blues.menu.dbg.enable_all.build.flags.debug=-g
+Blues.menu.dbg.none=None
```
fpistm: I don't recommend to change this order.
If you need to debug, the correct way is to enable the "Optimize for debugging" in the Arduino IDE menu. Then debug and symbol is activated.

zfields:
@fpistm Does the ST Low Power Library work with the U5?

All my tests are passing, except I cannot achieve Low Power using the same sketch I use on both the Cygnet and Swan.

Do you see anything wrong with my board configuration that would prevent me from achieving low power?

variants/STM32U5xx/U575C(G-I)(T-U)_U585CI(T-U)/variant_HERON.h
```diff
@@ -0,0 +1,246 @@
...
+#define PB10                      26      // USART3_VCP_TX
+#define PB13                      10      // D10
+#define PB14                      9       // D9
+#define PB15                      12      // D12
+#define PC13                      8       // USER_BTN
+#define PC14                      34      // OSC32_IN (LSE)
+#define PC15                      35      // OSC32_OUT (LSE)
+#define PH0                       27      // ENABLE_3V3
+#define PH1                       28      // DISCHARGE_3V3
+#define PH3                       20      // B
```
fpistm: Hard to check if all pin have an unique number as not ordered. Anyway I guess you have tested.
You can use this example to check if definition is correct:
https://github.com/stm32duino/STM32Examples/tree/main/examples/NonReg/CheckVariant

boards.txt
zfields: In my mind, you want to start a project with debug symbols, then once you have hardened your implementation, you will either optimize for size or speed.

This follows the precedent of adding `-DNDEBUG` to your compilation, only after you are happy with what you've built and you are ready to release a production build.

This will also allow our users to simply press the debug button and start debugging without having to "learn" how to set it up.

That being said, if this is a STM32duino experience faux pas, and you are not asking, but telling me to change it, then I will change it, no problem.

PR #2889, opened 2026-01-06
chore(pio): clean up cmsis package handling

fpistm:
First goal of this PR would be to use CMSIS 6 for pio build.
After some study, the build does not use the CMSIS package provided within the core but its own one which is version 5.9.0.
So I've remove the download of the CMSIS package provided within the core as it is useless.
CMSIS 6 support could be added if pio update to cmsis 6. 

/cc @valeros

### DIFF
--- a/.github/actions/pio-build/README.md
+++ b/.github/actions/pio-build/README.md
@@ -2,12 +2,6 @@
 
 This action build thanks PIO.
 
-## Inputs
-
-### `cmsis-version`
-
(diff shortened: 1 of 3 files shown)


valeros:
Hi @fpistm, thanks for pinging. Indeed, we install CMSIS separately, so as long as the Arduino Core doesn’t strictly require a newer version of CMSIS, we’re good.

PR #2919, opened 2026-03-02
feat(usb): integrate PluggableUSB-based USB stack for STM32

Aymane-ST:
## Summary

This PR integrates an Arduino-like **PluggableUSB** architecture into the STM32duino core USB stack, based on the work from [Levi--G/USBLibrarySTM32](https://github.com/Levi--G/USBLibrarySTM32).

The goal is to provide a PluggableUSB-compatible USB stack directly in the core.
This work follows and implements the ideas discussed in:

- **#2724 – USB discussion (Levi’s PluggableUSB / external USB library)**  

/cc @Levi--G


pschatzmann:
Hallo, I was adding USB class 2 audio support to my [AudioTools library](https://github.com/pschatzmann/arduino-audio-tools) which is based on TinyUSB.

I was considering to add support for the STM32 USB API, but I noticed that there is no clean extension mechanism. Could you  please give me a quick status of your work ? Is there a particular reason why this is stuck ?

Aymane-ST:
> Hallo, I was adding USB class 2 audio support to my [AudioTools library](https://github.com/pschatzmann/arduino-audio-tools) which is based on TinyUSB.
> 
> I was considering to add support for the STM32 USB API, but I noticed that there is no clean extension mechanism. Could you please give me a quick status of your work ? Is there a particular reason why this is stuck ?

Hi, thanks for your message.
We are currently studying the possibility of moving to TinyUSB instead of the current ST USB stack, as it could provide a cleaner and lighter integration path.
At the moment, the work is still under analysis, so there is no final decision yet.

pschatzmann:
My 5 cents on this: I would suggest to go ahead with this and just add a bit of documentation that highlights that TinyUSB can be used by setting USB Support to NONE. It took me quite a while to figure that one out.

This gives the best of both worlds and frees you from being dependent on the release cycle of TinyUSB...

PR #2922, opened 2026-03-06
I3C Controller Library

Aymane-ST:
Depends on LPS22DF pr: https://github.com/Aymane-ST/LPS22DF/pull/1

Tested on Nucleo H503RB and U385RG_Q


cparata (commented):
Hello Aymane,
regarding the example Entdaa, is it possible to implement it using the official LPS22DF library that you patched to add the I3C support? I think that the logic flow in this case should be:

I3C_BUS.begin();
I3C_BUS.rstdaa();
I3C_BUS.disec();
I3C_BUS.entdaa();
// 3) Identify the LPS22DF via WHO_AM_I
lps22df_instance.begin();

It is cleaner and perfectly aligned with the use case of the SETDASA and you don't need to use the helper files lps22df.cpp and lps22df.h. Maybe we can use a flag in the I3C constructor of the component to distinguish the use case where we use the SETDASA and the use case where we use the ENTDAA. So, we can eventually skip in the begin the SETDASA calls when we want to use the ENTDAA. I'm thinking that we can provide as optional in the LPS22DFClass I3C constructor the "staticAddr7" parameter; if it is not passed, we suppose that we are in the ENTDAA use case; if it is passed, we suppose that we are in the SETDASA use case. 
What about this proposal? Is it feasible according to you?

PR #2958, opened 2026-04-17
PA13 USB NOE commentiong out for L5

novaksta:
**Summary**

Comment out by default PA13. PA13 is SWDIO pin (very useful feature), however currently when using USB it PA13 gets NOE (very specific feature, typically not used).

Please also consider commenting out fro other devices, as PA13 is more useful than SWDIO vs. NOE.

**Validation**
Verified on L5 Nucleo board.


fpistm:
Hi @novaksta 
I understand your point. Anyway, this file is automatically generated and so next time the variants will be generated your change will disappear.
By default the generic provide access to all pins and so user have to configure for their needs.
By adding a dedicated variant and so a dedicated peripheral pins files where you can comment any line you want.
Or by defining at sketch level the `PinMap PinMap_USB[]` as it is a weak one.

Later I've already plan to split this array per signal like SDIO/SMMC one to ease config and so NOE could be init or not by default.

novaksta:
Hello @fpistm,
thank you for the suggestion -- workaround through definition PinMap_USB[], is often feasible, anyway you need to know such. Changing configuration may be a way -- but still it would be nice not confronting directly with the PeripheralPins.c when debugging (very often the case).

fpistm:
Will take this in account to the next update of USB pins usage.

PR #2961, opened 2026-04-23
feat: Blues Heron

zfields:
Adds support for the Blues Heron Feather DevKit (STM32U575OIY6QTR).


fpistm (changes requested):
Hi @zfields 
I've added a commit to clean the pins definitions.
As it seems not all pins are available then no need to have them in the `digitalPin` array.
I've also reorder to ease reading.

Feel free to kept or not the reordering but at kept the `digitalPin` change and `NUM_DIGITAL_PINS` changes.

zfields:
Adopted your adjustment, but reordered the defines (by name instead of index); maintaining alignment with our other board definitions (e.g. Swan, Cygnet).

_When they are ordered by name, it makes it easier for me to cross reference the schematic._

fpistm (approved):
I've squashed some commits to clean up the PR.
LGTM.
Ready to be merged. Is it ok for you @zfields ?

zfields:
Yes. It's ready to go when you are.

PR #2977, opened 2026-05-22
feat: add compile flag for board family

### LINKED ISSUE #2976: Request a compile flag for board family
I am trying to write test programs that are as portable as possible across hardware platforms.  It would be really helpful to add a compile flag for the board name (which is really a board family), like `-DARDUINO_NUCLEO_32`.

As a concrete example, Nucleo-32 boards do not have a user button, Nucleo-64 boards have an active low user button, and Nucleo-144 boards have an active high user button.  I have no way to figure out which family _e.g._ NUCLEO_F767ZI is in.

fpistm:
Ex:
`-DARDUINO_NUCLEO_64`

By Default for all Gerneric:
`-DARDUINO_GENERIC
`
Fixes #2976.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -18,6 +18,7 @@ menu.upload_method=Upload method
 Nucleo_144.name=Nucleo-144
 Nucleo_144.build.core=arduino
 Nucleo_144.build.board=Nucleo_144
+Nucleo_144.build.family=NUCLEO_144
 Nucleo_144.build.variant_h=variant_{build.board}.h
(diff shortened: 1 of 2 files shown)


pmunts:
Looks great to me.  I actually started working on this myself this morning so I was surprised to get your notice about this.  I was prepending to *.build.st_extra_flags; that didn't require changing platforms.txt.

fpistm:
Adding to st_extra_flags was an other solution but it is often override at board level. So it requires to add the define for all boards which override it. The way done in this PR requires only one define per high level menu and allows a default value for all generic menu.

PR #3000, opened 2026-06-26
I3C: add auto bus resolution, init/IRQ fixes, and memory cleanup

Aymane-ST:
This PR improves the I3C library by adding automatic instance selection from SDA/SCL pins, fixing controller initialization and IRQ routing, reducing temporary buffer usage, and aligning clock-source ownership with variant clock configuration.

It also updates examples to use the generic I3C object and removes the temporary stack-size workaround that is no longer needed.


cparata (approved):
LGTM

fpistm (approved):
LGTM.
Thanks for all iterations.

cparata (approved):
LGTM

PR #3024, opened 2026-07-15
fix(wrapper): enhance util(s) file management

fpistm:
- Use 'utils' naming for generated header files.
- Add 'utils' source file


Aymane-ST (approved):
LGTM :)

PR #3028, opened 2026-07-21
feat: update stm32cubeprog options

fpistm:
This PR includes:
- chore: use -b instead of -f for STM32CubeProgrammer
- feat: add freq option for swd upload
  Ex: `Nucleo_64.menu.pnum.NUCLEO_C562RE.upload.freq=3300`
- feat: add stlink_sn option for swd upload
    It allows to specify serial number of the ST-Link device (if multiple ST-Link devices are connected)   
    Goal is to use it over the boards.local.txt as it make non-sense to hardcode ST-Link SN in the boards.txt.
    Ex:  `Nucleo_64.menu.pnum.NUCLEO_C562RE.upload.stlink_sn=003100344142501020353451`

See https://github.com/stm32duino/Arduino_Tools/pull/119


Aymane-ST (approved):
LGTM : )

PR #3040, opened 2026-08-09
Add support register-based SPI slave

### LINKED ISSUE #3039: Add register-based SPI slave support with example code

The current SPI slave support needs to expose additional functionality in order for the user code to implement a register-based SPI slave.  

In a register-based SPI slave, the slave maintains a list of registers that the master can read or write.  The registers can be read to return state (e.g. temperature or mode) or written to change state (e.g. set mode, control outputs). 

By adding a method to read and write one byte at a time, the slave's code can implement a complete register-based SPI slave.  The slave code works by waiting for an interrupt from the master on SS.  It then read a byte containing the read/write bit and an address.  It then either reads the data to be written to the given address, or returns the data read from the given address.

Using the spi_transfer() method doesn’t work for this use case because the slave does not know in advance if the master is starting a read or a write request and must determine that on the fly.

**Additional context**

I have put together a fork with the necessary changes along with some sample code to demonstrated it.  I have also added timeouts to the new method so that the slave does not get stuck in an infinite while loop if something goes wrong.

I also added a method that can be used to reset the spi interface so that the slave can recover if it falls behind or detects other errors that cause it to fall out sync with the master.

philrittenhouse:
By adding a method to the SPI library to read and write one byte at a time, the slave's code can implement a complete register-based SPI slave. The new method includes timeouts so that the slave does not get stuck in an infinite while loop if something goes wrong.

An additional reset() method can be used to reset the spi interface This allows the slave to recover if it falls behind or detects other errors that cause it to fall out sync with the master.

Example code showing a simple register-based slave is included in the **SPI/examples/RegisterSlave** folder along with a python script that runs on a RPi and acts as the master.

Resolves: #3039
See Also: #2050

## Pull Request template

**Summary**

* [ ] Add support for register-based SPI slave
* [ ] Add SPI slave example code

The current SPI slave support needs to expose additional functionality in order for the user code to implement a register-based SPI slave.

In a register-based SPI slave, the slave maintains a list of registers that the master can read or write. The registers can be read to return state (e.g. temperature or mode) or written to change state (e.g. set mode, control outputs).

By adding a method to read and write one byte at a time, the slave's code can implement a complete register-based SPI slave. The slave code works by waiting for an interrupt from the master on SS. It then read a byte containing the read/write bit and an address. It then either reads the data to be written to the given address, or returns the data read from the given address.

Using the spi_transfer() method doesn’t work for this use case because the slave does not know in advance if the master is starting a read or a write request and must determine that on the fly.

**Validation**

* Ensure CI build is passed.

**Code formatting**

**Closing issues**

Fixes #3039 
Closes: #3039


fpistm:
Hi @philrittenhouse 
Honestly, I do not understand well the goal of your PR.
Seems to me it is a workaround and does not follow the SPI protocol.

In your example you unconditionally call reset.
Reset function is closed to the deinit one except clock are not disable and hal not deinit but as the init is called in your reset function why not simple call deinit then init? I'm not confident that the HAL spi handle state will be ok after several call.

The read/write function seems very than the transfet one except the last step to wait EOT.

About example, it is very specific, requires STM32FreeRTOS and RPi to run examples.

philrittenhouse:
Hi @fpistm 

Thank you for reviewing my PR!

> Honestly, I do not understand well the goal of your PR. Seems to me it is a workaround and does not follow the SPI protocol.

Sorry, I probably did not explain myself well.   The goal is to allow an STM32 device to emulate the same kind of SPI interface that devices like the BME280 environmental sensor or MCP4131 Digital Potentiometers use.  That is, the master reads and writes registers on the device in order to control their state or read back status.

> In your example you unconditionally call reset. Reset function is closed to the deinit one except clock are not disable
> and hal not deinit but as the init is called in your reset function why not simple call deinit then init? 
Yes, when the SS signal is raised  by the master the slave is interrupted.  The ISR resets the interface to flush any stale data and make sure it is in a known state.  This must be done very quickly so I created reset() as a light-weight version of calling deinit and init back to back.

> I'm not confident that the HAL spi handle state will be ok after several call.
I haven't run into any issues in my testing but that is worth looking in to.

> The read/write function seems very than the transfet one except the last step to wait EOT.
Yes.  Perhaps we could extend the existing transfer method in order to conditionally remove the wait for EOT.  I was reluctant to change the existing methods though, for fear of breaking existing code.   I'd appreciate any suggestions.

> About example, it is very specific, requires STM32FreeRTOS and RPi to run examples.
Yes, my intention was only to provide a guide that others could follow since their specific environment could be quite different.  I think it would be difficult to implement the slave code without the RTOS extensions but I could make a version of the master side that could be run on a second Nucleo if you think that would be more helpful than the RPi python script.

I will go through the rest of the comments and make the necessary changes.

Thanks again,
Phil

philrittenhouse:
It looks like the Platform IO build is failing because my demo has a dependency on STM32FreeRTOS.  I am not sure what to do about that.

fpistm:
> Sorry, I probably did not explain myself well. The goal is to allow an STM32 device to emulate the same kind of SPI interface that devices like the BME280 environmental sensor or MCP4131 Digital Potentiometers use. That is, the master reads and writes registers on the device in order to control their state or read back status.

OK. For me it is an application and should not be added as part of the SPI library. I found nothing like that in any other core.
Moreover example is very too specific and require STM32FreeRTOS as dependencies.
I suggest you create a library in your repo and inherit  from SPI Class.

philrittenhouse:
> I suggest you create a library in your repo and inherit from SPI Class.

OK I will do that instead

PR #3063, opened 2026-08-29
feat: reboot into the ST system bootloader on a 1200 bps touch

### LINKED ISSUE #706: Automatic restart in DFU Bootloader mode
When using the built in USB DFU bootloader, lets auto-reset into the bootloader mode from USB Serial using something like 1200-baud knocking

mikeysklar:
**Summary**

Makes the STM32CubeProgrammer (DFU) upload method work on its own: closing the USB CDC port at 1200 bps now reboots the board into the ST system memory bootloader, so a sketch can be uploaded without touching BOOT0 or power cycling.

* [x] Feature 1: report the USB 1200 bps touch through a weak hook
* [x] Feature 2: enter the ST system bootloader on that touch, for STM32F4
* [ ] Breaking changes

Fixes #706

**Motivation**

`dfuMethod` already exists for essentially every series, so the board can be *flashed* over DFU. What is missing is anything to *put* it into DFU, which is why uploading still means reaching for BOOT0. This closes that gap.

Two commits, each self-contained:

1. `feat(cdc): add a 1200 bps touch hook` — a weak `cdc_1200bps_touchHook()` in `cores/arduino/hooks.c`, called from `CDC_SET_CONTROL_LINE_STATE` when the host closes the port at 1200 bps. It follows `yield()` and `dtr_togglingHook()`, has no series specific content, and does nothing by default. `dtr_togglingHook()` cannot serve here: it is invoked from `USBD_CDC_Receive`, so it needs the host to send data, and a 1200 bps touch sends none.
2. `feat(bootloader): add ST system bootloader entry` — the policy, behind `-DBL_SYSTEM` on the existing `dfuMethod` entry.

**Design notes, each of which cost a measurement**

- **The branch cannot happen in the hook.** It runs in handler mode, so setting MSP and branching leaves the ROM inside an exception that never returns. Same jump code on the same board: **0 of 8** from inside the hook, board hangs until a power cycle, against **5 of 8** deferred to thread mode. The hook therefore only records the request and calls `NVIC_SystemReset()`, which *is* safe from handler mode.
- **`.noinit`, not a backup register.** Some series have none, and `.noinit` holds an arbitrary value after a power cycle, so the request is only trusted after a software reset, which is also what makes leaving DFU clear it rather than re-enter.
- **`premain()` for the branch**, before `init()` calls `HAL_Init()`, so the ROM gets the part close to reset state. No `Reset_Handler` override and no startup assembly change.
- **The request is retried, gated on a software reset.** The ROM clocks USB from the HSE but does not know which crystal is fitted, so it measures one against the HSI, and per AN2606 Rev 61 Figure 32/33 it resets the part when that misses: `HSE detected` -> no -> "Generate System reset". Section 28.2.1 says lower frequencies detect better, "it is better to use 8 MHz instead of 25 MHz"; this board is 12 MHz. A single attempt reached DFU **5 times in 12**. Retrying reaches it every time. The software reset gate is what makes leaving DFU clear the request rather than re-enter it. Pre-starting HSE and waiting for `HSERDY` does not help, so this is not a crystal startup race.

**Series coverage**

`systemBootloaderAddress()` remaps system memory to zero with `__HAL_SYSCFG_REMAPMEMORY_SYSTEMFLASH()` and returns 0, rather than carrying a table of per series addresses. That macro is provided by the HAL for F0, F2, F3, F4, C0, G0, G4, L0, L1 and L4 among others, so the implementation is not F4 specific. Where it is absent the `#else` is a compile time error rather than a silent no-op, and since `-DBL_SYSTEM` is only set by the `boards.txt` entry, no series that does not opt in is affected. `systemBootloaderAddress()` is `WEAK` so a board with the bootloader somewhere else can redefine it.

I have only been able to verify this on STM32F4. I enabled the flag for GenF4 alone for that reason, rather than turning it on for series I cannot test.

**Validation**

Board: Adafruit Feather STM32F405 Express, 12 MHz crystal. Core built from this branch. FQBN `STMicroelectronics:stm32:GenF4:pnum=FEATHER_F405,usb=CDCgen,xusb=FS,upload_method=dfuMethod`. Host: Linux, board on a hub with per port power control, flashed over SWD with OpenOCD for the initial load.

The sketch is a stock blink with **no bootloader code at all**, which is the point:

```cpp
void setup() { Serial.begin(115200); pinMode(LED_BUILTIN, OUTPUT); }
void loop() { digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN)); delay(200); }
```

Each trial power cycles the port, waits for enumeration, then opens the port at 1200 bps, drops DTR and closes it. Entry is read from `/sys/bus/usb/devices/.../idProduct`.

| test | result |
|---|---|
| touch reaches DFU, 12 trials | **12 / 12** |
| full round trip: touch, `dfu-util ... :leave`, back to sketch | **3 / 3** |
| the hook itself firing on a touch | 10 / 10 |
| same core built with `upload_method=swdMethod` | builds, 176 bytes smaller, feature absent |

**Not tested:** series other than STM32F4, and hosts other than Linux.

**Code formatting**

`astyle --options=CI/astyle/.astylerc` reports every changed file unchanged. No generated or vendored file is touched. `cmake/scripts/cmake_updater_hook.py` was re-run after the `boards.txt` change, with submodules checked out, and produced no diff.

**Prior art**

#710 covered this ground in 2019. Credit to @fpistm and @matthijskooijman: the `.noinit` plus software reset gate, the `premain()` placement, the `WEAK` address function that remaps rather than tabulating addresses, and the inline asm branch all follow that branch. This is written fresh against current `main` rather than rebased on it, since the file layout has moved since and that branch is unfinished.

Claude Code was used. The measurements above are USB enumeration state read from sysfs, on the hardware named, not estimates.

### DIFF
--- a/libraries/USBDevice/src/cdc/usbd_cdc_if.c
+++ b/libraries/USBDevice/src/cdc/usbd_cdc_if.c
@@ -60,6 +60,8 @@ __IO bool rtsState = false;
 __IO bool receivePended = true;
 static uint32_t transmitStart = 0;
 
+extern void cdc_1200bps_touchHook(void);
+
 #ifdef DTR_TOGGLING_SEQ
   /* DTR toggling sequence management */
   extern void dtr_togglingHook(uint8_t *buf, uint32_t *len);
@@ -195,6 +197,11 @@ static int8_t USBD_CDC_Control(uint8_t cmd, uint8_t *pbuf, uint16_t length)
         transmitStart = 0;
       }
       rtsState = (((USBD_SetupReqTypedef *)pbuf)->wValue & CLS_RTS);
+      /* Host closing the port at 1200 bps is the Arduino convention for
+         requesting a reboot into the bootloader. */
+      if (!dtrState && linecoding.bitrate == 1200) {
+        cdc_1200bps_touchHook();
+      }
 #ifdef DTR_TOGGLING_SEQ
       dtr_toggling++; /* Count DTR toggling */
 #endif
--- a/boards.txt
+++ b/boards.txt
@@ -6146,6 +6146,7 @@ GenF4.menu.upload_method.dfuMethod=STM32CubeProgrammer (DFU)
 GenF4.menu.upload_method.dfuMethod.upload.protocol=dfu
 GenF4.menu.upload_method.dfuMethod.upload.options=-v {upload.vid} -p {upload.pid} -a {upload.address} -s {upload.start}
 GenF4.menu.upload_method.dfuMethod.upload.tool=stm32CubeProg
+GenF4.menu.upload_method.dfuMethod.build.bootloader_flags=-DBL_SYSTEM
 
 GenF4.menu.upload_method.bmpMethod=BMP (Black Magic Probe)
 GenF4.menu.upload_method.bmpMethod.upload.protocol=gdb_bmp
--- a/cores/arduino/hooks.c
+++ b/cores/arduino/hooks.c
@@ -50,3 +50,25 @@ static void __empty_dtr_toggling(uint8_t *buf, uint32_t *len)
 }
 void dtr_togglingHook(uint8_t *buf, uint32_t *len) __attribute__((weak, alias("__empty_dtr_toggling")));
 #endif
+
+#if defined(USBCON) && defined(USBD_USE_CDC)
+/**
+ * Empty cdc_1200bps_touch() hook.
+ *
+ * Called when the host closes the CDC port while the line coding is set to
+ * 1200 bps, which is the Arduino convention for asking the board to reboot
+ * into its bootloader.
+ *
+ * Its defined as a weak symbol and it can be redefined to implement the
+ * bootloader entry a given board needs.
+ *
+ * It is called from the USB control transfer callback, so it runs in handler
+ * mode. Branching to a bootloader from here leaves the core inside an
+ * exception that never returns; schedule the work and perform it from thread
+ * mode or after a reset instead.
+ */
+static void __empty_1200bps_touch(void)
+{
+}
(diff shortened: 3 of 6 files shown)


libraries/USBDevice/src/cdc/usbd_cdc_if.c
```diff
@@ -195,6 +197,11 @@ static int8_t USBD_CDC_Control(uint8_t cmd, uint8_t *pbuf, uint16_t length)
         transmitStart = 0;
       }
       rtsState = (((USBD_SetupReqTypedef *)pbuf)->wValue & CLS_RTS);
+      /* Host closing the port at 1200 bps is the Arduino convention for
+         requesting a reboot into the bootloader. */
```
fpistm, line 201: Based on https://arduino.github.io/arduino-cli/1.5/platform-specification/#1200-bps-bootloader-reset, it is not the convention or did I miss something?

> use_1200bps_touch causes the selected serial port to be briefly opened at 1200 bps (8N1) before starting the upload.

The #3078 seems more inline with this feature.

mikeysklar: The spec line only covers the host side. arduino-cli's touch ([`Touch1200bps`](https://github.com/arduino/go-serial-utils/blob/HEAD/reset.go#L20-L38)) opens the port at 1200, drops DTR and closes it. ArduinoCore-samd triggers on that same pair, 1200 plus DTR low ([CDC.cpp](https://github.com/arduino/ArduinoCore-samd/blob/HEAD/cores/arduino/USB/CDC.cpp#L146-L151)), and this PR follows it.

fpistm: Ok. Thanks for the link. I will check.

PR #3074, opened 2026-09-07
fix(Serial): Uart::write() must not spin when begin() failed

### LINKED ISSUE #3071: [bug] Uart::write() deadlocks when Serial.begin() fails (LPUART with unsupported baud/clock)
First of all, I would like to appreciate all contributors on this project so much!
Without this, I(and many others) couldn't get STM32WL supported on Meshtastic(yes, it now runs on STM32WL!)

---

### Describe the bug

When `Serial.begin(baud)` fails during hardware initialization, `uart_init()` returns `false` and `_ready` is set to `false`.
However, `Uart::write()` in `cores/arduino/Serial.cpp` does not check `_ready` before entering its transmit path.

After a failed `begin()`, subsequent `write()` / `print()` calls proceed normally until the 63-byte TX ring buffer gets filled up.
At that point, `write()` enters an infinite spin:

Because the hardware was never successfully initialized, the TX interrupt never fires, `tx_tail` never advances, and the MCU stalls permanently.

---

### Steps to reproduce

1. Target an STM32 with LPUART1 (e.g. STM32WLE5 / Wio-E5 mini).
2. Have LSE selected as the LPUART1 kernel clock source (typical when RTC or low-power wake-up is configured).
3. Call `begin()` with a baud rate that LSE cannot satisfy:
   ```cpp
   SerialLP1.begin(38400); // fails: LSE (32.768 kHz) < 3 × 38400 = 115,200 Hz
   SerialLP1.println("This will freeze the CPU once the TX buffer fills.");
   ```
4. MCU deadlocks inside `Uart::write()` once the 63-byte TX buffer is exhausted.

---

### Context / Background

This was observed in [Meshtastic](https://github.com/meshtastic/firmware) , attempts on enhancing variant support for Seeed Wio-E5 mini (STM32WLE5, LPUART1 on PC0/PC1 for GPS/GNSS).
The GPS/GNSS driver probes common baud rates sequentially (9600 -> 38400 -> 115200).
When LSE is active as the LPUART1 kernel clock, the 38400 probe triggers the deadlock immediately on boot sequence.

On a side note, we can't use USART2 on PA_2_ALT1/PA_3_ALT1 that's available on Wio-E5/mini, mainly due to DFU's quirks.
This is also Wio-E5/LoRa-E5 specific context at this moment,
other STM32WL variant in meshtastic uses other pins w/USART for GPS/GNSS.

As this is a physical limitation, I've been testing some naive codes to switch the kernel clock to HSI, and now GPS/GNSS module works in all baud rate above.
(ref: https://github.com/meshtastic/firmware/compare/develop...t-miura:firmware:fix/stm32wl-lpuart-clock-wio-e5-gps , note that currently this will in inf. loop if HSI never becomes ready)

### Why is the `_ready` guard still needed even after switching the kernel clock to HSI?

There are two reasons as far as I concern:

#### 1. What if we failed to switch the kernel clock, and LPUART runs on LSE?

It's very, very unlikely to happen, but there are no guarantee that this won't be happen.
While we're trying to prevent LPUART-on-LSE being used on GPS/GNSS driver, but if `_ready` guard implemented, it'll be much reliable.

#### 2. `uart_config_lowpower()` concerns

Note that I haven't actually observed this possible issue, so it's low on my reason list:

First, I've read #2897, and agree with that line:
> It is up to end user to properly configure the clock depending on their needs.

Also `uart_config_lowpower()` function is available to switch LPUART1's kernel clock from LSE to HSI before re-calling `begin()`.

However, there is a code path in that function (under `__HAL_RCC_LPUART1_CLK_CONFIG`) that actually configures LSE,
not HSI due to a conditional compilation mismatch(`uart.c` lines 911–914).
If the system compiles through that branch, the "switch to HSI" silently does nothing or configures the wrong source, `begin()` fails again, and any subsequent `write()` deadlocks identically.

More broadly, `uart_config_lowpower()` is a separate opt-in call. 
Standard Arduino libraries (TinyGPS, sensor drivers, etc.) call `begin()` and `write()` directly. 
They have no reason to know that a particular `begin()` might fail or that a clock switch is needed first.

---

### Expected behavior

If `_ready` is `false`, `Uart::write()` should return `0` immediately rather than spinning on an interrupt that will never fire.

### Suggested Fix

Guard both `write()` overloads in `cores/arduino/Serial.cpp` with a `_ready` check:

```cpp
size_t Uart::write(const uint8_t *buffer, size_t size)
{
  if (!_ready) {
    return 0;
  }
  // ... existing implementation
}

size_t Uart::write(uint8_t c)
{
  uint8_t buff = c;
  return write(&buff, 1);  // delegates — no separate guard needed here
}
```

Since `write(uint8_t c)` already delegates to `write(const uint8_t*, size_t)`, only the latter needs the guard.

Thank you and kind regards.

saikumar-mandaji:
## Bug

If `Serial.begin(baud)` fails during hardware init (e.g. an LPUART clocked from LSE can't reach the requested baud rate), `uart_init()` returns `false` and `_ready` is set to `false` — but `Uart::write()` never checks `_ready` before entering its transmit path.

`Uart::write(const uint8_t*, size_t)` contains:
```cpp
while (!availableForWrite()) {
    // nop, the interrupt handler will free up space for us
}
```
Since the hardware was never actually brought up, the TX interrupt this loop waits on never fires, so once the 63-byte TX ring buffer fills up (a few `Serial.print()` calls after a failed `begin()`), this spins forever and the MCU deadlocks permanently. See #3071 for the full repro (STM32WLE5 LPUART1 on LSE, GPS driver baud-probing 9600 -> 38400 -> deadlock).

## Fix

Return `0` immediately if `!_ready`, before touching the buffer or entering the wait loop — this is exactly the guard suggested in #3071. `write(uint8_t)` needs no separate guard since it already delegates to this overload.

## Verification

- No hardware was available to reproduce the deadlock physically.
- I attempted a real compile check with `arm-none-eabi-g++ -fsyntax-only` against the actual STM32WLxx variant/CMSIS/HAL headers in this repo, but this checkout is missing the CMSIS-Core headers (`core_cm4.h` etc.) that a full Arduino IDE/arduino-cli install would normally fetch separately — they aren't vendored in the git tree itself, so a from-scratch clone can't be fully compiled without that tooling. I'm disclosing this honestly rather than claiming a green compile I didn't actually get.
- The change itself is a minimal, single early-return guard using the pre-existing `_ready` member (already set in `Uart::begin()` and already exposed via an accessor in `Serial.h`, both in the same file/class), so the risk profile is low even without a full build.

Fixes #3071

### DIFF
--- a/cores/arduino/Serial.cpp
+++ b/cores/arduino/Serial.cpp
@@ -583,62 +583,69 @@ void Uart::flush(uint32_t timeout)
 
 size_t Uart::write(const uint8_t *buffer, size_t size)
 {
-  size_t size_intermediate;
   size_t ret = size;
-  size_t available = availableForWrite();
-  size_t available_till_buffer_end = SERIAL_TX_BUFFER_SIZE - _serial.tx_head;
+  // If begin() never brought the hardware up (e.g. the requested baud
+  // rate can't be reached from the configured kernel clock), the TX
+  // interrupt this function waits on below will never fire. Bail out
+  // instead of spinning forever once the TX buffer fills up.
+  if (!_ready) {
+    ret = 0;
+  } else {
+    size_t size_intermediate;
+    size_t available = availableForWrite();
+    size_t available_till_buffer_end = SERIAL_TX_BUFFER_SIZE - _serial.tx_head;
 
-  _written = true;
-  if (isHalfDuplex()) {
-    if (_rx_enabled) {
-      _rx_enabled = false;
-      uart_enable_tx(&_serial);
+    _written = true;
+    if (isHalfDuplex()) {
+      if (_rx_enabled) {
+        _rx_enabled = false;
+        uart_enable_tx(&_serial);
+      }
+    }
+
+    // If the output buffer is full, there's nothing for it other than to
+    // wait for the interrupt handler to free space
+    while (!availableForWrite()) {
+      // nop, the interrupt handler will free up space for us
+    }
+
+    // HAL doesn't manage rollover, so split transfer till end of TX buffer
+    // Also, split transfer according to available space in buffer
+    while ((size > available_till_buffer_end) || (size > available)) {
+      size_intermediate = min(available, available_till_buffer_end);
+      write(buffer, size_intermediate);
+      size -= size_intermediate;
+      buffer += size_intermediate;
+      available = availableForWrite();
+      available_till_buffer_end = SERIAL_TX_BUFFER_SIZE - _serial.tx_head;
+    }
+
+    // Copy data to buffer. Take into account rollover if necessary.
+    if (_serial.tx_head + size <= SERIAL_TX_BUFFER_SIZE) {
+      memcpy(&_serial.tx_buff[_serial.tx_head], buffer, size);
+      size_intermediate = size;
+    } else {
+      // memcpy till end of buffer then continue memcpy from beginning of buffer
+      size_intermediate = SERIAL_TX_BUFFER_SIZE - _serial.tx_head;
+      memcpy(&_serial.tx_buff[_serial.tx_head], buffer, size_intermediate);
+      memcpy(&_serial.tx_buff[0], buffer + size_intermediate,
+             size - size_intermediate);
+    }
+
+    // Data are copied to buffer, move head pointer accordingly
+    _serial.tx_head = (_serial.tx_head + size) % SERIAL_TX_BUFFER_SIZE;
+
+    // Transfer data with HAL only is there is no TX data transfer ongoing
+    // otherwise, data transfer will be done asynchronously from callback
+    if (!serial_tx_active(&_serial)) {
+      // note: tx_size correspond to size of HAL data transfer,
+      // not the total amount of data in the buffer.
+      // To compute size of data in buffer compare head and tail
+      _serial.tx_size = size_intermediate;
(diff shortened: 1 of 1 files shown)


cores/arduino/Serial.cpp
```diff
@@ -583,6 +583,14 @@ void Uart::flush(uint32_t timeout)
...
 size_t Uart::write(const uint8_t *buffer, size_t size)
 {
+  // If begin() never brought the hardware up (e.g. the requested baud
+  // rate can't be reached from the configured kernel clock), the TX
+  // interrupt this function waits on below will never fire. Bail out
+  // instead of spinning forever once the TX buffer fills up.
+  if (!_ready) {
+    return 0;
+  }
+
```
fpistm: We tried to avoid several return in a function to be MISRA compliant.
Could you change your PR to only have one return.
Like move `ret` declaration before and set it to 0 if` !_ready` else 
```suggestion
  size_t ret = size;
  if (!_ready) {
    ret =  0;
  } else {
  ...
  }
    /*  Return 0 if not ready else transfer size requested */
  return ret;
```

fpistm (changes requested):
Hi @saikumar-mandaji 
Thanks for the PR. Makes sense.
Only one comment to address.

fpistm (approved):

t-miura:
Thank you so much for the fix, I will test with Meshtastic very soon!

